PR #17, opened 2020-11-04
Improved theme integration and related fixes

atomworks:


atomworks:
This is ready to be reviewed now!

That conflict in the Language.tsx, should just be able to switch to my version as I fixed the faults in how we pulled that one in.

src/Icons/Icon.tsx
```diff
@@ -8,7 +8,7 @@ export {IconSVGs};
 
 const IconWrapper = styled.div<{color:string}>`
   [stroke]{
```
JoshLipps, line 10: This is pretty weird looks for elements with a attribute stroke and override with CSS. But I guess this is OK based on this.
 https://www.w3.org/TR/SVG/styling.html#PresentationAttributes

JoshLipps (approved):

PR #23, opened 2020-12-26
Feature/tab component

fti-bcarpenter:
This PR has an addition of tabs component in scorer-ui-kit

### DIFF
--- a/src/Tabs/Tab.tsx
+++ b/src/Tabs/Tab.tsx
@@ -29,7 +29,7 @@ interface OwnProps {
 
 type Props = OwnProps & HTMLAttributes<HTMLDivElement>
 
-const Tab: React.FC<Props> = ({ children, tabFor }) => {
+const Tab: React.FC<Props> = ({ children, tabFor, ...props }) => {
   const { selected, setSelected }: ContextProps = useContext(TabContext);
 
   const onChangeTab = useCallback((tabId: string) => {
@@ -37,7 +37,7 @@ const Tab: React.FC<Props> = ({ children, tabFor }) => {
   }, [setSelected]);
 
   return (
-    <TabComponent onClick={() => onChangeTab(tabFor)}>
+    <TabComponent {...props} onClick={() => onChangeTab(tabFor)}>
       <TabLabel active={selected === tabFor}>
         {children}
       </TabLabel>
--- a/src/Tabs/Tab.tsx
+++ b/src/Tabs/Tab.tsx
@@ -1,5 +1,5 @@
 import React, { useContext, HTMLAttributes, useCallback } from 'react';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 import { TabContext, ContextProps } from './Tabs';
 
 const TabComponent = styled.div`
@@ -10,7 +10,9 @@ const TabComponent = styled.div`
 
 const TabLabel = styled.label<{ active: boolean }>`
   height: 40px;
-  font-family: Monorale;
+  ${({theme}) => css`
+    font-family: ${theme.fontFamily.ui};
+  `}
   font-size: 15px;
   font-weight: ${({ active }) => active ? '600' : '500'};
   letter-spacing: 0.09px;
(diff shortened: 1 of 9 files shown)


src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const TabComponent = styled.div`
+  margin-right: 39px;
+  display: flex; 
+  align-items: center;  
+`;
+
+const TabLabel = styled.label<{ active: boolean }>`
+  height: 40px;
+  font-family: Monorale;
```
atomworks, line 13: Moving forward we are trying to use the theme to set fonts, so new work ideally calls fonts from there (even though there are instances that still need updating to this). 

For setting a font try using:
```
${({theme}) => css`
    font-family: ${theme.fontFamily.ui};
`}
```
There are different ways to write this but this will give you the idea. Using this pattern means you can add in other things too. Theme will be available by default in this context so as long as this is in the Styled Component, it should be good to use!

For reference:
`theme.fontFamily.ui` will be Monorale.
`theme.fontFamily.data` will be Lato.

This means in future projects where fonts may change, we can handle this really easily. It's not enough to block a merge but definitely a big nice to have.

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,25 @@
...
+});
+
+type Props = HTMLAttributes<HTMLDivElement>
+
+const Tabs: React.FC<Props> = ({ children }) => {
+  const [selected, setSelected] = useState('');
+  const value = { selected, setSelected };
+
+  return (
+    <TabContext.Provider value={value}>
```
atomworks, line 19: Seems a bit redundant to set `value` if it's just passing along the useState vars. Perhaps just `value={ selected, setSelected }` would work?

This isn't a big thing really, just tidiness.

fti-bcarpenter: Done

src/Tabs/Tab.tsx
fti-bcarpenter: Done

fti-bcarpenter:
@atomworks suggested changes are done in PR. Please check.

src/Tabs/TabList.tsx
```diff
@@ -0,0 +1,30 @@
...
+  flex-direction: row;
+`;
+
+interface OwnProps {
+  defaultTabId: string
+}
+
+type Props = OwnProps & HTMLAttributes<HTMLDivElement>
+
+const TabList: React.FC<Props> = ({ children, defaultTabId, ...props }) => {
```
JoshLipps, line 16: Is this intentional? It seems the only component that we can override styles on is `TabList` shouldn't `TabContent` at least also have this or not be a div?

JoshLipps (commented):
Looks Good Just made one comment.

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,48 @@
...
+  padding-bottom: 14px;
+`;
+
+interface OwnProps {
+  tabFor: string
+}
+
+type Props = OwnProps & HTMLAttributes<HTMLDivElement>
+
+const Tab: React.FC<Props> = ({ children, tabFor }) => {
```
JoshLipps, line 32: Maybe `, ...props` here and below on TabComponent div

src/Tabs/TabContent.tsx
```diff
@@ -0,0 +1,26 @@
...
+const Container = styled.div`
+`;
+
+interface OwnProps {
+  tabId: string
+}
+
+type Props = OwnProps & HTMLAttributes<HTMLDivElement>
+
+const TabContent: React.FC<Props> = ({ children, tabId}) => {
```
JoshLipps, line 14: Same as above for ...props

fti-bcarpenter: Done

src/Tabs/Tab.tsx
fti-bcarpenter: Done

src/Tabs/TabList.tsx
fti-bcarpenter: Yes, ...props is not required here. Removed now.

PR #24, opened 2021-01-13
Fix cell hover for copy icon

isacoder:
### Requirements
 https://www.bugherd.com/projects/216308/tasks/9
 Tables: Copy button often can't be reached because hover area is left. (see attachment)
 
 ### Implementation
 Before this fix the problem seems to be that the copy button is absolute positioned -20px right and the icon overlaps in the area of the next table cell. 
 
I added conditional padding to header and cell depending of the config. This will keep the icon inside the cell.
I pushed the example config to have the copy icon in the created column.
 
 ### Screenshoots
 
 Screen Shot 2021-01-14 at 10 27 52


atomworks (approved):
👍  

Thank you!

PR #26, opened 2021-01-14
Feature/Textarea

isacoder:
### Requirements
[Bugherd #16](https://www.bugherd.com/projects/216308/tasks/16)
[New] TextArea: Using the TextInput component as an example, create a TextArea field.

### Implementation
Following the de indications, created two components. 

TextArea and TextAreaField
The story is based in TextAreaField.

Colors and actions were copied/duplicated from Input component with minimum styling update.
I see a potential abstraction of FeedbackMessage ( Container, Icon, Message) to avoid some of it and could become an atom of it's own.

Although is not part of the design other properties of Textarea are maxlength and minlength sometimes is great to display the number of how many chars are left when typing. This could be another improvement in a later design.

Please let me know your thoughts in this and any other posible improvements

### Screenshoots

Screen Shot 2021-01-19 at 19 26 33


src/Form/molecules/TextareaField.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { TextareaHTMLAttributes } from 'react';
+import styled from 'styled-components';
```
atomworks, line 2: This import isn't needed as it's not used.

storybook/src/stories/Form/TextAreaField.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+import styled from 'styled-components';
+import {  text, select } from "@storybook/addon-knobs";
+import { TextAreaField } from 'scorer-ui-kit';
+
+const Container = styled.div`
+  margin: 20px;
+`;
+
+export default {
+  title: 'Form/molecules',
```
atomworks, line 11: The other Input stories in Storybook use the title `Form/Input` which I think is probably better than using molecules. I agree though that perhaps `Input` isn't the best subfolder for this. Perhaps we can combine the two stories in `Form/Input` with this one under a different name. Perhaps `Form/Fields` or something like that? Maybe even just putting them all under `Form/molecules` would be good for now. 

What do you think?

storybook/src/stories/Form/TextAreaField.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+  component: TextAreaField,
+  decorators: []
+};
+
+
+export const TextareaField = () => {
```
atomworks, line 17: When the Storybook function name clashes with the UI Kit's component name you can use the same thing and just prefix is with `_`. For example, `_TextAreaField`.

This works though and doesn't have any effect elsewhere so it doesn't need changing, just a tip I thought I would share! :)

atomworks (changes requested):
In addition to some of the small comments there are a couple of bugs. 

- Background color on the feedback message seems to disappear when the field is focused. 
- I get a compile error because I think you renamed `TextAreaField.tsx` but Git ignored it as it's just a casing changes, so you need to force it to recognise that.

Other than that though, it's looking great! 🥳

storybook/src/stories/Form/TextAreaField.stories.tsx
isacoder: Hi! I'm glad you mention it, I was not sure where to add it actually.
I saw that **TextField** on Inputs, but **Small Input** was an atoms.

Technically speaking all the items under Form are "Fields" so I wouldn't go for Form/Fields.

I have no problem in adding it to Form/Input if it's just to have less  folders.
Also let me know if you are interested that I move Small Input to Inputs in this PR.

storybook/src/stories/Form/TextAreaField.stories.tsx
isacoder: Thanks for the tip, updated! 😄

src/Form/molecules/TextareaField.tsx
isacoder: Hi! In this case it's needed becase I'm sending props down to TextArea.
The specific scenario of sending "placeholder" as prop shows a error message.

isacoder:
> In addition to some of the small comments there are a couple of bugs.
> 
> * Background color on the feedback message seems to disappear when the field is focused.
> * I get a compile error because I think you renamed `TextAreaField.tsx` but Git ignored it as it's just a casing changes, so you need to force it to recognise that.
> 
> Other than that though, it's looking great! 🥳

Thanks for finding the focus bug, I can't believe it didn't happened to me before 😱 
This is not a problem in the TextField because the feedback message is not bellow.
I only removed the rule that was adding the shadow and removing background on the theme and replaced with the simple box shadow values.

It's kind of hardcoded outside the theme, so there is an area of improvement so if you have a recommendation on it please let me know.

storybook/src/stories/Form/TextAreaField.stories.tsx
atomworks: I'll add the sorting to a different issue in BugHerd and we and tackle that soon. Don't want to hold this merge up and fixing it later shouldn't cause any problems.

atomworks (approved):

PR #31, opened 2021-02-08
Feature/notification synchronization

isacoder:
### Requirements
This PR is  the update of UI requests and files from the last release that were pending but not blocking.

- Notification Synchronization that will trigger only after no other notification is active. 
- useNotification abstraction of useContext on it's own file
- Notification with text will close before triggering action


atomworks (changes requested):
That all looks great. One small request...

On line 52 of Notification.tsx can we change the animation speed from `slower` to `normal`? Because of the capitalisation issue on the branch I'm unable to do this myself quickly. Thank you!

https://github.com/future-standard/scorer-ui-kit/blob/0b835fcd3a17ea5cc730f5b94b2a2f36cd281142/src/Alerts/atom/Notification.tsx#L52

atomworks (approved):

PR #33, opened 2021-02-12
Feature/modal

isacoder:
### Requirements
[Bugherd#17](https://www.bugherd.com/projects/216308/tasks/17)

This is the  PR for Modal component with the use of Context that will have two main interactions

- Function to update the content
- Function to open/close

This will become the base to have multiple templates for modals

On this PR includes

- Base empty modal
- Template for simple Confirmation Modal
-- Title
-- Text
-- Two buttons

Screen Shot 2021-02-19 at 12 02 39

### Implementation

- Modal (Base modal)
- ModalContext( Provides modal and Updates props)
- useModal ( Allows any component to interact with the modal)

### Screenshots

Empty modal example with isEnableClose option true look at the orange part of the image for reference
can be close clicking outside
Screen Shot 2021-02-19 at 11 57 17

Template for Confirmation Modal can only be close selecting an option

Screen Shot 2021-02-19 at 11 57 08


src/Icons/Icon.tsx
```diff
@@ -13,22 +13,26 @@ const IconWrapper = styled.div<{color:string}>`
...
+const Icon : React.FC<IProps> = ({icon, size = 24, weight = 'regular', color = 'mono', className}) => {
 
   const getKeyValue = <U extends keyof T, T extends object>(key: U) => (obj: T) => obj[key];
 
   const iconWeight : number = dimensions.icons.weights[weight];
   const IconSVG = getKeyValue<keyof any, any>(icon)(IconSVGs);
   if(icon in IconSVGs && IconSVG){
 
-    return <IconWrapper color={color}>
+    return <IconWrapper className={className} color={color}>
```
atomworks, line 35: I'm not sure this is the best way to solve this problem. It sounds like it could be a bug that needs fixing in SVG generation perhaps. Could you show me where you used this className prop in one of your fixes? I couldn't find it in the files.

As we're using StyledComponents I feel like it's best to try and avoid use of class names as it can open things up to problematic solutions and workarounds potentially.

src/hooks/useClickOutside.tsx
```diff
@@ -0,0 +1,23 @@
+import react, {useEffect, useRef } from 'react';
+
+export type outClickType = (eve: MouseEvent) => void;
+export function useClickOutside(elRef : any , callback : outClickType) {
```
atomworks, line 4: This looks great. Can you think of any other places we might be able to use this? I can see it being useful but I'm not sure where exactly. If I can keep those ideas in mind I think it will help some of my design choices 😄

src/Alerts/atom/Modal.tsx
```diff
@@ -0,0 +1,139 @@
...
+  z-index: 99;
+  width: ${({ width }) => width ? width : `580px`};
+  padding: ${({ padding }) => padding ? padding : `30px 40px`};
+  border-radius: ${({ borderRadius }) => borderRadius ? borderRadius : `10px`};
+  border: ${({border}) => border ? border : `solid 1px hsl(0, 14%, 90%)`};
+`;
+
+// export interface IModalStyleProps {
+
+// }
```
atomworks, line 78: Can delete these to tidy up a little! Feel free to point out those kind of things on my PRs anytime by the way. I always miss the tidy up moment 😆

src/Alerts/atom/Modal.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+const Modal: React.FC<IModalProps> = ({
+  isOpen = false,
+  isCloseEnable = true,
+  closeText = '',
+  padding = '',
+  width = '',
+  borderRadius,
+  border = '',
+  boxShadow = '',
```
atomworks, line 101: I think these can be removed as per previous review comment.

src/Alerts/atom/ConfirmationModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+  rightButtonText = 'cancel',
+  rightButtonCallback = () => { },
+}) => {
+
+  const { setIsOpen } = useModal();
+
+  const handlePrimaryBtn = () => {
+    leftButtonCallback();
+    setIsOpen(false);
+  }
```
atomworks, line 67: With this, is there an easy way to `setIsOpen(false)` from executing if the primary action has failed?

For example, the user wants to delete a file and so we add a checkbox that says, "I confirm I want to delete this file". If they haven't checked it and the user presses the button, we need to be able to keep the modal open.

If this is more complex we can move this to its own improvement issue. Perhaps we need to consider being able to set  enabled / disabled states too. 🤔

example/src/pages/ModalPage.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+const CardModal = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+
+const ModalPage: React.FC = () => {
+
+  // This is how I imagine that the hook is going to end
+  //const {setIsOpen, updateCloseText, closeCallback, updateContent} = useModal();
```
atomworks, line 103: Is this  still relevant or are those extra concepts achievable with what you have implemented? Perhaps we can branch these out to their own tasks for future improvement if they are still to do, they sound like they could be worthwhile.

atomworks (changes requested):
This is looking really well engineered! You've really managed to abstract the brief and be really forward looking with it.

I've added a few comments throughout. Just let me know if anything doesn't make sense and you want to discuss it further. I think implementation oriented things are the priority and some stuff perhaps we can put in future iterations after this PR has merged in to the UI Kit.

src/hooks/useClickOutside.tsx
isacoder: Well this is an idea from the video I shared about custom hooks. 
I that video it is proposed to be used with a menu or with drawer.
I hope is useful It was a pain to implement with typescript but I learned a bunch about refs 🤣 

https://youtu.be/J-g9ZJha8FE?t=483

src/hooks/useModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+      isOpen: newStatus,
+    }
+    setModalProps(updateProps);
+  }, [modalProps]);
+
+  const createModal = (
+    closeText?: string,
+    isCloseEnable?: boolean,
+    width?: string,
+    padding?: string,
```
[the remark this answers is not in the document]
isacoder: ok, updating padding to boolean, default true;

src/hooks/useModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+    borderRadius? :string,
+    border? :string,
+    boxShadow?: string,
```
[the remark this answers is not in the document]
isacoder: Removing it for now as we talked, I'm going to leave border-radius: 5px default, then we can think of the theme value later.

src/hooks/useModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { stringify } from 'querystring';
+
+const useModal = () => {
+  const { modalProps, setModalProps } = useContext(ModalContext);
+
+  const onDismiss = useCallback(() => {
+    setIsOpen(false);
+  }, []);
+
+  const setIsOpen = useCallback((newStatus: boolean) => {
```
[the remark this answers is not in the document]
isacoder: Thanks for the suggestion, It makes more sense I will update it.

src/hooks/useModal.tsx
```diff
@@ -0,0 +1,57 @@
+import { ReactElement, useCallback, useContext, useEffect } from 'react';
+import { ModalContext } from '../context/ModalContext';
+import { IModalProps } from '../Alerts/atom/Modal';
+import { stringify } from 'querystring';
```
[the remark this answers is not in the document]
isacoder: 🤦‍♀️ VS Code does it again. Sorry I'll be more careful.

src/Alerts/atom/ConfirmationModal.tsx
isacoder: In that specific scenario I imagine that User should evaluate if the "Form" or "Elements in component" required are valid setting it too close. Something similar to this:

<Button
onClick ={ 
   () => {
      const areFieldsValid = validateFields()
      if(areFieldsValid) { 
            do something ... then setModalOpen(false) 
      }else {
         showInvalidFields();
        }
     }
}> Delete File </Button>

I think allowing the developer to use setModalOpen value gives a lot of power to have this flexible 🤔

example/src/pages/ModalPage.tsx
isacoder: Sorry this were just notes. I'll delete it.

example/src/pages/ModalPage.tsx
```diff
@@ -0,0 +1,207 @@
...
+    leftButtonText='Back'
+    leftButtonDesign='secondary'
+    leftButtonCallback={() => { console.log('Action Canceled [Back] clicked') }}
+    rightButtonDesign='primary'
+    rightButtonText='I understand'
+    rightButtonCallback={() => { console.log('Action Confirmed [I understand] clicked') }}
+  />
+
+  const openConfirmationModal = () => {
+    createModal('', false,'','','','','', addPersonModal);
```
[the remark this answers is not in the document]
isacoder: Thanks for the recommendation I have added as a type and It looks way better.

src/Icons/Icon.tsx
isacoder: The problem was that the Div that is wrapping the icon is adding 3px height by default (no idea why 😞 )
Because of that I couldn't align it vertically. I added flex to the wrapper via className.

Screen Shot 2021-02-25 at 18 08 56

However I took a longer look to the Icon.tsx and notice that it's exporting the IconWrapper.
I understand that I should be using the exported wrapper to interact with it. 
I have removed the className option from the Icon.tsx

src/hooks/useModal.tsx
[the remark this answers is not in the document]
atomworks: It's all good. It catches me out too from time to time. If it broke something it would be easier to spot but it always sneaks in when you're (often literally) not looking.

src/hooks/useClickOutside.tsx
atomworks: Wow, yeah... that's _extremely_ useful! All of a sudden I can see a ton of different things this is going to be useful for. 🥳

example/src/pages/ModalPage.tsx
atomworks: All good! Those notes come in useful during these reviews.

src/Icons/Icon.tsx
atomworks: Glad you could solve it this way. I'm sure there's some improvements we can make to make this process easier to understand and work with maybe

src/Alerts/atom/ConfirmationModal.tsx
atomworks: Is there an example or can we get an example for this? I think this is something we want people to be able to understand as I'm sure it will come up quickly.

atomworks (changes requested):
Regarding the conversation in `src/Alerts/atom/ConfirmationModal.tsx` about handling the callback and modal close moment, do we have / can we get an example (in examples) of the how that would look? I think it's something that we'll run in to quickly.

src/Alerts/atom/ConfirmationModal.tsx
isacoder: I have updated

isacoder:
> Regarding the conversation in `src/Alerts/atom/ConfirmationModal.tsx` about handling the callback and modal close moment, do we have / can we get an example (in examples) of the how that would look? I think it's something that we'll run in to quickly.

Hi! I have added an example of a login based in the other LoginPage.
It took me a bit more than expected because of an obvious mistake that was not so obvious.

It is important that the whole component is passed to createModal, not only the JSX ( it's required to pass the state variables definition and functions too). Although for informative modal is simple to create JSX with the message. I would recommend that when passing a customComponent the custom component is actually in a different file to avoid that confusion. ( 🙈  I didn't do it at the beginning).

**In the example you can only login with** 
username: user
password: fakepass123

Screen Shot 2021-03-01 at 19 17 39


```js
  const handleModalSubmit = useCallback(async (e: React.FormEvent<HTMLFormElement | HTMLButtonElement>) => {
    e.preventDefault();

    // Validate inputs and show errors
    const areInputsValid = validateFields(form);
  
    // if inputs are not valid return to allow user to correct before submiting
    if(!areInputsValid) { return; }

    const loginSuccess = await onSubmit(form);

    // if there was a login error return to allow user to correct and retry
    if(!loginSuccess) {return;}

    // if there were no errors close modal
    setModalOpen(false);

  },[onSubmit, setModalOpen, form]);
```

If you have any questions about this example please don't hesitate to call.
I would also appreciate your input about how easy is this example to understand.

atomworks (approved):
Looks great. Thank you for all that extra work :)

PR #34, opened 2021-02-19
Updated icons in DatePicker

atomworks:
And also removed a chunk of old commented out code that was used when refactoring during development.


isacoder (approved):
I have reviewed the code and run in storybook.
I cons look good at it's respective placement!

PR #37, opened 2021-03-01
Feature/table type reload fix

JoshLipps:
Fixed a few bugs in TypeTable that were impacting TrafficCounter:

1. Lots of linting
2. changed allChecked logic to use a useEffect instead of a useCallback (prevents calling every render) 
3. Fixed checkbox to actually listen for checked to change in its useEffect
4. Changed Row rendering to be on new row list instead of using useCallback

### DIFF
--- a/src/Form/atoms/Checkbox.tsx
+++ b/src/Form/atoms/Checkbox.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import styled, { css } from 'styled-components';
 
 import Icon from '../../Icons/Icon';
@@ -111,7 +111,7 @@ const Checkbox : React.FC<IProps> = ({ indeterminate = false, disabled, checked 
   const [ visualState, setVisualState ] = useState<CheckboxState>(checked ? CheckboxState.On : CheckboxState.Off);
 
   const customOnChange = (e: any) => {
-    const checked = e.target.checked
+    const checked = e.target.checked;
 
     setIsChecked(checked);
     if(onChangeCallback){ onChangeCallback(checked); }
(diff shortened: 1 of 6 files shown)


src/hooks/useCopyToClipboard.tsx
```diff
@@ -1,32 +1,34 @@
-export const useCopyToClipboard = (text : string) => {
-  return copyToClipboard(text)
-};
+import { useCallback } from 'react';
 
-const copyToClipboard = (str : string) : boolean => {
+export const useCopyToClipboard = () => {
+  const copyToClipboard = useCallback((str : string) : boolean => {
```
JoshLipps, line 4: Move copyToClipboard into a `useCallback` that is provided by `useCopyToClipboard` hook

src/Tables/molecules/TypeTable.tsx
```diff
@@ -91,4 +98,8 @@ const TypeTable : React.FC<IProps> = ({ columnConfig, selectable, selectCallback
   );
 };
 
-export default TypeTable;
\ No newline at end of file
+export default TypeTable;
+
+function isChecked({_checked=false} : IRowData){
```
JoshLipps, line 103: move this out so it doesn't need to be re declared on re render (its Pure)

src/Tables/molecules/TypeTable.tsx
```diff
@@ -49,18 +49,25 @@ interface IProps {
...
+  selectCallback? : (checked:boolean, id?: string | number)=>void
+  toggleAllCallback? : (checked: boolean)=>void
 }
 
-const TypeTable : React.FC<IProps> = ({ columnConfig, selectable, selectCallback, toggleAllCallback, rows, hasStatus = false, hasThumbnail = false, hasTypeIcon = false }) => {
+const TypeTable : React.FC<IProps> = ({
+  columnConfig,
+  selectable,
+  selectCallback = ()=>{},
+  toggleAllCallback = ()=>{},
```
JoshLipps, line 60: Added Default functions for optional functions, the removes the need for "is defined" checks when not specified.

src/Tables/molecules/TypeTable.tsx
```diff
@@ -49,18 +49,25 @@ interface IProps {
   hasStatus?: boolean
   hasThumbnail?: boolean
   hasTypeIcon?: boolean
-  selectCallback? : any
-  toggleAllCallback? : any
+  selectCallback? : (checked:boolean, id?: string | number)=>void
+  toggleAllCallback? : (checked: boolean)=>void
```
JoshLipps, line 53: Fixed types for these optional Callbacks.

src/Form/atoms/Checkbox.tsx
```diff
@@ -124,20 +124,22 @@ const Checkbox : React.FC<IProps> = ({ indeterminate = false, disabled, checked
 
     setVisualState( state );
 
-  }, [isChecked, setVisualState])
+  }, [checked, isChecked, setVisualState]);
```
JoshLipps, line 127: This allows checked to update correctly when external setting changes.

atomworks (approved):
Looks good! Thanks for tidying up some of those loose ends from the original implementation.

PR #40, opened 2021-03-02
Custom component Functionality to TypeTable

isacoder:
### Requirement

This PR covers 3 bugherd tasks

[Burherd#32](https://www.bugherd.com/projects/216308/tasks/32)
[Table] Upgrade to allow passing of React Elements to column values instead of just text

[Bugherd#35](https://www.bugherd.com/projects/216308/tasks/35)
[Table] Add action buttons options
Design comment #2 https://zpl.io/aB46w0k

[Bugherd#36](https://www.bugherd.com/projects/216308/tasks/36)
[Table] Add two line custom cell template

### Implementation

Adding the React element was not very intrusive but need to review styling options.

I have create the custom template for the two line as a Multiple line, it will depend in the amount of items received by the array. However I'm not sure how useful this is, it feels more like a helper than a feature. I'm a bit worried I might have misunderstood something.

And for this specific case scenario that deals with dates configuration I it's own component would have been a better solution.
Perhaps we can talk a bit more about it on a call. 

### Screenshoots

Screen Shot 2021-03-05 at 10 41 53


src/Form/molecules/ActionButtons.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const Container = styled.div`
+  display: flex;
+  ${StyledIconButton} {
+    margin-left: 15px;
+  }
+  ${StyledIconButton}:first-child {
+    margin-left: 0px;
+  }
+`;
```
isacoder, line 15: Currently ActionButtons keeps them in the same row, but I'm not sure if it would be something desirable to also be on top of each other. One Idea to consider is that It could have a directionProp.

storybook/src/stories/Tables/molecules/ActionsTable.stories.tsx
```diff
@@ -0,0 +1,215 @@
...
+import styled from 'styled-components';
+import { object, boolean } from "@storybook/addon-knobs";
+import { action } from '@storybook/addon-actions';
+
+
+import {
+TypeTable as TypeTableCustom,
+ActionButtons,
+IconButtonData,
+MultilineContent,
```
isacoder, line 11: These are the two new Templates for Cell

isacoder (commented):
I have added a few comments to help review

src/Form/molecules/ActionButtons.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+type IGroupButtonsData = {
```
atomworks, line 17: Should this be an `interface` instead?

src/Form/molecules/ActionButtons.tsx
atomworks: For the foreseeable future we only need them in a row as they are currently. There may be a case for vertical stacked in the future but it's not an essential requirement right now.

storybook/src/stories/Tables/molecules/ActionsTable.stories.tsx
atomworks: 👍

atomworks (approved):
There's only one small thing I spotted regarding use of `type` where `interface` would be better, if there's a reason that I'm missing though, let me know. Either way, I don't think we need to hold up merging but maybe you'll get a  chance to update that before we do. 

Great work. Really like the way you've approached things here.

PR #42, opened 2021-03-04
Feature/date time improvements use dates

JoshLipps:
## Refactor of Date filter Component. 

Is this necessary... :man_shrugging: 

Anyway I refactored this to use Date components internally. It also does 2 more important things: 
1. allows init of a singe date
2. in single time mode and date mode it returns a single date.
3. added back start < end enforcement
4. on mode change init to today... (weird but also why... are we changing modes)
I think it also fixes a bug where you could type weird large number into time boxes.

~~... found a bug please hold..... :bug:~~ Please feel free to move about the cabin again :airplane:


atomworks (approved):
Looks good. I removed a `console.log` for you btw.

PR #43, opened 2021-03-08
Feature/table loading

isacoder:
### Requirements

[Bugherd#18] (https://www.bugherd.com/projects/216308/tasks/18)
[Table] Add a loading state

Designs 
zpl://project?pid=5ee9f5a23a30524bf9e5b2a9&seid=6037717f25bda624a06e0f13

### Implementation

To show loading state the isLoading value must be true.
If the table is empty and is loading it will show the loading empty table state.

If the table is empty but is not loading it will show the "No Data" message. However I'm not sure if it would be better to have this message as optional. Since the default message is so simple and it would be complicated to cover all the cases.

### Screenshots

Loading without data
Screen Shot 2021-03-08 at 21 40 34
 
Loading with data

Screen Shot 2021-03-08 at 21 41 18

Empty table

Screen Shot 2021-03-08 at 22 25 46

### DIFF
--- a/src/Tables/atoms/TypeTableRow.tsx
+++ b/src/Tables/atoms/TypeTableRow.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback } from 'react';
-import styled from 'styled-components';
+import styled, {css} from 'styled-components';
 
 import Checkbox from '../../Form/atoms/Checkbox';
 import TableRowThumbnail from './TableRowThumbnail';
@@ -9,8 +9,11 @@ import Icon from '../../Icons/Icon';
 import { ITableColumnConfig, IRowData } from '..';
 
 
-const RowContainer = styled.div`
+const RowContainer = styled.div<{isEmpty: boolean}>`
   display: table-row;
+  ${({isEmpty}) =>  isEmpty && css`
+    visibility: hidden;
+  `};
 `;
 
 interface IProps {
@@ -30,9 +33,10 @@ const TypeTableRow : React.FC<IProps> = ({selectable = false, selectCallback, ha
     if(selectCallback){ selectCallback(checked, rowData.id); }
   }, [rowData.id, selectCallback]);
 
-  if(rowData?.columns?.length === 0) { return null;}
+  const isEmpty = rowData.columns.length === 0;
+
   return (
-    <RowContainer>
+    <RowContainer isEmpty={isEmpty}>
       {selectable ? <TypeTableCell hideDivider><Checkbox checked={rowData._checked} onChangeCallback={wrappedSelectCallback} /></TypeTableCell> : null}
       {hasStatus ?  <TypeTableCell hideDivider><TypeTableDeviceStatus status={rowData.header?.status} /></TypeTableCell> : null}
       {hasThumbnail ? <TypeTableCell hideDivider><TableRowThumbnail image={rowData.header?.image} /></TypeTableCell> : null}
--- a/src/Tables/atoms/TypeTableRow.tsx
+++ b/src/Tables/atoms/TypeTableRow.tsx
@@ -30,6 +30,7 @@ const TypeTableRow : React.FC<IProps> = ({selectable = false, selectCallback, ha
     if(selectCallback){ selectCallback(checked, rowData.id); }
   }, [rowData.id, selectCallback]);
 
+  if(rowData?.columns?.length === 0) { return null;}
   return (
     <RowContainer>
       {selectable ? <TypeTableCell hideDivider><Checkbox checked={rowData._checked} onChangeCallback={wrappedSelectCallback} /></TypeTableCell> : null}
--- a/src/Tables/molecules/TypeTable.tsx
+++ b/src/Tables/molecules/TypeTable.tsx
@@ -1,7 +1,7 @@
 import React, { useCallback, useEffect, useState } from 'react';
 import styled, {css} from 'styled-components';
-import Spiner from '../../Indicators/Spinner'
-;import TypeTableRow from '../atoms/TypeTableRow';
+import Spiner from '../../Indicators/Spinner';
+import TypeTableRow from '../atoms/TypeTableRow';
 import Checkbox from '../../Form/atoms/Checkbox';
 import { TypeCellAlignment, ITableColumnConfig, ITypeTableData, IRowData } from '..';
 
--- a/src/Tables/molecules/TypeTable.tsx
+++ b/src/Tables/molecules/TypeTable.tsx
@@ -21,7 +21,7 @@ const HeaderRow = styled.div`
   height: ${HEADER_HEIGHT};
 `;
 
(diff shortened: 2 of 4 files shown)


storybook/src/stories/Tables/molecules/LoadingTable.stories.tsx
```diff
@@ -0,0 +1,148 @@
...
+  padding: 100px;
+`;
+
+
+const columnConfigSample : ITableColumnConfig[] = [
+  {
+    header: 'When',
+    sortable: false,
+    cellStyle: 'normalImportance',
+    showStatus: true,
```
atomworks, line 29: Isn't a `showStatus` type of data.

storybook/src/stories/Tables/molecules/LoadingTable.stories.tsx
```diff
@@ -0,0 +1,148 @@
...
+    header: 'Status',
+    sortable: false,
+    cellStyle: 'normalImportance',
+    showStatus: true,
+  },
+  {
+    header: 'Temperature',
+    sortable: false,
+    cellStyle: 'normalImportance',
+    showStatus: true,
```
atomworks, line 41: Isn't a `showStatus` type of data.

src/Tables/molecules/TypeTable.tsx
```diff
@@ -42,26 +45,75 @@ const HeaderItem = styled.div<{fixedWidth?: number, alignment?: TypeCellAlignmen
 
 `;
 
+const LoadingText = styled.div`
+  color: hsla(195, 10%, 52%, 0.72);
+}
```
atomworks, line 50: Stray closing tag? (unless I missed something)

src/Tables/atoms/TypeTableRow.tsx
```diff
@@ -30,6 +30,7 @@ const TypeTableRow : React.FC<IProps> = ({selectable = false, selectCallback, ha
     if(selectCallback){ selectCallback(checked, rowData.id); }
   }, [rowData.id, selectCallback]);
 
+  if(rowData?.columns?.length === 0) { return null;}
```
atomworks, line 33: Seems like this is perhaps part of the cause for some strange layout shift moment when data drops in. We can chat about this a bit more on call.

src/Tables/molecules/TypeTable.tsx
```diff
@@ -42,26 +45,75 @@ const HeaderItem = styled.div<{fixedWidth?: number, alignment?: TypeCellAlignmen
...
+`;
+const LoadingBox = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  z-index: 99;
+  background-color: hsla(210, 14%, 97%, 0.85);
```
atomworks, line 57: Can we make the `background-color` the `pureBase` in  the theme? At the moment the grey doesn't work over white but a lot of designs use the grey. We may need to update it again but if it's tied in to the theme it's a step in the right direction, even if it doesn't fix the problem just yet. I'll have to do that on my themeing part!

storybook/src/stories/Tables/molecules/LoadingTable.stories.tsx
```diff
@@ -0,0 +1,148 @@
...
+  },
+];
+
+export const _LoadingTable = () => {
+  const isLoading = boolean("IsLoading", true);
+  const emptyTable = boolean("Show Empty Table", true);
+  const loadingText = text("loadingText", 'Loading Data..')
+  const columnConfig = object("Column Configuration", columnConfigSample);
+  const [rows, setRows] = useState<ITypeTableData>(initialRows);
+
```
atomworks, line 110: Can we also add the props to change the empty state text?

atomworks (changes requested):
Looking good although I have a few requests to help bring it together. We can chat about those on a call later today. I've added a few comments for code related feedback though!

-

storybook/src/stories/Tables/molecules/LoadingTable.stories.tsx
isacoder: showStatus is boolean :)
Is the dot to the left.

By the way I will remove the extra dots in the other columns, that were added by mistake 🙈 

Screen Shot 2021-03-11 at 19 22 29

src/Tables/molecules/TypeTable.tsx
isacoder: ok, I will update this BG to pureBase and put the opacity on 85% outside of the backgound-color value;

src/Tables/molecules/TypeTable.tsx
```diff
@@ -1,23 +1,26 @@
 import React, { useCallback, useEffect, useState } from 'react';
 import styled, {css} from 'styled-components';
-
+import Spiner from '../../Indicators/Spinner';
```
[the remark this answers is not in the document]
isacoder: Thank you for pointing it out 😱  not sure how is this even working. Mystery alias helper maybe ...

src/Tables/molecules/TypeTable.tsx
isacoder: Thank you for mentioning it. I guess I did a bad cut.
I have updated.

storybook/src/stories/Tables/molecules/LoadingTable.stories.tsx
isacoder: I have added the knobs action for the Empty state table.
Also I have updated the default  text to empty as we talked about it.

isacoder:
Updated the table to hide cells instead of just not rendering but still is jumping :/ 
I'll keep looking a way to improve it.

isacoder:
Hello @atomworks gave another attempt to prevent jumping with loading.
I think the engineer that uses this UI library can add a min-width based on design.
I have added an optional minWidth prop
Here is an example

[Video of interaction with min-width](https://drive.google.com/file/d/1tKxagzQVw7xy18gNLy4dmLF5gRK-nJYJ/view?usp=sharing)

src/Tables/atoms/TypeTableRow.tsx
isacoder: Deleted as per conversation

src/Tables/molecules/TypeTable.tsx
isacoder: updated

atomworks (approved):
👍

storybook/src/stories/Tables/molecules/LoadingTable.stories.tsx
atomworks: Ahhh, of course it is! Thanks 👍

src/Tables/molecules/TypeTable.tsx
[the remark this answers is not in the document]
atomworks: I think because it's a default export (with no name) you can name in whatever you want after it's imported.

PR #45, opened 2021-03-11
Feature/editable cell

isacoder:
### Requirement

[Bugherd#37](https://www.bugherd.com/projects/216308/tasks/37)
[Table] Add editable cell template

### Implementation

The Code of story does not have interaction 😞  perhaps it would be better to move this feature to the main TypeTable 🤔 .
Please review the functionality and if you agree I would move this functionality to the TypeTable story.

Because the information of having it separated is only in the Code where you can see how to:
 - Send the default value as a state that you will update in your data
 - Use the rowKey/id to send in the callback
 - Work with the callback functionality to update data.

Main interactions
- Click on the pencil to edit ( this was done with design consistency in mind however I think it would be better if the whole cell was clickable for edit like excel)
- Saving can be trigger by clicking on the button or by pressing ENTER key
- Cancel can be trigger by clicking on the button, pressing ESC key o clicking outside of the edit field

### Screenshoots

Screen Shot 2021-03-11 at 18 05 57

Screen Shot 2021-03-11 at 18 06 04

Here is a Video of the interaction

[Edit interaction](https://drive.google.com/file/d/1_S-EI52cyQYE7_j3Fyo1bmBl_th_UH0c/view?usp=sharing)


atomworks (changes requested):
Great work, you've managed to get the UI styled really nicely for this.

I have a few pointers to really wrap things up nicely on this if you're able to oblige.

- [x] When going between the value and edit field can we offset the input box so that the text stays in the same place?
- [ ] Set the edit icon weight to 'light' to match the weight used on the copy icon.
- [ ] Make sure the mouse cursor stays normal when hovering over the column name when we aren't editing (at the moment it shows as the editable text cursor).
- [ ] When we're not editing the field can we have an optional prop to pass a link and make that clickable? 
- [ ] In the story, can we use actions to show the callback states working on save?
- [ ] Also, bonus points in the story if you can use some timeouts to emulate the component making an API call (so we can see the saving state)?
- [ ] Perhaps related to the above, the saving state doesn't seem to reset after I hit save in the story. Presumably that's something that's handled in the story and not a bug in the component.

For links as we use React Router (which I am clueless on), I don't know if we need something other than just an anchor link so best to check with @JoshLipps  if you're unsure. 

I'll look forward to start using this pattern in my designs :)

src/Tables/index.ts
```diff
@@ -26,11 +27,14 @@ export interface ITableColumnConfig {
   hasCopyButton?: boolean
   width?: number // TODO: This!
 }
+
+export type IStatus = 'caution' | 'danger' | 'folder' | 'good' | 'neutral';
```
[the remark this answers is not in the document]
isacoder: The only two places I found these values are related to the table.

Screen Shot 2021-03-18 at 22 25 23

Both Components that required have access to it.

I have updated the code with this relation in any case to avoid updating one without updating the other.

Screen Shot 2021-03-18 at 22 24 07

However if you prefer it to be even higher, I can update it and put it in the general index.

isacoder:
**Hello Leonard I have made the requested updates, just missing two the Link that I will review with Josh and the cursor.**

> Make sure the mouse cursor stays normal when hovering over the column name when we aren't editing (at the moment it shows as the editable text cursor).

 I might not understand this requirement. If you mean this cursor?
image

All fields in the table show it because the user is able to select and copy paste the content.
Not sure if you want to remove that feature for all the table.
Personally I think trying to copy data is handy even when you don't have the fast copy button.
[Cursor - Mozzila](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor)

In any case if you prefer we can talk about it over the phone.

**Note aside:**
> In the story, can we use actions to show the callback states working on save?
I think you might have missed it but I created the atom story with this states, that option is available in the atom. Do you want to show it in the EditTable story too?

 > Also, bonus points in the story if you can use some timeouts to emulate the component making an API call (so we can see the saving state)?
 > Perhaps related to the above, the saving state doesn't seem to reset after I hit save in the story. Presumably that's something that's handled in the story and not a bug in the component.
 
 I did the sleep functionality to show the saving but was not working properly that is why I created the atom story 😓.
This is also de reason of the never ending saving. Now it's fixed.

src/helpers/index.tsx
```diff
@@ -0,0 +1 @@
+export const sleep = (delay: number) => new Promise((resolve) => setTimeout(resolve, delay));
```
atomworks, line 1: Because of how we're using this in Storybook examples I think it would be good to keep this helper localised to there and not available in the main UI Kit exports. Mostly because I'm worried that it might get misused in production code. Not a high chance but still a chance!

atomworks (changes requested):
Couple of small requests related to code and one other request if it's quick.

Could we add an example of the version with a link to the `EditableTable.stories.tsx`?

atomworks (approved):
Posted message in wrong PR. Ignore this.

atomworks (changes requested):
Sorry, I posted a comment on the wrong PR and I can't delete it so I had to reset to re-request changes. See the above comments for details.

src/helpers/index.tsx
isacoder: Understood, I will move it to be just on Storybook.

isacoder:
All changes requested have been applied.

- EditableTable.stories.tsx have the editable cell with link.
- Sleep on storybook
- IStatus updated to IDeviceStatus

atomworks (approved):

PR #46, opened 2021-03-12
Fix  disabled background and SubArea Header

isacoder:
### Requirements
[Bugherd25](https://www.bugherd.com/projects/216308/tasks/25)
For "Standard Button", "Disabled" works.
But for “With Icon” and “With Loading”, we still can push the button even “Disabled” is checked.

[Bugherd39](https://www.bugherd.com/projects/216308/tasks/39)
[Page Header] Allow setting a subarea title without needing a url. 

### Implementation
**Bugherd35 Disabled buttons**
:active was not working properly when the button was wrapping items.
:active:enable verifies that is not disabled

**Bugherd39**
I created a link and a non link where it doesn't have a url.

### Screenshots
**Bugherd35 Disabled buttons video**
[InteractiveVideo](https://drive.google.com/file/d/14uYQJ5d-UFMS15jBTPUFUypfWs85lpCr/view?usp=sharing)

**Bugherd39 Page header** 
Screen Shot 2021-03-12 at 18 35 15

### DIFF
--- a/src/Form/atoms/Button.tsx
+++ b/src/Form/atoms/Button.tsx
@@ -26,7 +26,7 @@ const StyledButton = styled.button<IStyledComponentProps>`
       background: ${theme.styles.form.button[design].hover.backgroundColor};
     }
 
-    &:active {
+    &:active:enabled {
       background: ${theme.styles.form.button[design].active.backgroundColor};
     }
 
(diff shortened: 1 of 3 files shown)


atomworks (approved):
Approved from my side... adding you to review my quick changes to fix the styling side which I've needed to do for a while (because of the theme integration issues)

isacoder:
> Approved from my side... adding you to review my quick changes to fix the styling side which I've needed to do for a while (because of the theme integration issues)

✅ Code review
✅  functional review 

I like that you added the not allowed cursor 😄
All looks good to me.

PR #47, opened 2021-03-17
Feature/table sort

isacoder:
### Requirements

[Bugherd#41](https://www.bugherd.com/projects/216308/tasks/41)

[Table] Add sortable headers.

### Implementation

After the informal review we had of the draft implementation on this feature.
I focused on adding the sortConfig values directly on columnsConfig
and made all including sortable optional.

### Screenshots
Sort
Screen Shot 2021-03-17 at 19 54 41

Hover sort
Screen Shot 2021-03-17 at 15 45 18

TableHeaderTitle

Screen Shot 2021-03-17 at 20 45 46


src/Tables/molecules/TypeTable.tsx
```diff
@@ -49,29 +48,55 @@ interface IProps {
   hasStatus?: boolean
   hasThumbnail?: boolean
   hasTypeIcon?: boolean
+  initialSortColumnId?: boolean
   selectCallback? : (checked:boolean, id?: string | number)=>void
   toggleAllCallback? : (checked: boolean)=>void
+  sortCallback? : (ascending: boolean, columnId?: string) => void
```
isacoder, line 53: I updated the sortCallback with columnId as optional as we talked about it.
Because of that I'm handling the column toggles with the index value of the columns.

isacoder (commented):
I have added some comments in the code to help with the review

src/Tables/molecules/TypeTable.tsx
```diff
@@ -51,27 +50,54 @@ interface IProps {
...
+        col.sortActive = false;
+      }
+    });
+    const newAscending = undefined ? true: !sortSpec[columnKey].ascending;
+    updatedSort[columnKey].ascending = newAscending
+    const colId = (updatedSort[columnKey].columnId === undefined) ?  '' : updatedSort[columnKey].columnId;
+    sortCallback(newAscending, colId);
+    setSortSpec(updatedSort);
+  },[sortSpec])
+
```
isacoder: Like I mention before the handling is with the index key

src/Tables/molecules/TypeTable.tsx
```diff
@@ -51,27 +50,54 @@ interface IProps {
...
+      } else {
+        col.sortActive = false;
+      }
+    });
+    const newAscending = undefined ? true: !sortSpec[indexKey].ascending;
+    updatedSort[indexKey].ascending = newAscending
+    const colId = (updatedSort[indexKey].columnId === undefined) ?  '' : updatedSort[indexKey].columnId;
+    sortCallback(newAscending, colId);
+    setSortSpec(updatedSort);
+  },[sortSpec])
```
isacoder, line 99: Like I mention before the handling the toggling with the index key

src/LineUI/LineUI.tsx
```diff
@@ -110,8 +110,8 @@ const LineUI: React.FC<LineUIProps> = ({
     }
 
     const { naturalHeight, naturalWidth, clientHeight } = imgRef.current;
-    const h = fixedImgDimensions?.y ?? naturalHeight;
-    const w = fixedImgDimensions?.x ?? naturalWidth;
+    const h = fixedImgDimensions?.y ? fixedImgDimensions?.y : naturalHeight;
+    const w = fixedImgDimensions?.x ? fixedImgDimensions?.y : naturalWidth;
```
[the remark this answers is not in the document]
JoshLipps: Fixed on main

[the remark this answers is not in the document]
JoshLipps: also instead of `fixedImgDimensions?.y ? fixedImgDimensions?.y : naturalHeight`
we could `fixedImgDimensions?.y || naturalHeight`

src/Tables/molecules/TypeTable.tsx
```diff
@@ -80,10 +106,21 @@ const TypeTable : React.FC<IProps> = ({
           {hasStatus ? <HeaderItem fixedWidth={10} /> : null}
           {hasThumbnail ? <HeaderItem fixedWidth={70} /> : null}
           {hasTypeIcon ? <HeaderItem fixedWidth={35} /> : null}
-
           {columnConfig.map((column, key) => {
-            const {alignment, header, hasCopyButton} = column;
-            return <HeaderItem key={key} alignment={alignment} hasCopyButton={hasCopyButton}>{header}</HeaderItem>;
+            const {alignment, hasCopyButton, sortActive, ascending } : ITableColumnConfig = column;
+            return <HeaderItem
+                      key={key}
```
JoshLipps, line 112: IMO this key should use `columnId` by default and fall back to index for columnID

src/Tables/molecules/TypeTable.tsx
```diff
@@ -80,10 +106,21 @@ const TypeTable : React.FC<IProps> = ({
...
+                      alignment={alignment}
+                      hasCopyButton={hasCopyButton}
+                      >
+                        <TableHeaderTitle
+                          {...column}
+                          columnKey={key}
```
JoshLipps, line 118: Can we rename this to `columnId` to match the other interface?

src/LineUI/LineUI.tsx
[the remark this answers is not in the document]
isacoder: Thanks! I will merge and put it back to verify storybook runs well.

src/Tables/molecules/TypeTable.tsx
isacoder: I talked about it in the code notes I left.
Since you mention that the UI shouldn't rely on getting the columnId. I'm actually not using it.
This is the indexKey.

I could rename it to indexKey to avoid confusion.

src/Tables/molecules/TypeTable.tsx
isacoder: I'm not sure, that is the part I really struggled with.
Like, since columnId is optional, can some columns have it and some don't?
Can this cause handling problems?


What I understood from our last conversation was that the UI Library should not work around the Id's that expects from the outside. That is why I went with index as Id's at the end for the management of the UI and made optional to return the columnId in case the it was sent because I'm just a messenger of that data 🚚

src/LineUI/LineUI.tsx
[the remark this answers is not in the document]
isacoder: Merged! 👯

src/Tables/atoms/TableHeaderTitle.tsx
```diff
@@ -0,0 +1,83 @@
...
+  ascending,
+  columnKey,
+  toggleSort
+}) => {
+  return(
+      <HeaderTitle
+        sortable={sortable}
+        isSortActive={isSortActive}
+        ascending={ascending}
+        onClick= {() => {toggleSort(columnKey)}}
```
JoshLipps, line 75: Minor: should this be a useCallback.

src/Tables/molecules/TypeTable.tsx
```diff
@@ -49,29 +48,55 @@ interface IProps {
...
+    updatedSort.forEach((col, key) => {
+      if(key === columnKey) {
+        col.sortActive = true;
+      } else {
+        col.sortActive = false;
+      }
+    });
+    const newAscending = undefined ? true: !sortSpec[columnKey].ascending;
+    updatedSort[columnKey].ascending = newAscending
+    const colId = (updatedSort[columnKey].columnId === undefined) ?  '' : updatedSort[columnKey].columnId;
```
JoshLipps: IMO this is a bit strange. If the the `colId` should either be the index or the user provided ID, how can the user know which column has changes otherwise.

src/Tables/atoms/TableHeaderTitle.tsx
isacoder: I might have misunderstood the question but this toggleSort is on a useCallback.

toggleSort here is a executed prop. This function is wrapped in a useCallback handled by the parent component TypeTable.

Screen Shot 2021-03-26 at 16 53 16 1

src/Tables/molecules/TypeTable.tsx
isacoder: #### I agree that it looks strange.

The only reason I implemented like this is to allow columnId to be optional. ( we talk that columnId shouldn't be required if sorting is not required in the table being the main reason to prevent breaking previous implementations.)

**The only scenario** where I think the UI library user would use sorting and would be too lazy to provide a columnId  is if the table has only **one sorting column**. In that case the user would know what is sorting criteria.

If you want this part to be refactor I can think of two other ways to implement it.

#### **Option A** 
Each Column has it's own sortingCallback instead one sorting handler and the user would know what to sort because each implementation it's independent.

```js
const sortByName = (ascending: boolean) => { awesomeData being sorted by name}
const sortByUploadTime = (ascending: boolean) => {  awesomeData being sorted by uploadTime}

const columnConfigSample : ITableColumnConfig[] = [
  {
    columnId: 'deviceName',
    header: 'Device Name',
    sortable: true,
    sortActive: true,
    ascending: true,
    cellStyle: 'firstColumn',
   sortCallback: sortByName
  },
  {
    columnId: 'status',
    header: 'Status',
    sortable: true,
    showStatus: true,
    cellStyle: 'normalImportance',
    sortCallback: sortByUploadTime,
  }
]
```

#### Option B 
Adding a sorting type kind of the one that we talk. ISortConfig would be optional but if you use it columnId is required.

```js

type ISortConfig = {
  columnId: string,
  active?: boolean,
   isAscending?: boolean
}
```

The User would still send the sortCallback in the main props and the column config would look like this.

```js
const columnConfigSample : ITableColumnConfig[] = [
  {
    header: 'Device Name',
    cellStyle: 'firstColumn',
    sortConfig: {columnId:"deviceName", active: true, isAscending: true}
  },
  {
    header: 'Status',
    showStatus: true,
    cellStyle: 'normalImportance',
  }
]
```
Sorry for the long message in advance.
Since this will be used in many places in the future and you are familiar with the use cases I think it was worth it to discuss it more.

atomworks (approved):
If @JoshLipps is happy with any code changes requested, I'm happy with the code.

I have one usability related change I'd like in regards to keep some consistent behaviour over sorting order. At the moment if you are sorted by ascending and you click to another column and click back, it will then be toggled to descending. 

I think the best option for consistency will be to keep ascending or descending as a higher level thing. So if you're sorting by name ascending and click cost, then it would be cost (ascending). So asc / desc basically persists across columns choices.

If that's simple and can be done quickly here before a release, let's do it. If it's going to hold things up, let me know and then we can put it in a follow up BugHerd ticket. What do you think?

src/Tables/atoms/TableHeaderTitle.tsx
JoshLipps: there is a implicit function declaration there. that is all i meant.

src/Tables/molecules/TypeTable.tsx
JoshLipps: I prefer you always callback with a column ID. In the case where a  is provided use that, in the case where no column id is provided by user fallback for an index based column ID, something like `column_${index}` for the default columnID might help with clarity. 


IMO you should/could set that default `columnID` once at the initial  layer and use it everywhere (not just here)

src/Tables/atoms/TableHeaderTitle.tsx
isacoder: Yes this is an implicit declaration I will update it to explicit if you prefer.
However I will not add a useCallback to it because that creates a very weird behavior of only re-rendering only if another HeaderTitle is sorted.

isacoder:
Hello @JoshLipps and @atomworks thank you for all the suggestions in this PR 😃 
I hope this is closer to the expectations but please don't hesitate in recommend more changes.

I have updated this PR with the next.

- TableHeaderTitle toggleSort is on a  explicit function
- sortCallback returns columnId that if not provided by the user it will send column_${index} string.
- I updated the the ascending handler to a higher level as @atomworks
-- Optional defaultAscending prop
-- The data of defaultExampleData in the story is sorted to be consistent with the flag of active **ascending true** sent in the configuration.
-- Hovering on a different column that the current sorted will show now the same sort that the current.
-- The ascending flag will only toggle if the sort of that column is active.

JoshLipps (approved):
LGTM

PR #48, opened 2021-03-19
Feature: Add usePoll setTimout based hook

JoshLipps:


isacoder:
Sorry for my delay in reviewing this.
I think you have already updated the useClickOutside  in a different PR, in any case I verified that this one and usePoll is available from the 'scorer-ui-kit' import.

I used the usePoll just to understand how it worked.
From my understanding it runs the callback in a chained manner when it renders and finishes on unmount.
Since there is not a cancel/stop option. I would recommend to comment this for Documentation.

 ✅ functionality
 ✅ code review

isacoder (approved):

PR #49, opened 2021-03-22
Tag element and list of tags

isacoder:
### Request
Tags
[Tags] Add tags UI element to UI kit
[Bugherd#45](https://www.bugherd.com/projects/216308/tasks/45)

[Functionality specifics the tag](https://app.slack.com/docs/T0ALXERMZ/F01R5DCL0DB?origin_team=T0ALXERMZ&origin_channel=DMLC36D7T)

### Implementation
This PR covers 3 new components

- Tag
- TagLink
- TagList

**Tag**
I added the size feature taking it from the icon size.
however I didn't made the padding proportional.
This could be easier with EM's I think but would need to have at least a second reference to calculate the right EM.
Since you mention this could be change in the feature didn't put to much time on this.

**TagLink**
It's just a tag wrapped with a Link of react router with the "to" option like we talked.

**TagList**
Just the list in the story I showed that colors can be different for every tag.

### Screenshoots
**Tag**
Screen Shot 2021-03-25 at 22 12 51

**TagLink**
Screen Shot 2021-03-25 at 22 12 31

**TagList**
Screen Shot 2021-03-25 at 22 11 45


atomworks (changes requested):
Sorry this one has taken so long to give feedback for. 

There's a couple of change requests here when you're able to take a look.

- [ ] We don't need to provide any alternative colors. The default one is great.
- [ ] For the TagLink, when hovering can we use the primary color?
- [ ] Would it be possible to move the link functionality of TagLink in to the main Tag component and conditionally enabled the link behaviour (like enableHover) if a `to` prop is provided?

storybook/src/stories/Misc/atoms/Tag.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+  for(const key in IconSVGs){
+    iconList[key] = key;
+  }
+
+  return iconList;
+};
+
+const Container = styled.div``;
+
+const colorOptions = { EmtpyValue: "", Mono: "mono", Dimmed: "dimmed", Subtle: "subtle", Inverse: "inverse", Primary: "primary", Danger : "danger" };
```
atomworks, line 24: Typo on empty!

isacoder: Removed

isacoder:
I have updated the basic colors as the design on zeppling.
I realized it wasn't in the PR before.
The colors are not the basic colors of SVG.

I want to confirm that this is correct. Sorry for not asking before.

https://zpl.io/agxkp51
zpl://screen?sid=604edad8c702888f4a043947&pid=5ee9f5a23a30524bf9e5b2a9

atomworks:
> I have updated the basic colors as the design on zeppling.
> I realized it wasn't in the PR before.
> The colors are not the basic colors of SVG.
> 
> I want to confirm that this is correct. Sorry for not asking before.
> 
> https://zpl.io/agxkp51
> zpl://screen?sid=604edad8c702888f4a043947&pid=5ee9f5a23a30524bf9e5b2a9

Sorry, I totally missed this. Yes, those should be correct.

atomworks (approved):

PR #52, opened 2021-03-29
Prop to hide logout link in drawer

atomworks:
This is required on some Edge side projects where this isn't relevant. 

I originally wanted to do conditional hiding based on if `logoutLink` or `onLogout` were passed as props but as we set a default for onLogout I wasn't sure how to detect that as an empty function. If you can think of a better was to refactor this, happy to do that. Otherwise I've just gone with a simple boolean to show and hide.

### DIFF
--- a/src/Global/TopBar.tsx
+++ b/src/Global/TopBar.tsx
@@ -213,6 +213,7 @@ interface IProps {
   logoutLink?: string;
   loggedInUser: string;
   useSearch?: boolean;
+  useLogout?: boolean;
   searchPlaceholder?: string;
   showLanguage?: boolean;
   userDrawerBespoke?: ReactElement;
@@ -224,6 +225,7 @@ const TopBar : React.FC<IProps> = ({
   useNotifications = false,
   logoutLink = '/logout',
   useSearch = false,
+  useLogout = true,
   searchPlaceholder = 'Search for devices, analysis tasks, etc.',
   userSubmenu = [],
   userDrawerBespoke,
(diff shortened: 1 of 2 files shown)


isacoder:
> This is required on some Edge side projects where this isn't relevant.
> 
> I originally wanted to do conditional hiding based on if `logoutLink` or `onLogout` were passed as props but as we set a default for onLogout I wasn't sure how to detect that as an empty function. If you can think of a better was to refactor this, happy to do that. Otherwise I've just gone with a simple boolean to show and hide.

Hello Leonard!
I don't mind to have a prop to explicitly show, hide that element.
However if you want to go the other way you could do it this way. 

```js
const TopBar : React.FC<IProps> = ({
  logoutLink,
  onLogout,
}) => {
```
Then you would be able to conditionally verify that you have the required props to render this.

```js
          {(logoutLink &&  onLogout) ?
            <Logout>
              <LinkMenu>
                <LinkMenuItem><LinkMenuItemA onClick={logoutHandler} href={logoutLink}>Logout</LinkMenuItemA></LinkMenuItem>
              </LinkMenu>
            </Logout>
          : null }
```

and of course this would required to verify that in the logout handler to avoid warnings

```js
  const logoutHandler = useCallback(async(e: React.MouseEvent<HTMLAnchorElement, MouseEvent>)=>{
      e.preventDefault();
   if(!logoutLink || !onLogout) {return;}

    await onLogout();
    window.location.assign(logoutLink);
  },[logoutLink, onLogout]);
```

isacoder (changes requested):
✅ Functionality 
⚔️ code

I tried the functionality of this change in storybook and worked as expected.
However I notice that the props names are kind of weird.

useSearch
useLogout
useNotification

"useSometing" is the naming for hooks in react.

This props are boolean to add or remove elements.
I think a better pattern we have been using in the UI kit is "hasSomething"
I suggest to be updated to 

hasSearch
hasLogout
hasNotifications

atomworks:
As discussed, it's better to keep the show / hide as a boolean value prop for use in potential future use cases. However I've updated the naming of the show / hide related props to drop the word `use` as a prefix and use `show` to avoid the name pattern clashing with React hook naming conventions.

atomworks:
Just tweaked it again as per your earlier comment @isacoder. It's now consistent with other components prop names

isacoder (approved):
✅ Code
✅ Functionality

Thanks for the update!

PR #53, opened 2021-03-31
LineUI - Add boundaryOffset option

fti-dshelke:
Add boundryOffset in LineUI to prevent user to put the vertices of quadrangles/line near the border. This feature is required if image used in LineUI is undistorted.

### DIFF
--- a/example/src/pages/LinePage.tsx
+++ b/example/src/pages/LinePage.tsx
@@ -13,6 +13,7 @@ import {
   Button,
   TextField,
   Label,
+  Input
 } from 'scorer-ui-kit';
 import { LineUIOptions } from '../../../dist/LineUI';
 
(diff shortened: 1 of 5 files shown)


JoshLipps (commented):
Looks great. Can you also update `LineUIRTC` and `LineUIVideo` components with these changes as well?

atomworks (approved):
This looks great, thanks for doing this.

In addition to Josh's comments, if possible, could you add an example of this to the examples folder? It should be easy to do, just clone a page like `example/src/pages/LinePage.tsx`, add it to the router in `example/src/App.tsx` and the link to `example/src/pages/LinksPage.tsx` too. Of course adding an example of your boundary offset the cloned page too.

If you need to get this merged more urgently, we can follow that up in a different PR. I've marked this approved but I would appreciate if you can do that though I understand if there is some time sensitivity here.

atomworks (commented):
I've changed this to "Comment" as I think pre-approving like that is problematic! Sorry for the confusion.

fti-dshelke:
Requested changes are done. Please review.

JoshLipps (approved):

PR #54, opened 2021-04-01
Feature/slider

isacoder:
### Requirements
[Burgherd#46](https://www.bugherd.com/projects/216308/tasks/46
[Input] Value slider component

**Designs**
https://zpl.io/a7jrv1Y
https://zpl.io/2jNkzkO

### Implementation

I made some assumptions based on the design like the colors.
I saw in a comment on zeppelin that they were supposed based in the AlertBar.

Most of the functionality is based in the [input range](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) definition

3 components with stories were made.

SliderInput, the base slider.
DurationSlider
PercentageSlider

I think there is area of improvement in the Duration Slider but I preferred to keep it flexible over giving more functionality without knowing the specifics of the cases.

And the biggest area of want is in the flickering 😞 
I didn't put a lot of time on trying to fix this, perhaps a debounce in the input would be helpful. 
I committed to have it functional first then improve it.

### Screenshoots

Screen Shot 2021-04-01 at 22 49 47

Screen Shot 2021-04-01 at 22 48 25

Screen Shot 2021-04-01 at 22 48 36

Screen Shot 2021-04-01 at 22 48 44


src/Form/atoms/SliderInput.tsx
```diff
@@ -0,0 +1,282 @@
...
+    if(inputCallback) {
+      inputCallback(selectedValue);
+    }
+    setThumbValue(thumbLeftPostion(selectedValue, minValid, maxValid));
+  },[selectedValue]);
+
+  return(
+    <SliderWrapper disabled={disabled}>
+      <Rail>
+      </Rail>
```
JoshLipps, line 251: self close please

src/Form/atoms/SliderInput.tsx
```diff
@@ -0,0 +1,282 @@
...
+            bgColor={thumbColor}
+          />
+        <Thumb
+            data-value={selectedValue}
+            leftValue={thumbValue}
+            data-percentage={thumbValue}
+            bgColor={thumbColor}
+        />
+      </ThumbWrapper>
+      <HiddenInput
```
JoshLipps, line 267: Nice

src/Form/atoms/SliderInput.tsx
isacoder: Updated :D

JoshLipps (approved):
LGTM

PR #55, opened 2021-04-05
Remove min-height and add blur to backdrop in modal

isacoder:
### Requirement

1.- Some designs have less than 300px height and current implementation doesn't allow to implement those.
2.- @atomworks mention that it would be great that the backdrop has blur.

### Implementation

Reviewed a couple of options fr the min-height with @atomworks and we decided just removing it will be better for now.

### Screenshots 

New Empty Modal
Screen Shot 2021-04-05 at 12 59 12

Previous templates, since content was bigger than 300px won't show any  difference
Screen Shot 2021-04-05 at 15 19 17

### DIFF
--- a/src/Alerts/atom/Modal.tsx
+++ b/src/Alerts/atom/Modal.tsx
@@ -61,7 +61,6 @@ const CloseButton = styled.button<{ selected?: boolean }>`
 
 const LightBox = styled.div<{ padding?: boolean, width?: string}>`
   position: relative;
-  min-height: 300px;
   margin: 27px 0 0;
   box-shadow: 0 10px 15px 0 hsla(205, 42%, 60%, 0.15);
   background-color: hsl(0, 0%, 100%);
--- a/src/Alerts/atom/Modal.tsx
+++ b/src/Alerts/atom/Modal.tsx
@@ -16,7 +16,10 @@ const Container = styled.div`
   align-items: center;
   justify-content: center;
   background-color: hsl(0, 0%, 0%, 0.2);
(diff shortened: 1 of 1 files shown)


src/Alerts/atom/Modal.tsx
```diff
@@ -16,7 +16,10 @@ const Container = styled.div`
   align-items: center;
   justify-content: center;
   background-color: hsl(0, 0%, 0%, 0.2);
+  -webkit-backdrop-filter: blur(5px);
+  backdrop-filter: blur(5px);
```
atomworks, line 20: This is way easier than the method we had to use before! I'm glad this has better browser support at last! 🥳

src/Alerts/atom/Modal.tsx
```diff
@@ -16,7 +16,10 @@ const Container = styled.div`
   align-items: center;
   justify-content: center;
   background-color: hsl(0, 0%, 0%, 0.2);
```
atomworks, line 18: I wasn't sure about this when we added the initial version but this color is incorrect. It's more of a grey with a blueish hue. Zeplin says: `hsla(202, 33%, 95%, 0.51)`

atomworks (changes requested):
Looks great. One very quick semi-related request though if you can! Just a color tweak.

src/Alerts/atom/Modal.tsx
isacoder: Thank you for making this correction.
I have updated.

atomworks (approved):
Looks great, thanks!

PR #58, opened 2021-04-08
Update Duration and Percentage Slider Titles to use Label component

isacoder:
### Requirements
This is fix request from the FTI team

```
For other input component label  font-weight in ui-kit is 500, but in slider component doesn't have font-weight for the label. Is there any way so we can modify the styles of slider label? (Please check the attached snapshot)
```

### Implementation
I had not added the titles as Label component's the first time.

### Screenshots
Screen Shot 2021-04-08 at 18 15 37

### DIFF
--- a/src/Form/atoms/Label.tsx
+++ b/src/Form/atoms/Label.tsx
@@ -16,7 +16,7 @@ const LabelText = styled.span`
 `;
 
 interface OwnProps {
-  htmlFor?: string
+  htmlFor: string
   labelText: string
 }
 type Props = OwnProps & LabelHTMLAttributes<HTMLLabelElement>
--- a/src/Form/atoms/Label.tsx
+++ b/src/Form/atoms/Label.tsx
@@ -16,16 +16,18 @@ const LabelText = styled.span`
 `;
 
 interface OwnProps {
-  htmlFor: string
+  htmlFor?: string
   labelText: string
 }
 type Props = OwnProps & LabelHTMLAttributes<HTMLLabelElement>
 
 const Label : React.FC<Props> = ({ htmlFor, labelText, children, ...props }) => {
-  return <StyledLabel htmlFor={htmlFor} {...props}>
-    <LabelText>{labelText}</LabelText>
-    {children}
-  </StyledLabel>;
+  return (
+    <StyledLabel htmlFor={htmlFor} {...props}>
+      <LabelText>{labelText}</LabelText>
+      {children}
+    </StyledLabel>
+  );
 };
 
 export default Label;
--- a/src/Form/molecules/DurationSlider.tsx
+++ b/src/Form/molecules/DurationSlider.tsx
@@ -3,6 +3,7 @@ import styled from 'styled-components';
 import {ITimeUnit} from '../../index';
 import SliderInput, {ISlider} from '../atoms/SliderInput';
 import {getTextTimeUnit} from '../../helpers';
+import Label from '../atoms/Label';
 
 
 /**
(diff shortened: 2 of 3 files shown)


src/Form/atoms/Label.tsx
```diff
@@ -16,16 +16,18 @@ const LabelText = styled.span`
...
-  return <StyledLabel htmlFor={htmlFor} {...props}>
-    <LabelText>{labelText}</LabelText>
-    {children}
-  </StyledLabel>;
+  return (
+    <StyledLabel htmlFor={htmlFor} {...props}>
+      <LabelText>{labelText}</LabelText>
+      {children}
+    </StyledLabel>
+  );
```
atomworks, line 30: What's the relevance of adding the brackets to the return here?

src/Form/atoms/Label.tsx
```diff
@@ -16,16 +16,18 @@ const LabelText = styled.span`
 `;
 
 interface OwnProps {
-  htmlFor: string
+  htmlFor?: string
```
atomworks, line 19: Is there no way to make this change without moving this down to optional? As this is a common basic usability attribute for use in forms, it's better if we try and keep that enforced consistently which is why it was chosen to be mandatory in the original development of this component.

I know the percentage slider isn't a standard UI component but there should still be some basic associated we could do in this context I believe.

atomworks (commented):
Just got a couple of questions over some decisions here that I'd like to get some of your thinking on before we proceed.

src/Form/atoms/Label.tsx
isacoder: eslint complains about tags not being aligned.

you could say cosmetic/readability improvement 🤔

src/Form/atoms/Label.tsx
isacoder: ok, so here is my thinking process around it.

htmlFor is same as for label
It receives the id of the input 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

id should be unique.
currently we do not support internally uniqueness in our components. ( uuids)
we use index sometimes when we map thinks but generally we have an area of opportunity here. 

In the zeplin designs I noted that it would be expected to have more than one slider component in a page.
That means that creating a static value for the htmlFor like 'percentage-slider' will not be posible since both inputs will have the same id, being this a bad practice. At implementation time I decided that the simplest way to solve it is to leave it option which made sense to me because maybe sometimes we want to use a label without having an input. 

Screen Shot 2021-04-13 at 17 48 22

If you want this change and be always required, there are two ways to handle this :

1) Pass to the htmlFor a static value like 'percentage-slider'. We would just ignore that this goes against the standard.

2) Have the htmlFor required in the sliders. Then the UI Kit Developer would need to always assign that value and care that is unique.  This will brake current usage of the slider. However since it's being used for the first time the change might not be that much intrusive.

Please let me know what would you prefer.

src/Form/atoms/Label.tsx
atomworks: I've not seen that error before. Perhaps my linter isn't setup properly... I have noticed some of my code being fixed recently for that reason so I suspect I may need to check in to that. Thanks!

src/Form/atoms/Label.tsx
atomworks: Thanks for the full answer here, I appreciate you taking that time. We can probably chat more about this on a call later today but I wanted to share some notes here as I'm thinking about it now.

My instinct here is that we want the label to require the htmlFor attribute because for most standard form elements it's better to keep that convention moving forward. This slider though is the outlier and while it doesn't make sense to have it here in the current iteration, I would rather avoid this forcing the decision on other elements.

I think we need to look more precisely at how we're doing forms anyway so I think there's a long term part to this ultimately. Short term though, perhaps there is a simple way to provide a unique ID maybe. Let's chat about this more

isacoder: Update from the phone call.
We choose was Option 1.

isacoder:
I did the preview with multiple sliders to verify it didn't had a problem with the same htmlFor and id's and worked fine.

Screen Shot 2021-04-23 at 15 59 37

atomworks (approved):

PR #59, opened 2021-04-12
Add extra-small button with loading and icon update

isacoder:
### Requirements
[Bugherd#40](https://www.bugherd.com/projects/216308/tasks/40)

[Buttons] Add extra small variant.

### Implementation
Made added the extraSmall style by hand, not sure if this will complate later the auto generated one of zepplin.
Did small refactor to have button padding in dimensions file.

### Screenshots 

**Extra small base**
Screen Shot 2021-04-12 at 20 05 51

**Extra small with icon**
Screen Shot 2021-04-12 at 20 05 22

**Extra small with loading**
Screen Shot 2021-04-12 at 20 04 37

### DIFF
--- a/src/Form/atoms/ButtonWithIcon.tsx
+++ b/src/Form/atoms/ButtonWithIcon.tsx
@@ -3,10 +3,10 @@ import styled, { css } from 'styled-components';
 
 import Button from './Button';
 import Icon from '../../Icons/Icon';
-import { IButtonProps } from '..';
+import { IButtonProps, TypeButtonSizes } from '..';
 
 
-const TextContainer = styled.div`
+const TextContainer = styled.div<{size:TypeButtonSizes}>`
   height: inherit;
   flex: 1;
   order: 1;
@@ -14,8 +14,9 @@ const TextContainer = styled.div`
   justify-content: center;
   align-items: center;
   white-space: nowrap;
-  padding: 0 20px;
-  ${({theme}) => theme && css`
+
+  ${({theme, size}) => theme && css`
+    padding: ${theme.dimensions.form.button.padding[ size ]};
     transition: padding ${theme.animation.speed.slow} ${theme.animation.easing.primary.easeInOut};
   `}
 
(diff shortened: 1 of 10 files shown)


src/themes/common.tsx
```diff
@@ -13,9 +13,18 @@ export const dimensions = {
...
-      normal: '40px',
-      large: '60px'
+      height: {
+        extraSmall: '20px',
+        small: '30px',
+        normal: '40px',
+        large: '60px'
+      },
+      padding: {
+        extraSmall: '3px 10px',
```
atomworks, line 23: We shouldn't need `3px` here as the `line-height` is more reliable to center (and that seems to inherit the right value at the moment). Can we have this as `0` as per the other padding sizes?

src/Form/atoms/ButtonWithIcon.tsx
```diff
@@ -38,37 +39,34 @@ const IconContainer = styled.div<{ position?: string }>`
...
 
 interface IProps extends IButtonProps {
   icon: string
   position?: 'left' | 'right'
 }
 
-const ButtonWithIcon : React.FC<IProps> = ({design, size, onClick, disabled, position, icon, children, ...props}) => {
+const ButtonWithIcon : React.FC<IProps> = ({design, size = 'normal', onClick, disabled, position, icon, children, ...props}) => {
 
   const iconSize = size === 'large' ? 20 : 16;
```
atomworks, line 57: We need to make the icon (and for the loading version too) smaller for the xSmall size. I think `12` is what we want for that. Perhaps this can be refactored and added to the theme dimensions file too.

src/Form/atoms/ButtonWithLoading.tsx
```diff
@@ -3,20 +3,21 @@ import styled, { css } from 'styled-components';
...
 
-const LoadingButton = styled(Button)<{ loading: boolean }>`
-  ${({loading, theme, design}) => loading && css`
+const LoadingButton = styled(Button)<{ loading: boolean, size: TypeButtonSizes }>`
+  ${({loading, theme, size}) => loading && css`
     cursor: wait;
     background: ${ theme.styles.form.button['primary'].active };
-
+    ${(size === 'extraSmall') && css`
+        padding: 3px 20px 3px 10px;
```
atomworks, line 14: This is kind of a strange instance I guess so this doesn't need to be perfect, I think we can refactor later.

If you can switch out the `3px` for `0` though that would be great.

src/Form/atoms/ButtonWithLoading.tsx
```diff
@@ -54,26 +55,32 @@ const LoadingContainer = styled.div<{ design: TypeButtonDesigns, show?: boolean,
...
+      ?  (size === 'extraSmall') ? '-10px' : '-20px'
+      : '0' 
+    };
   ` : css`
-    margin-left: ${ loading === 'true' ? '-20px' : '0' };
+      margin-left: ${ loading === 'true'
+      ? (size === 'extraSmall') ? '-10px' : '-20px'
+      : '0'
+    };
   `}
```
atomworks, line 73: These negative value margin tweaks shouldn't be needed. I had a quick look and it seems the padding that has been added is why you're needing to compensate like this here for the small size buttons so I think once that padding has been addressed it should make a lot (if not all of this) redundant.

atomworks (changes requested):
Couple of `padding` and `margin` change requests on this review and I think one other related refactor request. Hopefully all fairly straight forward!

PR #60, opened 2021-04-13
Feature/line UI area name

fti-dshelke:
Add area name property in lineset data to show it in the center of the line-ui area.

### DIFF
--- a/src/LineUI/LineSet.tsx
+++ b/src/LineUI/LineSet.tsx
@@ -222,7 +222,8 @@ const LineSet : React.FC<ILineSetProps> = ({ getCTM, boundaries, unit, size, lin
     });
 
     midpoint = { x: midpoint.x / pointsLength, y: midpoint.y / pointsLength };
-    return <AreaLabelText fontSize={`${unit * 14}px`} styling={lineSetData.styling || 'primary'}  x={midpoint.x - (4 * unit)} y={midpoint.y + (6 * unit)}>{lineSetData.areaName}</AreaLabelText >
+    const Textlen = lineSetData.areaName?.length || 1;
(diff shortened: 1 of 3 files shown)


src/LineUI/LineSet.tsx
```diff
@@ -203,12 +211,27 @@ const LineSet : React.FC<ILineSetProps> = ({ getCTM, boundaries, unit, size, lin
     />
   );});
 
+  const AreaLabel = () => {
+    const pointsLength = lineSetData.points.length;
+    if (pointsLength < 3) return null;
+    let midpoint = { x: 0, y: 0 };
```
JoshLipps, line 217: const

src/LineUI/LineSet.tsx
```diff
@@ -203,12 +211,27 @@ const LineSet : React.FC<ILineSetProps> = ({ getCTM, boundaries, unit, size, lin
     />
   );});
 
+  const AreaLabel = () => {
```
JoshLipps, line 214: can we move this component out of the LineSet and just pass props for the points and areaName?

src/LineUI/LineSet.tsx
fti-dshelke: We are reassigning value to midpoint

src/LineUI/LineSet.tsx
fti-dshelke: Done

JoshLipps (approved):

src/LineUI/LineSet.tsx
JoshLipps: meh ...IMO this means you could use a reduce?

PR #61, opened 2021-04-13
Add CurrentUser as optional prop in top bar

isacoder:
### Requirements

[Bugherd55](https://www.bugherd.com/projects/216308/tasks/55)

[User Menu] Allow for current user to be hidden (use similar approach to how logout hide works)

### Implementation
I leave it at hasCurrentUser as true by default since it wasn't specified.

### Screenshot

Screen Shot 2021-04-13 at 19 25 25


atomworks (approved):
Great! Thank you.

PR #63, opened 2021-04-14
Update checkbox with checkmark svg

isacoder:
### Requirements
[Bugherd#28](https://www.bugherd.com/projects/216308/tasks/28)
[Checkbox] Change checked icon to checkmark (not X)

### Implementation
This icon is not part of the listed general icons as requested from @atomworks 

### Screenshoots
Screen Shot 2021-04-14 at 20 32 33


atomworks:
I got so used to seeing those X marks in the checkbox so it's strange to see the proper check. Looks much better now though!

atomworks (approved):

PR #66, opened 2021-04-26
Feature/x small button

isacoder:
### Requirements
[Bugherd#40](https://www.bugherd.com/projects/216308/tasks/40)

[Buttons] Add extra small variant.

### Implementation

 - Small refactor with dimensions for height, paddings and iconSize in buttons. 
 
This is the second PR for this component. ([Previous PR](https://github.com/future-standard/scorer-ui-kit/pull/59))
I decided to start new to try without my previous misconceptions of the buttonWithLoading and withIcon.

 
 ### Screenshots
Screen Shot 2021-04-26 at 18 16 44

### DIFF
--- a/src/Form/index.ts
+++ b/src/Form/index.ts
@@ -41,7 +41,7 @@ export {
 
 export type TypeFieldState = 'default' | 'disabled' | 'required' | 'valid' | 'invalid' | 'processing';
 export type TypeButtonDesigns = 'primary' | 'secondary' | 'danger';
-export type TypeButtonSizes = 'xSmall' | 'small' | 'normal' | 'large';
+export type TypeButtonSizes = 'xsmall' | 'small' | 'normal' | 'large';
 
 interface ButtonProps {
   size?: TypeButtonSizes
--- a/src/Form/index.ts
+++ b/src/Form/index.ts
@@ -41,7 +41,7 @@ export {
 
 export type TypeFieldState = 'default' | 'disabled' | 'required' | 'valid' | 'invalid' | 'processing';
 export type TypeButtonDesigns = 'primary' | 'secondary' | 'danger';
-export type TypeButtonSizes = 'small' | 'normal' | 'large';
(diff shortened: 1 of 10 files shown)


src/Form/index.ts
```diff
@@ -41,7 +41,7 @@ export {
 
 export type TypeFieldState = 'default' | 'disabled' | 'required' | 'valid' | 'invalid' | 'processing';
 export type TypeButtonDesigns = 'primary' | 'secondary' | 'danger';
-export type TypeButtonSizes = 'small' | 'normal' | 'large';
+export type TypeButtonSizes = 'xSmall' | 'small' | 'normal' | 'large';
```
atomworks, line 44: Can keep this all lower case? I think `xsmall` is more consistent and matches the `xlarge` that `SpinnerSize` uses.

src/themes/common.tsx
```diff
@@ -13,9 +13,24 @@ export const dimensions = {
...
+        small: '0 20px',
+        normal: '0 20px',
+        large: '0 20px',
+      },
+      iconSize: {
+        xSmall: 12,
+        small: 14,
+        normal: 16,
+        large: 20
+      }
```
atomworks, line 33: I think it might be better to group things by size here so we have for example:
```
...
xsmall: {
  height: `20px`,
  padding: `0 10px`, 
  iconSize: `12`
}
```

I think that group will make this area easier to update with theming tweaks and has the advantage in styled components of being able to output the object which will convert to CSS rather than mapping each property. (although I'm not sure if it's smart enough to filter out `iconSize`).

atomworks (changes requested):
Just a couple of comments more related to structure and naming.

src/themes/common.tsx
isacoder: I have updated the group as requested.

About the theming, I don't think I understand that part. I'm not sure if it's a request or if you are thinking aloud for some update of the future. Please let me know.

atomworks (approved):
Thanks for the super fast revision from that last review! LGTM

PR #68, opened 2021-05-07
Add following icons

sasakisasaki:
* Humidity

* Noise

* Co2

* Pause

* Play


sasakisasaki:
I added some icons which are given by @atomworks .
I hope this PR makes sense.

JoshLipps:
replaced by custom icons solution

PR #69, opened 2021-05-11
Fix checkAll in table and disable hovers in check component

isacoder:
### Requirements
[Burherd#24](https://www.bugherd.com/projects/216308/tasks/24)

Bug: If a table has Select All but no entries then it is marked as checked. It should be uncheck and disabled or maybe just hidden.

### Implementation
I made the review of this on the Loading table of storybook.
Realized that checkbox had hover styling when disabled and modified a little the theme.
I felt a dejavu about it and I'm not sure if @atomworks  told me not to change that before hehehe..

Please review that the disabled check is working as designed.

#### Before Fix Screenshots

AllChecked and enabled with empty loading
Screen Shot 2021-05-11 at 16 22 10

AllChecked and enabled when  background data loading
Screen Shot 2021-05-11 at 16 22 18
#### After Fix Screenshots

Empty AllCheckbox unchecked and disabled with loading
Screen Shot 2021-05-11 at 17 58 58

Data loading AllCheckbox checked disabled
Screen Shot 2021-05-11 at 17 58 45

Data loading Some checked, AllCheckbox not check disabled
Screen Shot 2021-05-11 at 18 18 28

### DIFF
--- a/src/themes/light/styles.js
+++ b/src/themes/light/styles.js
@@ -256,7 +256,8 @@ export const styles = {
             "unchecked": {
                 "disabled": {
                     "backgroundColor": "hsla(0, 0%, 98.8%, 1.000)",
-                    "borderColor": "hsla(210, 20%, 90.2%, 1.000)"
+                    "borderColor": "hsla(210, 20%, 90.2%, 1.000)",
+                    "cursor": "not-allowed",
                 },
                 "default": {
                     "borderColor": "hsla(208, 24.6%, 77.6%, 1.000)"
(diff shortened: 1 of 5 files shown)


src/themes/light/styles.js
```diff
@@ -268,7 +269,8 @@ export const styles = {
             "checked": {
                 "disabled": {
                     "boxShadow": "inset 0px 1px 5px 0px hsla(205, 50.3%, 30%, 0.051)",
-                    "backgroundColor": "hsla(210, 20%, 90.2%, 1.000)"
+                    "backgroundColor": "hsla(210, 20%, 90.2%, 1.000)",
+                    "cursor": "not-allowed",
```
atomworks, line 273: As the theme is generated by Sketch and this is something I can't represent for output without some heavier customisation to the plugin I wrote, could we manage the cursor type directly in the CSS / StyledComponent for now?

atomworks (changes requested):
This is great. I have one small request relating to something that will get lost if I update the theme but otherwise this is perfect. Thank you!

atomworks (approved):
Looks good, thanks!

PR #72, opened 2021-05-13
Add custom icon support

JoshLipps:
I was using Dynamic imports but had many wierd edge cases. Instead I just make a separate Icons package. we can publish to npm the free set of icons. And any custom icons can be published privately using that Repo. see https://github.com/future-standard/icons/blob/main/Usage.MD

To test using the Private github branch checkout this repo: 
https://github.com/JoshLipps/scorer-ui-kit-example


isacoder (approved):
✅  Live reviewed functionality 

Josh will add the npm dedupe to the icons repo documententation and then all is G2G

PR #73, opened 2021-05-13
Modal and Notifications on storybook

isacoder:
### Requirement
[Bugherd#34](https://www.bugherd.com/projects/216308/tasks/34)
[Storybook] Make working story for modals
[Bugherd#33](https://www.bugherd.com/projects/216308/tasks/33)
[Storybook] Make working story for global notifications.

### Implementation
Moved all the examples from ModalPage and NotificationsPage into Storybook, didn't move the table one.

Code is practically the same, the only interesting part of the PR is the knobs

CardModal and LoginModal code is currently in storybook since they were made as non polished examples.
If you prefer to have them as part of the UI Library let me know. ( This might require extra specs)

### Screenshoots

**Modals Folder**
Screen Shot 2021-05-13 at 20 33 13

Notification Story
Screen Shot 2021-05-13 at 20 36 37

Mix Example (Modal and notifications)
Screen Shot 2021-05-13 at 20 37 21


src/context/NotificationContext.tsx
```diff
@@ -2,77 +2,76 @@ import React, { useState, useCallback } from 'react';
 import Notification, { INotificationProps } from '../Alerts/atom/Notification';
```
isacoder, line 2: I did an auto-format from the vscode tool, there is actually no "real change in this file"

src/context/index.tsx
```diff
@@ -1,11 +1,5 @@
-import {
-  NotificationProvider,
-} from './NotificationContext';
-
-import {
-  ModalContext,
-  ModalProvider,
-} from './ModalContext';
+import NotificationProvider from './NotificationContext';
+import ModalProvider, { ModalContext } from './ModalContext';
```
isacoder, line 3: Storybook was complaining that there was no "default export" in the contexts

src/context/NotificationContext.tsx
```diff
@@ -2,77 +2,76 @@ import React, { useState, useCallback } from 'react';
...
 };
 
 const defaultContext: NotificationContextType = {
-  sendNotification : () => console.log("This is the context initialization should not appear"),
-}
+  sendNotification: () => console.log("This is the context initialization should not appear"),
+};
 
-const NotificationContext = React.createContext<NotificationContextType>(defaultContext);
+export const NotificationContext = React.createContext<NotificationContextType>(defaultContext);
```
atomworks, line 12: I tend to feel it's better practice to group the exports rather than having one in the middle of code. So perhaps in this instance it's good to declare the `const` here andexport this and any others at the end of the file.

atomworks (changes requested):
That's great getting this working! It  looks so good having this in the Storybook.

There's one small comment regarding exports. I assume it will be best to merge the outstanding PRs in order so that request shouldn't cause any big delayed to getting this merged. If we need to push on and get this merged I think we could still do that though.

src/context/NotificationContext.tsx
isacoder: I don't have a preference for this so I'll go ahead an update

atomworks (approved):
Thank you!

PR #77, opened 2021-06-14
Add example with mixed values for duration slider

isacoder:
### Requirements

Update the duration slider UI kit component to allow a mix of units

### Implementation

Previous implementation allowed to have custom marks and the slider does not take account of the unit but to update the top right label. I have updated to allow the user to control this tag and value. The rest of the component stays the same.

### Screenshots

Screen Shot 2021-06-14 at 21 41 07


atomworks (approved):
Looks good to me. Thanks for the quick response on this.

PR #82, opened 2021-06-15
Feature/responsive login

isacoder:
### Requirements

[Bugherd#70](https://www.bugherd.com/projects/216308/tasks/70)
Update login to be responsive based on scorer login.

### Implementation

There are 3 breakpoints all managed with media queries

- Mobile
- Ipad portrait
- Desktop

The Logo might not look centered at glance but it's only because the placeholder logo has empty space to the right.

Login.tsx is not scorer ui kit component but an example page.
I saw in the traffic counter sass project that this page is copied as it is, because of that I didn't update the localizations. Also noticed in the scorer_ui project that there are a lot of other functionalities that are not in this base component mostly related to validation, which makes me think that is one of the reason this is not a "provided" component.

### Screenshoots

This functionality can be reviewed at the examples page

**Mobile**
Screen Shot 2021-06-15 at 17 54 57

**Ipad portrait**
Screen Shot 2021-06-15 at 17 54 17

**Desktop**
Screen Shot 2021-06-15 at 17 54 02


atomworks (changes requested):
A small request to how keyframes are implemented.

I would like however to spend some more time on this. The existing template had some inconsistencies with the design which are still present here so it would be good to get this all tightened up so this becomes the favourable source of the login UI moving forward.

example/src/pages/Login.tsx
```diff
@@ -28,22 +26,34 @@ const Box = styled.div<{ margin?: string; flex?: string;}>`
...
-  padding: 70px 60px;
-  flex: 1;
-  @keyframes fadeIn{
-    0% {
-      opacity:0;
+    max-width: 320px;
+    margin: auto;
+    padding: 44px 0;
+    min-height: calc(100vh - 126px);
+    @keyframes fadeIn{
```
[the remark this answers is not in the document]
isacoder: I have updated the keyframes using styled components and also the theming, although the values are not a match. Please review it working as expected.

isacoder:
Hello after reviewing the design provided https://zpl.io/25QA5m3

I have made these changes that I have previously reviewed with @atomworks 

**Design** 

- Left image content should be half and form the other half
- Add Pinkish gradient option too
 - Background Image in the left should be the only the middle to down portion of it. 
 - Logo should be bigger   Content - Paragraph of explanation before form fields 
 - Forgot password link
- Copyright section with links.

  Mobile: 

   - Background Image cross should be centered and not in a side

example/src/pages/Login.tsx
```diff
@@ -1,20 +1,32 @@
...
+import { ButtonWithLoading, TextField, PasswordField, Form, AlertBar, AlertWrapper } from 'scorer-ui-kit';
 import GhostLogo from '../svg/ghost-logo.svg';
 import {LoginScreen} from '../svg';
+import {Link} from 'react-router-dom';
 
 const widthDesk = 480;
 
+const gradients = {
+  "primary" : `linear-gradient(-45deg, #5CA0D1, #7DB8DB)`,
+  "secondary" : `linear-gradient(139deg, hsl(250, 60%, 62%), hsl(0, 46%, 54%))`,
```
JoshLipps, line 12: Minor: can we use consistent units here? also why are these template strings?

JoshLipps (approved):
LGTM.

example/src/pages/Login.tsx
isacoder: Hi!, yes sorry I left the values as they were, will update with the hsl ones.

isacoder: Also to answer the template strings was added not to prevent any error, just thought it would be natural to use it like that since they are used in the styled component values, but I can update to normal strings.

atomworks: It's better to stick with using hsl / hsla for colours as they're much more readable and easy to work with than hex of rgb.

atomworks (approved):

PR #84, opened 2021-06-24
Feature/slider improvements

isacoder:
### Requirements

[Bugherd75](https://www.bugherd.com/projects/216308/tasks/75)
[Duration Slider] Update ghost behaviours (see comments)

[Bugherd76](https://www.bugherd.com/projects/216308/tasks/76)
[Duration Slider] Have the duration value on the top right update during the drag action.

[Bugherd77](https://www.bugherd.com/projects/216308/tasks/77)
[Duration Slider] For top right value indicator, unit should be styled differently to value.
https://app.zeplin.io/project/609a0a289f658b3d761f5ecd/screen/60b8a885921d0ea43b8c549a

### Implementation

New prop onlyMarkSelect to make selection only to have the marks values.

The implementation of this features affected the behaviour with percentage slider as well.
As reviewed with @atomworks percentage slider will not show ghostThumb unless is passed the prop onlyMarkSelect.

Performance improvements:
I removed the thumb states in an attempt to make the this feature performant.
The ghost is a simple variable and the selected is a ref variable to carry the value thru renders.
Also verifying the state only updates when the value changes and callbacks are only called when value changes too.
Although the jumps show less in local, there should be no jumps in production.

Touch screen:
I added the touch screen capabilities but I will need to try them in real device once is deployed in demo.

Localization:
Now that unit and value can be modify from outside I have updated also the possibility to send values and units directly from outside for localization.

Josh helped me to add the serve option to run the build like production in local for storybook.
This is also added in this PR.

### Screenshoots

Screen Shot 2021-06-24 at 10 28 25


JoshLipps (approved):

JoshLipps (approved):
LGTM

src/Form/atoms/SliderInput.tsx
```diff
@@ -100,13 +103,22 @@ const thumbLeftPosition = (value: number, min: number, max: number ) => {
...
+};
+
 /**
  * Rules for a getValidMin
  * Max needs to be bigger than Min
  * if Min is not available and Max is positive default is 0
  * if Min is not available Max is negative min will be reduce by 1
  */
-const getValidMin = (max: number, min?: number) : number => {
+const getValidMin = (max: number, min?: number) : number => {
```
atomworks, line 121: Could you check to make sure your VS Code is set to strip trailing whitespaces please. :)

atomworks (approved):
Just one comment about trailing whitespace but nothing that needs to hold up merging.

Issue #103, opened 2021-07-05
[Table] Add header grouping


atomworks:
There's designs available on Zeplin for this. In regards to execution it would be great to see...

- Proposal for how this can work as part of (or alongside) the `ITableColumnConfig` setup we have now.
- An implementation that has scope to be adapted later on in a potential responsive table update.

Quick design screenshot for reference:
![image](https://user-images.githubusercontent.com/808289/124446629-90a09d00-ddbb-11eb-8d0e-7f0e197dc71c.png)

PR #131, opened 2021-07-06
Add Header grouping TypeTable

isacoder:
### Requirements
 Issue #103 header grouping 

### Implementation

My proposal for this update is two main changes a prop in the main table 
I'm thinking that maybe there will be some other tables that might want to have smaller text headers without providing groups.

```js
hasHeaderGroups: boolean
```
I decided to put the groupTitle in the  ITableColumnConfig for posible flexibility in the feature, in case someday we might want to re-organize columns. 

```js
export interface ITableColumnConfig {
  columnId?: string
  groupTitle?: string <---
  header: string
  sortable?: boolean
  sortActive?: boolean
  cellStyle: TypeCellStyle
  alignment?: TypeCellAlignment
  showUnit?: boolean
  showStatus?: boolean
  hasCopyButton?: boolean
  width?: number // TODO: This!
  minWidth?: number
}
```

I took the opportunity to abstract the Header from the TypeTable into a new component since this feature was growing big.

I saw in the designs that there is a line bellow the heading titles for "non-groups" header.
I wasn't sure if this is also required on this Issue or if there will be a follow up. Please let me know.

Screen Shot 2021-07-07 at 19 01 24

### Screenshots

I added this feature to the TypeTable Story since this one can show more broadly the interactions.

Screen Shot 2021-07-07 at 18 29 56


atomworks (changes requested):
I've only done a brief look over the code, first glance is good but I'll look in more detail next review. Before that though there's a few things in regards to layout that need to be updated as it's a little off design.

Could you check and adjust the following please:

- [x] Check the alignment of the group heading and column heading (where left aligned). In the Storybook example, both Ds in Devices should be perfectly aligned.
- [ ] The font weight (and maybe color) are incorrect.
- [ ] There isn't enough spacing vertically between the group header and column headers.
- [ ] Storybook related: Can we have the group header cover the Created column too and have usage cover the last two columns?

Regarding the first and last point, later on I think we'll have to review how this grouping works if the alignment is central or to the right. For now this tidies up the example a little!

Just let me know if you have any questions, thank you!

isacoder:
> I've only done a brief look over the code, first glance is good but I'll look in more detail next review. Before that though there's a few things in regards to layout that need to be updated as it's a little off design.
> 
> Could you check and adjust the following please:
> 
> * [x]  Check the alignment of the group heading and column heading (where left aligned). In the Storybook example, both Ds in Devices should be perfectly aligned.
> * [x]  The font weight (and maybe color) are incorrect.
> * [ ]  There isn't enough spacing vertically between the group header and column headers.
> * [x]  Storybook related: Can we have the group header cover the Created column too and have usage cover the last two columns?
> 
> Regarding the first and last point, later on I think we'll have to review how this grouping works if the alignment is central or to the right. For now this tidies up the example a little!
> 
> Just let me know if you have any questions, thank you!

Hello!

It's so good to have your designer eyes on this 🤦‍♀️  can't believe I didn't see the color error in the header grouping.
I have updated to the Zepplin values but it still looks a bit off, let me know if it should be different.

The 2px margin left was totally outside of my scope but thankfully you also notice that one.

About the Story I might have not understood correctly the required update.

Here is an example with only one grouping.

Screen Shot 2021-07-08 at 18 08 12

Here is the example with both groupings (looks a bit odd due the right alignment)

Screen Shot 2021-07-08 at 18 07 50


About the distance between Group heading and Column heading how far should it be.. this is currently how it looks compare to the design.

Screen Shot 2021-07-08 at 18 09 54

atomworks (approved):
👍

PR #133, opened 2021-07-07
Feature: Add theme fallback helper

JoshLipps:
enable use of partial dark theme even when not all defaults have been set.

### DIFF
--- a/src/themes/dark.tsx
+++ b/src/themes/dark.tsx
@@ -4,6 +4,35 @@ const colors = {
 
 };
 
+const styles = {
+  "form": {
+    "button": {
+      "primary": {
+          "default": {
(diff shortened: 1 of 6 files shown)


src/themes/dark.tsx
```diff
@@ -4,6 +4,35 @@ const colors = {
 
 };
 
+const styles = {
+  "form": {
+    "button": {
+      "primary": {
+          "default": {
+              // "backgroundImage": "linear-gradient(135.00deg, blue 0%, green 100%)"
```
JoshLipps, line 12: I reallt think this is what the dark primary should be... :grinning:

JoshLipps: image

atomworks (approved):

isacoder (approved):
✅ Thanks for adding the support of partial fallback theme. All looks good :D

PR #134, opened 2021-07-08
Fix: Speed up the CI by not building 2 times during release

JoshLipps:
https://github.com/future-standard/scorer-ui-kit/blob/0373025aa5d7b88fccda550395f52d14722779a6/package.json#L20 
pre publish only already makes it build on publish.

### DIFF
--- a/.github/workflows/publish-beta.yml
+++ b/.github/workflows/publish-beta.yml
@@ -13,7 +13,6 @@ jobs:
         with:
           node-version: 12
       - run: npm install
-      - run: npm run build
       - uses: JS-DevTools/npm-publish@v1
         with:
           token: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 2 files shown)


isacoder (approved):
I ran this version with ACT and the result was successful but i notice that prepublishOnly didn't run :/ not sure if this is because it's on local. 

I think It might be better to merge it and use it  just to make sure is a local thing.

 if not, probably we should remove the prepublish command on the package.json and leave the npm run step in the workflow 🤔

JoshLipps:
> I ran this version with ACT and the result was successful but i notice that prepublishOnly didn't run :/ not sure if this is because it's on local.
> 
> I think It might be better to merge it and use it just to make sure is a local thing.
> 
> if not, probably we should remove the prepublish command on the package.json and leave the npm run step in the workflow 
> 
> Screen Shot 2021-07-08 at 20 32 06

you can check the actual, non dry runs and it does run. See the existing releases.

Issue #105, opened 2021-07-05
[Top Bar - Custom Drawer] Add a custom drawer feature to the top bar


atomworks:
Currently where we have the Notifications and User Menu in the top bar that open up their contents in draws on the side of the browser (and full screen in responsive mode), I want to add an optional customisable option to allow any project to add:

- An icon to the header.
- That icon can have a status color dot / counter.
- The drawer contents will be blank and allow for a custom component to be injected.

PR #135, opened 2021-07-09
CustomDrawer in TopBar

isacoder:
### Requirements
The requests of Issue #105 

Custom drawer with:

- An icon to the header.
- That icon can have a status color dot / counter.
- The drawer contents will be blank and allow for a custom component to be injected.

### Implementation

I created a StatusIcon to cover Dot and counter cases.

I notice in the design that the icon with status is smaller than the others, however the icon with the numbers looked kind of weird smaller than the counter so I left it same size, but let me know if it should be different.

### Screenshoot

**Status Icon**
Screen Shot 2021-07-13 at 10 45 20

**Desktop Version**
Screen Shot 2021-07-13 at 10 44 16

**Mobile Version**
Screen Shot 2021-07-13 at 10 44 41

### DIFF
--- a/src/Global/molecules/TopBar.tsx
+++ b/src/Global/molecules/TopBar.tsx
@@ -132,7 +132,7 @@ const NotificationsContainer = styled.div`
     margin-right: -17px;
 `;
 
-type IDrawerKeys = 'user' | 'notifications' | 'custom' | undefined;
+type IDrawerKeys = 'user' | 'notifications' | 'custom' | null;
 
 
 const TopBar: React.FC<ITopBar> = ({
@@ -152,14 +152,14 @@ const TopBar: React.FC<ITopBar> = ({
   onLanguageToggle = () => { }
 }) => {
 
-  const [openDrawer, setOpenDrawer] = useState<IDrawerKeys>(undefined);
+  const [openDrawer, setOpenDrawer] = useState<IDrawerKeys>(null);
 
   const toggleDrawers = (drawerKey: IDrawerKeys) => {
     setOpenDrawer(
       prevDrawer => {
-        // if prevDrawer is open, just update to undefined to close
+        // if prevDrawer is open, just update to null to close
         if (prevDrawer === drawerKey) {
-          return undefined;
+          return null;
         }
 
         return drawerKey;
--- a/src/Global/molecules/TopBar.tsx
+++ b/src/Global/molecules/TopBar.tsx
@@ -2,10 +2,10 @@ import React, { useState } from 'react';
 import styled, { css } from 'styled-components';
 
 import Icon from '../../Icons/Icon';
+import StatusIcon from '../../Icons/StatusIcon';
 import UserMenu from '../molecules/UserMenu';
 import { ITopBar } from '../index';
 import NotificationsHistory from './NotificationsHistory';
-import { IStatusDot } from '../..';
 
 const Container = styled.div`
   z-index: 9;
@@ -132,37 +132,6 @@ const NotificationsContainer = styled.div`
     margin-right: -17px;
 `;
 
-const IconStatusWrapper = styled.div`
-  position: relative;
-`;
-
-const StatusCounter = styled.div<{color?: IStatusDot}>`
-  background-color: ${({theme, color}) => color ? theme.colors.status[color] : 'hsla(0, 0%, 91.8%, 1.000)'};
-  position: absolute;
-  right: 7px;
-  top: -12px;
-  border-radius: 50%;
-  height: 14px;
-  width: 14px;
-  font-size: 10px;
-  color: white;
-  display: flex;
-  align-items: center;
-  justify-content: center;
-`;
-
-const StatusDot = styled.div<{color?: IStatusDot}>`
-  background-color: ${({theme, color}) => color ? theme.colors.status[color] : 'hsla(0, 0%, 91.8%, 1.000)'};
-  width: 10px;
-  height: 10px;
-  border: solid 2px hsl(0, 0%, 100%);
-  border-radius: 50%;
-  position: absolute;
-  top: -6px;
-  right: 7px;
-`;
-
-
 type IDrawerKeys = 'user' | 'notifications' | 'custom' | '';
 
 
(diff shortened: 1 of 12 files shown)


src/index.tsx
```diff
@@ -271,6 +275,8 @@ export type ITimeUnit = 'seconds' | 'minutes' | 'hours';
 
 export type IMediaType = 'img' | 'video'
 
+export type IStatusDot = 'caution' | 'danger' | 'good' | 'neutral';
```
atomworks, line 278: We're going to need the blue color in here too which is `folder`. The problem (as I'm sure you're already thinking) is that we're using the status references from SMC and so that makes no sense here. 

I think perhaps we can add `highlight` as an option in the theme and here. Ideally we replace `folder` but only if we're certain that we're not using it in the code at all. Would you be able to add it to the theme manually? I've already added the reference to my theme generation setup so it shouldn't get lost later.

src/Global/molecules/TopBar.tsx
```diff
@@ -130,6 +132,9 @@ const NotificationsContainer = styled.div`
     margin-right: -17px;
 `;
 
+type IDrawerKeys = 'user' | 'notifications' | 'custom' | '';
```
atomworks, line 135: What is the empty string being used for here? Seems like it's better to allow for a null or undefined value instead?

storybook/src/stories/Global/TopBar.stories.tsx
```diff
@@ -42,8 +115,27 @@ export const _TopBar = () => {
...
+        hasSearch,
+        hasLogout,
+        hasNotifications,
+        logoutLink,
+        searchPlaceholder,
+        hasLanguage,
+        hasCurrentUser,
+        notificationsHistory
+        }
+        }
```
atomworks, line 136: Is this a linting thing? It's a bit confusing to have the opening `{` on the same lines but the closing `}` on separate lines.

src/Icons/StatusIcon.tsx
```diff
@@ -0,0 +1,54 @@
...
+import Icon from './Icon';
+import { IStatusDot } from '..';
+
+const Container = styled.div`
+  position: relative;
+  display: inline-block;
+`;
+
+const StatusCounter = styled.div<{ color?: IStatusDot }>`
+  background-color: ${({ theme, color }) => color ? theme.colors.status[color] : 'hsla(0, 0%, 91.8%, 1.000)'};
```
atomworks, line 12: I think if there isn't a status color set for this we should make it transparent by default. It would be good to keep the fallback value but use that to set transparency. `'hsla(0, 0%, 91.8%, 0)'`

Then if we add an eased transition of around 750-1000ms perhaps, when this status icon changes the animation should help make that state change more noticeable but without being over the top.

If you feel it's better to do this with `opacity` for any reason then that works too. Maybe that helps with the numbers. Also if you think it's all a terrible idea, please feel free to let me know :)

src/Icons/StatusIcon.tsx
```diff
@@ -0,0 +1,54 @@
...
+  padding: 2px;
+  font-size: 10px;
+  color: white;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+`;
+
+const StatusDot = styled.div<{ color?: IStatusDot }>`
+  background-color: ${({ theme, color }) => color ? theme.colors.status[color] : 'hsla(0, 0%, 91.8%, 1.000)'};
```
atomworks, line 28: Same feedback as the status counter!

src/Icons/StatusIcon.tsx
```diff
@@ -0,0 +1,54 @@
...
+  icon: string
+  status?: IStatusDot
+  counter?: number
+}
+
+const StatusIcon: React.FC<IStatusIcon> = ({icon, status, counter}) => {
+  return (
+    <Container>
+      {status && (counter === undefined) && <StatusDot color={status} />}
+      {(counter !== undefined) && <StatusCounter color={status}>{counter}</StatusCounter>}
```
atomworks, line 48: As it would be good to have this present at all times (and simply transparent) I think only need to conditionally show one or the other which might make this logic a bit easier to tidy up and keep readable. It's not quick at a glance if both of the things could be present.

atomworks (changes requested):
Got a few changes to request but really please with how this is looking, especially the mobile part. Really happy with how you handled the counter too. Looks great.

Let me know if there's any questions or if something doesn't make sense.

storybook/src/stories/Global/TopBar.stories.tsx
```diff
@@ -16,9 +16,82 @@ export default {
...
+];
 
 
+// unread read can be empty array when there are no notifications
+const allNotifications: INotificationsHistory = {
+  unread: unreadNotifications,
+  read: readNotifications,
+  noNotificationsText: 'NO NEW NOTIFICATIONS',
+  readNotificationsText: 'NEW',
+  unreadNotificationsText: 'READ',
```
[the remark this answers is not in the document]
isacoder: Sorry, the capitalization was enforced but I forgot to put lowercase for the example in storybook last time.

src/index.tsx
isacoder: Added the highlight as requested. 
How can you even think or removing folder?.
That color it's totally in fashion now days  🤭

src/Global/molecules/TopBar.tsx
isacoder: makes a lot of sense, I'll update.

storybook/src/stories/Global/TopBar.stories.tsx
isacoder: I guess it's at the same level because if it was on one line it would look like this.
I have updated to be in the same line that te last variable to avoid confusion.

{...{ lots,of, props}}

src/Icons/StatusIcon.tsx
isacoder: The logic on this one iss hard 🙈 
I have updated with a ternary operator, let me know if it's better understood,  if not I will make a function to render this one outside.

```js
     {status && (counter === undefined)
        ? <StatusDot color={status} />
        :  (counter === undefined) ? null : <StatusCounter color={status}>{counter}</StatusCounter>}
```

src/Icons/StatusIcon.tsx
isacoder: Hi!, I added the fallback and the transition.
Let me know how it looks.

atomworks (changes requested):
One last quick request!

atomworks (approved):

Issue #113, opened 2021-07-06
[Select Input] Add a select component! (Normal & Compact sizes)


atomworks:
This is so overdue that I keep forgetting it's not in there already because why wouldn't it be? 😱 

There are designs for this but I'll get those confirmed to make sure it goes as simply as possible.

atomworks:
I've created the formal designs for these. They're available on Zeplin here:
https://zpl.io/blO76Nz

PR #138, opened 2021-07-14
Feature/select

isacoder:
### Requirements

This PR resolves issue #113 that requests a select with two sizes with a label

### Implementation

Sorry for the delay, I misunderstood the requirements since the word "input" was in the description I though this was going to be a full custom component. Then after a conversation with @atomworks made me realize this required a "normal select" with style and not having style for the dropdown options 😓 
 
I hope I have implemented as expected but let me know if this needs to be different.

The hardest part was the placeholder as this is not "native" of the select, I created an invalid option that is shown by default but can't be selected.

I applied the same logic as the Tag with link for this SelectField, when no Label props are sent the label will not be added.

Updated the folders of Storybook because Form was growing big.

### Screenshoots

**Chrome**

Screen Shot 2021-07-14 at 18 42 19

Screen Shot 2021-07-14 at 18 42 27

**Firefox**

Screen Shot 2021-07-14 at 18 41 21

Screen Shot 2021-07-14 at 18 41 29


src/Form/atoms/SelectField.tsx
```diff
@@ -0,0 +1,140 @@
...
+interface ILabel {
+  htmlFor: string
+  labelText: string
+}
+
+interface OwnProps {
+  labelProps?: ILabel
+  styleSize?: ISelectSizes
+  placeholder?: string
+  selectWidth?: string
```
atomworks, line 84: Can we just use `width` for this as it's the Select component?

src/Form/atoms/SelectField.tsx
```diff
@@ -0,0 +1,140 @@
...
+`;
+
+interface ILabel {
+  htmlFor: string
+  labelText: string
+}
+
+interface OwnProps {
+  labelProps?: ILabel
+  styleSize?: ISelectSizes
```
atomworks, line 82: Simplify `styleSize` to `size` to stay consistent with how we're doing things in the `Button` component.

src/Form/atoms/SelectField.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const [activePlaceholder, setPlaceholderInactive] = useState<boolean>(true);
+
+  const handleOnChange = useCallback((e) => {
+    setPlaceholderInactive(prev => {
+      if (prev) { return false; }
+      return prev;
+    });
+
+    const { value } = e.target;
```
atomworks, line 107: Better to keep this at the top of the function for readability.

src/Form/atoms/SelectField.tsx
```diff
@@ -0,0 +1,140 @@
...
+    position: absolute;
+    top: calc(50% - 7px);
+    right: 10px;
+  }
+`;
+
+interface ILabel {
+  htmlFor: string
+  labelText: string
+}
```
atomworks, line 78: I don't think this is consistent with the other form elements we have. This seems like a sensible choice although we need to consider consistency first. What are your thoughts here?

atomworks (changes requested):
We've chatted quickly on Slack about this but just wanted to recap the hands on test points below (and a couple of other small bugs I noticed). Then there's a few code comments too. Please let me know if anything doesn't make sense. 

- [ ] Width prop isn't working correctly, seems like main container needs the value and internal parts should scale to that.
- [ ] The files that have been moved, can we change that folder name from `InputStories` to just `Input` as we're already in a stories folder so we don't need that distinction.
- [ ] Font weight on placeholder and value seems too heavy.
- [ ] Vertical alignment of font seems off (could be the font weight making it look like this though)
- [ ] Left padding is slightly off.
- [ ] Clicking on the down arrow doesn't register click (and can start a text highlight moment)
- [ ] Make the select text un-text-selectable (related to above)

src/Form/atoms/SelectField.tsx
isacoder: I wanted to make having a Label as optional in this component but if label is used I wanted to make htmlFor a required to enforce the developer to provide that value.

We could drop the Label functionality if you want and just have the Select component.

src/Form/atoms/SelectField.tsx
isacoder: I was trying to make it flexible like the input, I will just leave it with 100% and let it be adjust to the container.

src/Form/atoms/SelectField.tsx
isacoder: sorry, I had a problem using size directly in the select because it's a property.
I have moved the value to the container and handled there.

isacoder: Sorry updating again.. so size in the select is the amount of rows allowed to be displayed.
I will keep the interface and the name to avoid confusion.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select

atomworks: Ahhhh, I didn't realise that size was a prop on select already. It's not something I've used much so it didn't occur to me.

Thinking ahead a little, the sizes we use for form fields should always ever be one of two sizes. The standard size (most commonly used) and the compact size. It's not like buttons where we have four sizes. So perhaps an option we can later integrate in to how we handle other fields would be to simply have this as a `boolean` value such as `isCompact`. What do you think?

src/Form/atoms/SelectField.tsx
atomworks: I think when we provide a `<select>` wrapped with the label then it's good to leave the select at 100%. If we're importing it directly I think there is use cases where it makes sense to allow an option width as you've described. 

By default it should just scale automatically but I can see times were it may be used in some flex based columns were we'd want to set the sizing. Perhaps in that case a max-width option would be good too. 

We can iterate later so feel free to make a decision you feel is best and we can adjust in the future as we use it and see how it's being used.

src/Form/atoms/SelectField.tsx
atomworks: Perhaps something like this might be a good middle ground. That way the label is option but if you do provide it then both of those props become mandatory.

```
interface ILabel {
  htmlFor: string
  text: string
}

interface IProps {
    label?: ILabel
}
```

atomworks (commented):
I've answered the open questions so I'll submit this as a review to set the  status accordingly. Re-request a review any time you need anything else from me on this.

isacoder:
Hi I have updated the story to have a fixwidth option by getting providing the selectWrapper to be used, also added an extra example that is a page list inspired by some of the cases I've sen in the projects.

In the previous update I notice the font-width was 500 but updated to 400 since for lato there is not 500.. but I wasn't sure if this fixed the weight issue, please let me know.

Screen Shot 2021-07-30 at 16 39 06

src/Form/atoms/SelectField.tsx
isacoder: I have updated as suggested and added sameRow prop optional to make the label in the same row that the select.

src/Form/atoms/SelectField.tsx
```diff
@@ -114,31 +127,31 @@ const SelectField: React.FC<ISelect> = ({
       if (prev) { return false; }
       return prev;
     });
-
-    onChange(value);
-  }, [onChange]);
+    console.log(value);
```
atomworks, line 130: Best to avoid `console.log` in production code.

atomworks (approved):
Looks good. I've added one comment about a stray `console.log` so if you get a chance to remove that before merging, that would be great. No enough to delay any release though.

PR #153, opened 2021-08-03
Avatar Upload feature iteration

### LINKED ISSUE #141: [Scorer UI Kit] - Avatar Uploader improvements
As crop tool is being used inside AvatarUploader component we are not able to change options
(title, cancelBtnTxt, cropBtnTxt ) of crop tool

- prop for receiving a base img

isacoder:
### Requirements
This is an update for  the Croptool
- Allow to pass a image
- Update hasCroptool
- Add props to pass button text to Croptools

closes #141

### Implementation

**Default img**
The Avatar uploader didn't have the option before.

**hasCrop update**
After the hasCrop was updated the logic with both was implemented this way, please confirm this is the expected.
basically if an image is provided and has crop will only edit that image in any other case the feature will allow the user to update the image and crop it if the crop feature is enabled.

Please confirm the logic of both together.

| initial button text  | defaultImg   | hasCrop   | button text after actions  |
|------------------|-------------|------------|------------------|
| Crop Image.        |     ✅              |  ✅               |  Crop Image         | 
|  Select file           |                       |  ✅              |  Replace photo    |
|  Replace photo   |      ✅              |                   |  Replace photo    |
|  Select file            |                      |                    |  Replace photo    |

Currently this feature can be reviewed in the [demo link](https://future-standard.github.io/scorer-ui-kit/storybook/?path=/story/form-file-management--avatar-uploader)

**Screenshots**

Screen Shot 2021-08-04 at 11 30 44

Screen Shot 2021-08-04 at 11 30 51

Screen Shot 2021-08-04 at 11 30 56

### DIFF
--- a/src/Form/atoms/ButtonWithIcon.tsx
+++ b/src/Form/atoms/ButtonWithIcon.tsx
@@ -31,6 +31,7 @@ const IconContainer = styled.div<{ position?: string }>`
   justify-content: center;
   align-items: center;
   overflow: hidden;
+  flex-shrink: 0;
 
   ${({ position }) => css`
     order: ${ position && position === 'left' ? 0 : 2 };
--- a/src/Form/molecules/CropTool.tsx
+++ b/src/Form/molecules/CropTool.tsx
@@ -178,7 +178,7 @@ const CropTool: React.FC<ICrop> = ({
       );
     // triggers a render to repaint canvas
     setLoadDimensions((prevState) =>
-      ({ 
+      ({
         ...prevState,
         cropLeft: newCrop.left,
         cropTop: newCrop.top,
@@ -262,7 +262,7 @@ const CropTool: React.FC<ICrop> = ({
     if (!viewDimensions.isResizing) { return; }
     viewDimensions.isResizing = false;
   }, []);
-  
+
   const updateSelect = useCallback((left, top, width, height) =>{
     if (!cropRef.current) { return; }
     cropRef.current.style.left = `${left}px`;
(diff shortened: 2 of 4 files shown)


atomworks (changes requested):
There's one implementation thing and one small naming thing in the review comments itself.

The bigger thing I think it the `hasCrop` behaviour. To reference a design (see below), once the uploader has an image there are two actions associated with it. One is to (re)crop it, the other is to remove it (and reset it to an empty state). It seems this value swaps out the buttons. We're also using a different button style here as once an image is uploaded the buttons become more secondary actions.

I think it's useful to have some preset buttons though I'm unsure if how current implementations handle this, so we may need to reach out to check.

image

isacoder:
Hi @atomworks, I have updated the buttons and the behaviour of the tool please review that is working as expected.

case | Initial button text                                  | currentImg   | hasCrop   |
------|-------------------------------------|-------------|------------|
 1 |  Replace photo  and Remove button  |      ✅              |                   |  
2 | Crop Image and Remove button         |     ✅              |  ✅               |  
3 |  Select file                                            |                       |  ✅              |  
4 |  Select file                                            |                      |                    |  


**Case 1**
Note that the replace button is a FileInputButton. This component currently doesn't have an icon feature.
I could add it but wanted to first review the behaviour. If this update is wanted please advice in the expected icon for this feature.

Screen Shot 2021-08-12 at 16 59 58


**Case 2**
Screen Shot 2021-08-12 at 17 01 12


**Case 3**
Screen Shot 2021-08-12 at 17 01 29


**Case 4**
Screen Shot 2021-08-12 at 17 01 41

atomworks (approved):
Thanks. 

For future reference, the icon buttons in the screenshot look like they're not laying out as they should but that's not a bug to deal with in this PR, I'll make an issue for it. 

I think we need to have more consistency with icon buttons and normal buttons too here but as the designs don't pair replace and remove at the same time this isn't an issue for now, just something to keep in mind.

PR #156, opened 2021-08-05
Fix closeText in Modal over screen and improve MediaModal hook

### LINKED ISSUE #142: [MediaBox/ Modal] - Close Text overflows the screen
### Issue

The video currently there is no limit to he height in modal probably we should limit it

### Screenshoot example

Screen Shot 2021-07-28 at 18 45 55

### LINKED ISSUE #155: [MediaBox/ Modal] - useMediaModal hook
### Description

Currently the modal and the mediabox can be use together to create a media modal.
This is flexible but could be easier to use with a hook. 

Also media url are invalid and sometimes fail and needs to be checked before sending the urls.
this could be improved in the url were check at the item rendered level.

isacoder:
### Requirements
closes Bug #142 The video currently there is no limit to he height in modal probably we should limit it
closes Feature #155 - useMediaModal hook

### Implementation

- Updated MediaBox component to limit height and width space for modal but it's retro-compatible because this is only added if prop hasModalLimits is true
- Updated modal to have isModalOpen value available since it was mention that it would be helpful
- created useMediaModal hook to manage modal and MediaBox together with some strict values like width: auto, no padding and should always be close able.
- onError, onLoad props features for MediaBox
- Zoom in TableThumbnail only available if there is preview

currently this feature can be reviewed in the demo link.
https://future-standard.github.io/scorer-ui-kit/#/table

### Screenshoot

Screen Shot 2021-08-06 at 11 41 04

Video of interaction
Basically TableThumbnail component verifies preview link works if it doesn't zoom and click is not available. 
Video and image are now responsive width and hight.

https://drive.google.com/file/d/1Qe31NbonhWJB4oqB4rw4P7wlvbbxUwVN/view?usp=sharing

### DIFF
--- a/example/src/pages/TablePage.tsx
+++ b/example/src/pages/TablePage.tsx
@@ -96,7 +96,7 @@ const initialRows : ITypeTableData = [
     id: 'device-id-4',
     header: {
       image: "http://placekitten.com/2934/3102",
-      mediaUrl:"http://placekiten.com/2934/3102",
+      mediaUrl:"http://wrong-url-placekitten.com/2934/3102",
       mediaType: 'img'
     },
     columns:
--- a/example/src/pages/TablePage.tsx
+++ b/example/src/pages/TablePage.tsx
@@ -80,8 +80,8 @@ const initialRows : ITypeTableData = [
   {
     id: 'device-id-3',
     header: {
-      image: "http://placekitten.com/1300/1300",
-      mediaUrl:"http://placekitten.com/1300/1300",
+      image: "http://placekitten.com/1934/3102",
+      mediaUrl:"http://placekitten.com/1934/3102",
       mediaType: 'img'
     },
     columns:
--- a/example/src/pages/TablePage.tsx
+++ b/example/src/pages/TablePage.tsx
@@ -94,10 +94,25 @@ const initialRows : ITypeTableData = [
   },
   {
     id: 'device-id-4',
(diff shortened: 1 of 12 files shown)


atomworks (changes requested):
There's a couple of issues I've noticed here so wanted to address those before going through the code.

- The disabled hover behaviour has some impact on usability. It's definitely good to disable it when there is no content at all (such as the final item in the example) but when there is thumbnail but no large version it is better to keep the zoom on hover which still has value. For implying a modal is available it's better to use `cursor` properties to do this.

- Seems like aspect ratio gets blown out and stretch on images when I checked it on linked Examples. I wanted to check the video here too but there didn't seem to be anything here.

image

isacoder:
Welp.. I couldn't replicate but since we review it together I have updated in an attempt to try something in your setup..
basically max-width and max-height 100% base.

https://future-standard.github.io/scorer-ui-kit/#/table

Since we talked about it I have updated the modal in storybook and the Tables to have modals too.

https://future-standard.github.io/scorer-ui-kit/storybook/?path=/story/alerts-modals--media-modal

Previous version can be review in this Story by removing the prop ' Has Modal Prop'
You will need to retrigger the modal it since story book does not refresh this.

Screen Shot 2021-08-12 at 19 03 53

atomworks (approved):
👍

PR #157, opened 2021-08-10
Feature/ Search-Input for FilterBar

### LINKED ISSUE #147: [Filter Bar] - Search Input
### Description 

This is part of a bigger feature that is Filter bar, the description of the full feature is on design issue #109

-- Input  with Icon

design: https://zpl.io/brGpWYO

isacoder:
### Requirements

closes #147 - FilterBar SearchInput

### Implementation

I used the theming for compact input since it look more closer to the zeppelin values. Also there was not an example for non placeholder text, so took the value from the theme. 

However I saw that there were values in typography for SearchInput for the Topbar in the theme, but is very different.
Not sure if this one should have a new theme value. 

I wanted this search input to be able to use in the filter bar but also in the Dropdown component, so I have added the customization.

Screen Shot 2021-08-10 at 18 54 51

Added in dimensions Strong as weight for Icons because heavy looked thiner than the design however it looks a little bit chopped :(. Please advice in this point, should I keep it heavy or do I need to update the icon with more space in the icons repo?

### Screenshoots

Screen Shot 2021-08-10 at 18 57 29

Screen Shot 2021-08-10 at 18 58 09

### DIFF
--- a/src/themes/common.tsx
+++ b/src/themes/common.tsx
@@ -8,8 +8,7 @@ export const dimensions = {
     weights: {
       light: 1,
       regular: 1.5,
-      heavy: 2,
-      strong: 3,
+      heavy: 2
     }
   },
   form: {
--- a/src/themes/common.tsx
+++ b/src/themes/common.tsx
@@ -8,7 +8,8 @@ export const dimensions = {
     weights: {
       light: 1,
       regular: 1.5,
-      heavy: 2
+      heavy: 2,
+      strong: 3,
     }
   },
   form: {
--- a/src/Form/atoms/SearchInput.tsx
+++ b/src/Form/atoms/SearchInput.tsx
@@ -1,22 +1,37 @@
 import React, { InputHTMLAttributes } from 'react';
-import styled, {css} from 'styled-components';
-import Icon, {IconWrapper} from '../../Icons/Icon';
+import styled, { css } from 'styled-components';
+import Icon, { IconWrapper } from '../../Icons/Icon';
 
-const Container = styled.div`
-  ${({theme}) => css`
-    border: 1px solid ${theme.styles.form.input.default.normal.borderColor};
-    height: 30px;
-    padding: 2px 2px 2px 10px;
-    align-items: center;
+const Container = styled.div<{ hasBorder: boolean }>`
+  ${({ theme, hasBorder }) => css`
+    ${hasBorder && css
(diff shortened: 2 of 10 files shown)


src/themes/common.tsx
```diff
@@ -8,7 +8,8 @@ export const dimensions = {
     weights: {
       light: 1,
       regular: 1.5,
-      heavy: 2
+      heavy: 2,
+      strong: 3,
```
atomworks, line 12: The weights I have in Sketch are 1, 1.5, 3 and 5 so we definitely need an extra level but it those heavier weights are rarely used so if we're needing them in this feature we need to look at if there's a problem in the icon workflow somewhere. 

Originally the icons should have been setup to scale in the browser in the same way that Sketch handles it. This is part of the reason we have a script to parse icons to add them to the kit. Does the weighting on this feature have the same issue if you choose different icons or is it just an issue with this icon?

atomworks (changes requested):
This looks good although I think the naming of the component needs a change. This is an input that gets used in forms but I think components in the `Form` section should be focused on forms that are used in pages for things like settings, etc. 

So this is more of a utility input used for specific UI elements and should rarely get used outside of other UI kit elements. Let's try and work out a name to better reflect this. This might be better to live as some kind of atom, maybe in global? Or something new? We can chat on Slack if you have some ideas!

src/themes/common.tsx
isacoder: About the icon weights

Screen Shot 2021-08-18 at 17 10 42

Well most of them have this problem and some do not upscale that good 🙈 

Zone
Screen Shot 2021-08-18 at 17 15 39

Retry Job
Screen Shot 2021-08-18 at 17 17 02

Temperature
Screen Shot 2021-08-18 at 17 17 24

Usage
Screen Shot 2021-08-18 at 17 18 16

Critical
Screen Shot 2021-08-18 at 17 14 32

Probably we shouldn't use this weight for now..

isacoder:
After the Slack conversation I have updated the SearchInput to BasicSearchInput and moved it to Misc.

I'll be waiting on the retro of the Icons to follow up with any requested changes.

atomworks (commented):
After our discussion I know there'll be some changes to separate out the icon issue to its own PR. Other than that though, I think we're good.

Let me know when you've done what you need to and I'll do one last check really quickly and get this approved.

atomworks (approved):

PR #158, opened 2021-08-11
Add FilterButton with story - Do not merge

### LINKED ISSUE #146: [Filter Bar] - Dropdown filter
### Description

This is part of a bigger feature that is Filter bar, the description of the full feature is on design issue #109

Dropdown filter
-- dropdown button with icon and arrow flip
-- filter search input
-- loading state
-- Quantity of options displayed
-- Options to select can be multiple or single (checkbox or radio button

### LINKED ISSUE #148: [Filter Bar] - Button with icon tag style
### Description 

This is part of a bigger feature that is Filter bar, the description of the full feature is on design issue #109

- Button with icon tag style
-- More options
-- less options
-- Add search input

design: https://zpl.io/brGpWYO

isacoder:
## Requirements

closes #148 FilterBar Button
closes #146 FilterBar Dropdown

#### Radio Buttons and Checkbox
- Implemented new Radio buttons, these work as a normal radio button just with style.
- I used  the theming values of Checkbox for radio buttons, I think the theme update should definitelly consider this either in naming or adding separated values for radio.
- I have added the parentHover option to Radio and Checkbox buttons. I wasn't so sure of doing this.
I saw 3 options 
-- Use current component and allow the parent to control the hover state (selected implementation)
-- Use current component and export Inner and Outer divs to allow the parent to control the hover 
-- To have this implemented independently inside this component with duplicated behaivour that might make hard to follow up an update in both places.

I'm willing to change it if you have a preference in either of the two options or if you have a different recommendation.

#### Selected returned value

Handling multiple types of values was the biggest challenge of this feature.
I notice in the examples that the state is initialized in "any" as the select allowed to received strings, numbers or ListMenu (title, value). I made a returned type IFilterDropdownValue to cover all scenarios (multi select too)

export type IFilterDropdownValue = string | string[] | number | number[] | ListItem | ListItem[] | null;

I looked into the examples of previous the filters, I notice that the "selected" value was using some translation tools

https://gitlab.com/futurestandard/poc-projects/tokai-led-monitor-cloud/-/blob/master/src/components/molecules/Select.tsx#L96

```js
<SelectOption selected={selected === value} onClick={() => { select(value); }} key={title}>
     {selected === value ? <CheckMarkIcon src={CheckMarkSVG} /> :<RadioBox />}<OptionText title={title.toString()}>{t(`${title}`)}</OptionText>
</SelectOption>
```

I don't think the select should handle the localization internally added examples of how to accomplish this in storybook.

I have merged the BasicSearchInput into this PR because it was not merged with main yet the code will be shown here too, sorry for the inconvenience. 

## Screenshots

I made a video showcasing the feature, to explain some of my concerns about the data.
The list of data is provided by the developer user, the selection is also controlled by the developer user.
However I see that this feature might have a complication when using Strings arrays as the provided list.

The video is 5mins long, I hope is not to much 🙈 

[Showcase video](https://loom.com/share/25bed912e8be4add9150e85b098abf16)

### FilterBar button

Screen Shot 2021-08-11 at 18 04 56

### Radio button

Screen Shot 2021-08-19 at 14 20 47

### Dropdown List
Screen Shot 2021-08-20 at 10 43 22

### DIFF
--- a/src/Form/atoms/Checkbox.tsx
+++ b/src/Form/atoms/Checkbox.tsx
@@ -51,11 +51,11 @@ const IconWrapper = styled.div<{color: ISvgIcons['color']}>`
   }
 `;
 
-const Container = styled.label<{visualState?: CheckboxState, disabled?: boolean}>`
+const Container = styled.label<{visualState?: CheckboxState, disabled?: boolean, parentHover?: boolean}>`
   user-select: none;
 
 
-  ${({visualState, disabled, theme: { styles }}) => visualState === CheckboxState.Off && css`
+  ${({visualState, disabled, parentHover, theme: { styles }}) => visualState === CheckboxState.Off && css`
     ${CheckboxOuter}{
       ${styles.form.checkbox.unchecked.default};
     }
@@ -66,6 +66,12 @@ const Container = styled.label<{visualState?: CheckboxState, disabled?: boolean}
       }`
     };
 
+    ${!disabled && (parentHover !== undefined) && parentHover && css`
+      ${CheckboxOuter}{
+        ${styles.form.checkbox.unchecked.hover};
+      }
+    `};
+
     ${disabled && css`
       ${CheckboxOuter}{
         ${styles.form.checkbox.unchecked.disabled};
@@ -74,7 +80,7 @@ const Container = styled.label<{visualState?: CheckboxState, disabled?: boolean}
     `}
   `}
 
-  ${({visualState, disabled, theme: { styles }}) => visualState === CheckboxState.On && css`
+  ${({visualState, disabled, parentHover, theme: { styles }}) => visualState === CheckboxState.On && css`
     ${CheckboxOuter}{
       ${styles.form.checkbox.checked.default};
       border: none;
@@ -85,6 +91,12 @@ const Container = styled.label<{visualState?: CheckboxState, disabled?: boolean}
         border: none;
       }`
     };
+
+    ${!disabled && (parentHover !== undefined) && parentHover && css`
+      ${CheckboxOuter} {
+        ${styles.form.checkbox.checked.hover};
+      }
+    `};
 
     ${disabled && css`
       ${CheckboxOuter}{
@@ -112,10 +124,11 @@ interface IProps {
   checked?: boolean
   disabled?: boolean
   indeterminate?: boolean
+  parentHover?: boolean
   onChangeCallback?: (checked: boolean, indeterminate?: boolean) => void;
 }
 
-const Checkbox : React.FC<IProps> = ({ indeterminate = false, disabled, checked = false, onChangeCallback }) => {
+const Checkbox : React.FC<IProps> = ({ indeterminate = false, disabled, checked = false, parentHover, onChangeCallback }) => {
 
   const [ isChecked, setIsChecked ] = useState<boolean>(checked);
   const [ visualState, setVisualState ] = useState<CheckboxState>(checked ? CheckboxState.On : CheckboxState.Off);
@@ -143,7 +156,7 @@ const Checkbox : React.FC<IProps> = ({ indeterminate = false, disabled, checked 
   const iconWeight : number = dimensions.icons.weights['regular'];
 
   return (
-    <Container onChange={customOnChange} {...{indeterminate, disabled, visualState}}>
+    <Container onChange={customOnChange} {...{indeterminate, disabled, visualState, parentHover}}>
       <CheckboxOuter>
         <CheckboxInner>
           {visualState === CheckboxState.On ? <IconWrapper color='inverse'><CheckMark color='inverse' stroke='inverse' size={12} weight={iconWeight} /></IconWrapper> : null}
--- a/src/Form/atoms/FilterButton.tsx
+++ b/src/Form/atoms/FilterButton.tsx
@@ -1,0 +1,136 @@
+import React, { ButtonHTMLAttributes} from 'react';
+import styled, { css } from 'styled-components';
+import { TypeButtonSizes } from '..';
+import { resetButtonStyles } from '../../common';
+import { dimensions } from '../../themes/common';
+import Icon, { IconWrapper } from '../../Icons/Icon';
+
+
+const StyledButton = styled.button<{size: TypeButtonSizes, fontSize: string,  isOpen?: boolean, hasFlipArrow?: boolean }>`
+  ${resetButtonStyles};
+  background-color: hsl(200, 23%, 97%);
+  border-radius: 3px;
+  border: solid 1px hsl(207, 16%, 86%);
+  box-shadow: 0 4px 9px 0 hsla(204, 22%, 67%, 0.07);
+  color: hsl(0, 0%, 50%);
+
+  ${({ theme, size, fontSize }) => theme && css`
+    font-family: ${theme.fontFamily.ui};
+    font-size: ${fontSize};
+    height: ${theme.dimensions.filterBar.button[ size ].height};
+    transition: opacity ${theme.animation.speed.normal} ${theme.animation.easing.primary.easeOut};
+
+    ${IconWrapper} {
+      padding: ${theme.dimensions.filterBar.button[ size ].iconPadding};
+      display: flex;
+      align-items: center;
+
+      [stroke]{
+        stroke: ${theme.colors.icons.dimmed};
+      }
+    }
+
+    &:hover:enabled, &:active:enabled {
+      background-color: hsl(205, 100%, 72%);
+      border: solid 1px hsl(205, 100%, 72%);
+      color: ${theme.colors.icons.inverse};
+
+      ${IconWrapper} {
+        [stroke]{
+          stroke: ${theme.colors.icons.inverse};
+        }
+      }
+    }
+
+    &:disabled {
+      cursor: not-allowed;
+      opacity: 50%;
+    }
+`};
+
+  ${({ theme, isOpen, hasFlipArrow }) => isOpen && hasFlipArrow && css`
+    background-color: hsl(205, 100%, 72%);
+    border: solid 1px hsl(205, 100%, 72%);
+    color: ${theme.colors.icons.inverse};
+
+    ${IconWrapper} {
+      [stroke]{
+        stroke: ${theme.colors.icons.inverse};
+      }
+    }
+  `};
+`;
+
+const InnerContainer = styled.div`
+    display: flex;
+    align-items: center;
+`;
+
+/**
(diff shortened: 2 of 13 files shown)


src/Form/atoms/Checkbox.tsx
```diff
@@ -51,11 +51,11 @@ const IconWrapper = styled.div<{color: ISvgIcons['color']}>`
   }
 `;
 
-const Container = styled.label<{visualState?: CheckboxState, disabled?: boolean}>`
+const Container = styled.label<{visualState?: CheckboxState, disabled?: boolean, parentHover?: boolean}>`
```
atomworks, line 54: This is a heavy way to handle hover states from parents. What context is this being used in?

The browser behaviour defaults should mean that anything wrapped in the semantically correct way should pass the hover state to the input. If that isn't working it's better to address the problem there first and get the CSS working as intended. Using props for hover should be a last resort.

storybook/src/stories/Form/Input/RadioButton.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+    <Container>
+      <Label htmlFor={'option1'} labelText={'Option 1'}>
+        <RadioButton {...{ name, disabled, currentChecked }} id={'option1'} value='option1' onChangeCallback={handleChange}/>
+      </Label>
+      <Label htmlFor={'option2'} labelText={'Option 2'}>
+        <RadioButton {...{ name, disabled, currentChecked }} id={'option2'} value='option2' onChangeCallback={handleChange}/>
+      </Label>
+      <Label htmlFor={'option3'} labelText={'Option 3'}>
+        <RadioButton {...{ name, disabled, currentChecked }} id={'option3'} value='option3' onChangeCallback={handleChange}/>
+      </Label>
```
atomworks, line 38: This is the layout for text / select / password style inputs but not best for radio buttons. The layout of text for radio buttons (and checkboxes) should be text to the right.

storybook/src/stories/Form/FilterDropdown.stories.tsx
```diff
@@ -0,0 +1,239 @@
...
+import { boolean, text, select, number } from "@storybook/addon-knobs";
+import { action } from '@storybook/addon-actions';
+import {
+  FilterDropdown,
+  IFilterDropdownValue,
+  PageHeader,
+} from 'scorer-ui-kit';
+
+export default {
+  title: 'Form/atoms',
```
atomworks, line 13: Can we move filter specific parts to the `Filters` section as they'll always have that association but aren't really an input form in the same way as other things here.

src/Form/atoms/FilterButton.tsx
```diff
@@ -0,0 +1,136 @@
+import React, { ButtonHTMLAttributes} from 'react';
+import styled, { css } from 'styled-components';
+import { TypeButtonSizes } from '..';
+import { resetButtonStyles } from '../../common';
+import { dimensions } from '../../themes/common';
+import Icon, { IconWrapper } from '../../Icons/Icon';
+
+
+const StyledButton = styled.button<{size: TypeButtonSizes, fontSize: string,  isOpen?: boolean, hasFlipArrow?: boolean }>`
```
atomworks, line 9: I appreciate the forward thinking but we don't need any size variations at this point in time. As they're not styled to fully scale, it would be better to remove this and `fontSize` until we have a specific use case.

storybook/src/stories/Form/Buttons/FilterButton.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import {IconSVGs} from '@future-standard/icons';
+import { text, select, boolean } from '@storybook/addon-knobs';
+import { FilterButton } from 'scorer-ui-kit';
+import styled from 'styled-components';
+
+export default {
+  title: 'Form/Buttons',
```
atomworks, line 8: Could we move this to the `Filters` section instead of having it in `Form` please.

storybook/src/stories/Form/Buttons/FilterButton.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export const _FilterButton = () => {
+  const iconList = generateIconList();
+
+  const buttonText = text('Text', 'Search Example String Field');
+  const iconName = select("Icon", iconList, Object.keys(iconList)[0]);
+  const hasFlipArrow = boolean('Has Flip Arrow', false);
+  const isOpen = boolean('is Open', false);
+  const disabled = boolean('disable', false);
+  const buttonSize = select("Size", { Xsmall: 'xsmall', Small: "small", Normal: "normal", Large: "large" }, "small");
```
atomworks, line 34: I've mentioned removing these from the component. Let's remove it from here too if we don't have a planned use.

src/Form/atoms/RadioButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+import styled, { css } from 'styled-components';
+
+const InnerRadio = styled.div`
+  width: 10px;
+  height: 10px;
+  border-radius: 50%;
+  user-select: none;
+`;
+
+const OuterRadio = styled.label<{ isChecked: boolean, disabled: boolean, parentHover?: boolean }>`
```
atomworks, line 11: Please see the previous comment related to `parentHover`.

src/Form/atoms/FilterOption.tsx
```diff
@@ -0,0 +1,126 @@
...
+  return (
+
+    <Container
+      {...{ onClick, disabled, selected, id }}
+      onMouseEnter={handleMouseEnter}
+      onMouseLeave={handleMouseLeave}
+      {...props}
+    >
+      {(optionType !== 'text') && renderOption(id, optionType, selected, disabled, value, parentHover)}
+      <StyledLabel htmlFor={id} {...{ buttonSize }}>{label}</StyledLabel>
```
atomworks, line 121: I think this may be why you're having to pass `parentHover` here to get the behaviours to work. It's going to be better to find a way to have the element render inside a `label` tag.

src/Form/atoms/FilterOption.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const [parentHover, setParentHover] = useState<boolean>(false);
+
+  const handleMouseEnter = useCallback(() => {
+    setParentHover(true);
+  }, []);
+
+  const handleMouseLeave = useCallback(() => {
+    setParentHover(false);
+  }, []);
```
atomworks, line 110: Shouldn't be needed.

src/Form/atoms/FilterButton.tsx
```diff
@@ -0,0 +1,136 @@
...
+    case 'normal':
+      return '14px';
+
+    case 'large':
+      return '16px';
+
+    default:
+      return '12px';
+  }
+}
```
atomworks, line 89: Shouldn't be needed right now

src/Form/molecules/FilterDropdown.tsx
```diff
@@ -0,0 +1,537 @@
...
+const TopLine = styled.div`
+  background-color: hsl(205, 100%, 72%);
+  height: 3px;
+  border-top-left-radius: 3px;
+  border-top-right-radius: 3px;
+`;
+
+const InnerBox = styled.div`
+border-bottom-left-radius: 3px;
+border-bottom-right-radius: 3px;
```
atomworks, line 66: Indents ;)

src/Form/molecules/FilterDropdown.tsx
```diff
@@ -0,0 +1,537 @@
...
+    if (!isItemSelected) {
+      newSelected.push(item);
+    }
+    return newSelected
+  }
+
+  return item;
+}
+
+const getVisibleList = (list: IFilterItem[], maxItems: number, selected: IFilterDropdownValue): IFilterItem[] => {
```
atomworks, line 243: It would be good to see some basic comments to more easily identify what's going on with this function.

src/Form/molecules/FilterDropdown.tsx
```diff
@@ -0,0 +1,537 @@
...
+interface IDropOpen {
+  isOpen: boolean,
+  position: IOpenPos,
+}
+interface IFilterDropdown {
+  buttonIcon: string
+  buttonText: string
+  list: IFilterItem[];
+  selected: IFilterDropdownValue;
+  buttonSize?: TypeButtonSizes
```
atomworks, line 378: Shouldn't be needed right now (and not plans for it, unless I missed some thinking  here)

src/Form/molecules/FilterDropdown.tsx
```diff
@@ -0,0 +1,537 @@
...
+    setContentState((prev) => {
+      const isOpen = false;
+      return { ...prev, isOpen };
+    })
+
+    setSearchText('');
+    setVisibleList(getVisibleList(list, maxDisplayedItems, selected));
+  }, [list, maxDisplayedItems, selected]);
+
+  useClickOutside(mainRef, handleClose);
```
atomworks, line 438: This is a little lost here amongst the functions. Can you move it up to under the state or perhaps down to next to the useEffects?

atomworks (changes requested):
This was a pretty heavy PR. I can see how everything was built together although in the future I think if we're able to break something like this down in to the three main parts, that would help a lot in reviewing.

Some requests:

- [ ] When you select an option in a dropdown, leave it open and let the user close this manually by re-clicking the filter button or clicking outside.
- [ ] For filter specific components, even though they're semantically form inputs, we should keep them in the Filters section as they're very specific to that usage.

Future thoughts:
- We may want to make debounce an option in that search input. It'll be needed in implementation but perhaps it's a filter we can add here later so it saves time when using.

Components are all looking great though. Can't wait until we can start to see these getting used in projects!

src/Form/atoms/Checkbox.tsx
isacoder: I admit this is a strange fix, but perhaps we can have a conversation on slack about the best way to approach it.

storybook/src/stories/Form/Input/RadioButton.stories.tsx
isacoder: I see, I will update.

src/Form/molecules/FilterDropdown.tsx
isacoder: I will move it along with handleClose to the top, I can't put it above that function because  uses that function. It can't be use before being declared.

isacoder:
Closing this one since https://github.com/future-standard/scorer-ui-kit/pull/169  filterBars is merged

PR #162, opened 2021-08-23
Fix  Message of no data going over line of Grouping Header

### LINKED ISSUE #144: [TypeTable] - Subheader of Grouping table goes over the message of no data
### Issue

Screen Shot 2021-07-29 at 12 00 55

isacoder:
### Requirement
This is a PR to fix the error on issue 144
closes #144  - Subheader of Grouping table goes over the message of no data 

### Implementation
The problem was that the Height Header was fixed to 50px (although because is display: table, height behaves as min-height).  Now with grouping is about 59px we need to contemplate this situation  and sometimes titles are long and have two lines which increments the height.

Now height is calculated and saved in a state and passed to the LoadingBox and the EmptyTableBox.
However notice that this is calculated on mount and re-calculated only onBreakpoint Change with the hook.

I added the hasHeaderGroups prop selection on Storybook but Storybook can't handle this well  because the Storybook wont remount the component when the hasHeaderGroups changes. 😞

I'm adding screenshoots that I took by sending false and true on the hasHeaderGroups from the beginning, to show that this will be rendered correctly although changing the prop on the fly will look like it doesn't.

#### No data with hasHeaderGroups
Screen Shot 2021-08-23 at 20 33 16

#### No data without hasHeaderGroups

Screen Shot 2021-08-23 at 20 34 27

#### Loading with hasHeaderGroups
Screen Shot 2021-08-23 at 20 35 31

#### Loading without hasHeaderGroups

Screen Shot 2021-08-23 at 20 34 34

### DIFF
--- a/src/Tables/molecules/TypeTableHeader.tsx
+++ b/src/Tables/molecules/TypeTableHeader.tsx
@@ -1,9 +1,8 @@
-import React, { useState, useCallback, Fragment, useRef, useEffect } from 'react';
+import React, { useState, useCallback, Fragment} from 'react';
 import styled, { css } from 'styled-components';
 import { TypeCellAlignment, ITableColumnConfig } from '..';
 import Checkbox from '../../Form/atoms/Checkbox';
 import TableHeaderTitle from '../atoms/TableHeaderTitle';
-import useBreakpoints from '../../hooks/useBreakpoints';
 
 const HeaderRow = styled.div`
   display: table-row;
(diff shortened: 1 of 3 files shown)


src/Tables/molecules/TypeTableHeader.tsx
```diff
@@ -1,12 +1,12 @@
...
 import styled, { css } from 'styled-components';
 import { TypeCellAlignment, ITableColumnConfig } from '..';
 import Checkbox from '../../Form/atoms/Checkbox';
 import TableHeaderTitle from '../atoms/TableHeaderTitle';
 
-const HeaderRow = styled.div<{ HEADER_HEIGHT: string }>`
+const HeaderRow = styled.div`
   display: table-row;
-  height: ${({ HEADER_HEIGHT }) => HEADER_HEIGHT};
+  height: 50px;
```
isacoder, line 9: height behaves like min-height on table elements, so this one can stay like 50px;

atomworks (approved):
👍

PR #163, opened 2021-08-24
Add missing disable values of theme of SmallInput

### LINKED ISSUE #118: [Storybook] In Form -> Input -> Small Input, error appears when State=Disabled.
Originally reported by @sasakisasaki in BugHerd. At first look, this issue _seems_ to be the result of a missing property in the theme.

image

isacoder:
### Requirements

closes #118 
[Storybook] In Form -> Input -> Small Input, error appears when State=Disabled. 

### Implementation
It was failing because disable focus was assigned but missing in theme.
Copied the values of disable normal.

No disable assignation value provided in storybook for the input value, although it doesn't break, the visual value would not appear correctly when disable "state" was provided. Also added the "not allowed" cursor when disable was true.

### Screenshoots
Screen Shot 2021-08-24 at 12 56 50

### DIFF
--- a/src/Form/atoms/SmallInput.tsx
+++ b/src/Form/atoms/SmallInput.tsx
@@ -32,6 +32,9 @@ const StyledInput = styled.input<{ fieldState : TypeFieldState, padRight?: numbe
     padding-right: ${p.padRight + 17}px;
   `}
 
+  &:disabled {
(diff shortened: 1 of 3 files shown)


atomworks (approved):
👍

PR #167, opened 2021-08-26
TabsWithIcon feature

### LINKED ISSUE #161: [Tabs] Add detailed tab design variant
The artwork for this feature can be seen here:
zpl://screen?pid=5fb77e662d5c19b7d12dca42&sid=611b47d1d7725132623790c3

isacoder:
### Requirement

closes #161 - Add new Design Tabs variant

### Implementation

This PR has two new components 
- TabWithIcon 
- TabsWithIconBar

I have added paddingLeft for TabsWithIconBar because I notice that the left part needs to be the same that the rest of the content but not sure if this is usually the way to show it. I know that there is sometimes a global container with max-width.

I didn't added a hover color since in the other tabs examples there was none and also was not specified but if you want to add it please let me know and the expected colors.

Updated Folders for Tabs

#### TabWithIcon
Screen Shot 2021-08-26 at 19 38 01

#### TabsWithIconBar
Screen Shot 2021-08-26 at 19 34 09


atomworks (approved):
👍

PR #177, opened 2021-09-15
Add AreaUploadManager with Style

### LINKED ISSUE #164: [File Input] Add generic drop / select variant
The artwork for this feature is available here:
zpl://screen?sid=611b47cc19f62011248466bb&pid=5fb77e662d5c19b7d12dca42

It just needs to be that dropzone although you can see from the second screen that there will need to be an alternate layout. I think the content and the height of the element need to have a lot of flexibility but the drag and drop interface and the ability to return a list of files is important.

The first use of this feature is intended to:
• Handle a large amount of files with restricted types (jpg & csv in the case)
• Show a different layout after the initial file selection (this isn't a feature needed in the component itself but we will need a way to manage it externally)

We can chat more about this as needed, just let me know.

isacoder:
### Requirements

closes #164 Add Generic drop select variant

This uploading manager component allows the user to drop or select multiple files and if the Developer specifies types will reject files that are non compliant to the types.

The state is handled in the component for the files so if more files are added the files the files will be reviewed for duplicates based on the names.

### Implementation

**Icons**
I have used Icons for the files that are not the ones in Zeplin but are available in the current UI Kit.
I could try using local or private options just to test this icons, but didn't wanted to stop sending the PR before asking.
I tried to make similar the size of the icons to the design but didn't worried to make it pixel perfect because these are not the design icons. The size weight issue is very obvious in this occasion, the design is slimmer but they don't look bad.

**Loading button**
I just realized that the Input File button is a normal button without loading.. I'm thinking that this case of many files might be worth to do the update to have the loading state. Please let me know if you want this change or if this should be handle at page level or there was another design solution in mind.

#### First Screen
Screen Shot 2021-09-29 at 21 50 42

#### Second Screen
Screen Shot 2021-09-29 at 21 50 49


atomworks (commented):
There's one small code comment but otherwise it all looks good. I do have one request though in terms of how we're handling that section with the text and icons in the upload area. Could we make this whole section overridable with a component passed in as a prop?

I think there will be a lot of different variations re-using this component so having that flexibility will be really valuable. If you prefer, we could do that in a follow up PR but if it's easy to put in to this one then that would be great. Just let me know what you think!

atomworks (changes requested):
Changes look great. Because `BigIconsSummary` is a part of a specific design pattern I don't think we should expose it in the UI kit so if we can remove it from `index.ts` and drop the story too that would be great. That way we retain the ability to iterate and deprecate it later as the design patterns evolve as people won't have started using it for other things.

PR #183, opened 2021-09-23
Update StatusIcon with or without status has same size

isacoder:
### Description
Status icon is smaller than expected in real cases. 
There was a change request to have the same size regardless if it will have a Status Dot or not.

### Screenshots
Before
Screen Shot 2021-09-23 at 21 30 06

After
Screen Shot 2021-09-23 at 21 25 42


atomworks (approved):
That's great, thank you!

I was about to say I wish all PRs were this long but then I would have so many PRs to check. 😄

Issue #181, opened 2021-09-23
[Filter Bar] Minor Feature Iterations


atomworks:
Here's a few things I'd like to address in a future iteration of the filter bar feature. I've kept them grouped for now but we can break it apart to other issues when we review these to include in an upcoming sprint. These are quick notes so I can elaborate more when we're able to address there. Also, many of these are design related thoughts rather than comments on implementation.

- [ ] Add a way to close additional string based search fields. (perhaps a close button at the end of the field, or simply on blur if it's empty we hide it again). 
- [ ] Remove the filter from one of the dropdown in examples 
- [ ] Placeholder font in search should be a little lighter.
- [ ] Add the date picker component we have to the filter bar / examples
- [ ] Create an animation to draw awareness to specific new field when new string search being added

PR #194, opened 2021-10-12
Feature/search filters iteration

### LINKED ISSUE #201: Filterbar - Basic Search - result breaks with up to 255 characters
If we enter long search text in SearchField, then it breaks the FilterResult UI. SearchInput Field may have length up to 255 characters

Screen Shot 2021-10-22 at 17 49 49

isacoder:
### Description 

This pr is related to #181 -  Add a way to close additional string based search fields. (perhaps a close button at the end of the field, or simply on blur if it's empty we hide it again).

Update: 28 oct

fix #201 

 I added the close button for the hide inputs as requested.
 
 I have added to this PR some of the things we talked during last call.
 
 - BasicSearchInput background is now white background default and you can send noBackground prop to remove the background.
 - Made a quick implementation of a width transition when a new input is added, not sure it looks as expected.
 
 ### Reviewing/Testing steps 
 
 This can be reviewed in FilterInputs and Filterbar by clicking in the more options then click on the search input option.
 
Screen Shot 2021-10-12 at 20 23 47

### DIFF
--- a/src/Filters/molecules/FiltersResults.tsx
+++ b/src/Filters/molecules/FiltersResults.tsx
@@ -36,6 +36,10 @@ const FilterLabel = styled.div`
 `;
 const FilterLabelText = styled.div<{hasIcon?: boolean}>`
   padding: ${({hasIcon}) => hasIcon ? '0 15px 0 9px' :'0 15px 0 0' };
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  overflow: hidden;
+  max-width: 300px;
 `;
 
 const ClearTextButton = styled.button`


src/Filters/molecules/FilterInputs.tsx
```diff
@@ -1,15 +1,50 @@
...
   border-radius: 3px;
   flex: 1 1 200px;
 `;
 
+const CloseSearchButton = styled.button`
+  ${resetButtonStyles};
+  flex-shrink: 0;
+  flex-grow: 0
+  flex-basis: auto;
+  border-left: solid 1px hsl(207,16%,86%);
```
atomworks, line 29: Just the X by itself looks great so let's drop this border.

src/Filters/molecules/FilterInputs.tsx
```diff
@@ -1,15 +1,50 @@
...
+import { resetButtonStyles } from '../../common/index';
+
+const fadeInAnimation = keyframes`
+  0% {
+    width: 50%;
+  }
+  100% {
+    width: 100%;
+  }
+`;
```
atomworks, line 16: Animation looks great. Could we add a little opacity to the transition too?

I think for the whole thing might be too much so perhaps at 0% `opacity: 0;` is good and then maybe hit `opacity: 1` at around 50% could be good.

atomworks (changes requested):
Some minor bugs I noticed that are harder to relate to code:

- [ ] The string fields that are added are a couple of px too tall.
- [ ] The string field searches seem to be missing the same subtle drop shadow as the buttons.

---

There's two more major things I'd like to revise too, we can chat about these more as needed. If they are easy to do in this PR we can but *they're not actually related to the work in this PR*, it's just things I've noticed while reviewing this that I'd previously not noticed or thought about properly.

*Input field widths* - It's a really nice touch to have these flex but we should have an upper limit on the width. When we have a UI with limited options and especially when viewed on a higher res screen, these searches will overly big. I think about 320px is a good size limit for these.

*Multi-filtering* - Multi-filtering should be the default behaviour (even if that means a code breaking change). It solves a problem we have specifically with our APIs currently but it's not a universal thing and shouldn't really be an established pattern we accept. 

Can we currently disable filters manually via props? I think as there is a certain complexity to how filters stack in many cases it may be better to just allow this and leave it to the responsibility of the engineer implementing to manage this for their specific use case.

isacoder:
> **Multi-filtering** - Multi-filtering should be the default behaviour (even if that means a code breaking change). It solves a problem we have specifically with our APIs currently but it's not a universal thing and shouldn't really be an established pattern we accept.
> 
> Can we currently disable filters manually via props? I think as there is a certain complexity to how filters stack in many cases it may be better to just allow this and leave it to the responsibility of the engineer implementing to manage this for their specific use case.

I think I can update this in the DatePicker, since it's becoming a big refactor and I'm updating the disable  part as well.

isacoder:
I moved the close button to the BasicSearcher, box-sizing was not solving my overflow parent issue .. in any case sense we will probably use it as a clear input pattern perhaps is ok :)

PR #197, opened 2021-10-14
Refactor FilterBars to allow Dropdown Date Picker

isacoder:
### Description

This PR is related to #181 - Add the date picker component we have to the filter bar / examples
This PR fixes   https://github.com/future-standard/scorer-ui-kit/issues/202 - FilterBar - Filters config default values does not update

 **Questions**
 - DatePicker doesn't have a way to clear the current date beyond the results clear bar ( like a button of clear). Currently clearing can only happen from the Results bar (and unfortunately the only way I was able to achieve it is umounting the picker 😞 ) . Do you want to update the UI to have a clear button or something like the searchers?.
 
 - I have updated allowMultifilter as true default. I update it this way to keep the name since I'm not sure if it was used already, so this update will potentially break only if the Dev was expecting not to add `allowMultifilter={false}`. I could have remove the prop and have one called singleFilterValue or something. What do you think is better?
  
  **Extra updates**
  - I extracted the logic of the dropdown (button and container) to FilterDropHandler component to be reused with the DatePicker or any other thing that might be needed in the future.
  - I'm allowing not to send dropdowns or searchInput configuration, previously to achieve this the Dev would have need to send empty array, now FilterBar allows to leave them undefined. 

  - Fixed the missing default assignations. I tried to work around the BaseComponent props like value and initialValue while supporting the ideal selected from Filter Results.(#202 fix)
  
  - Refactor of IFilterResults. It took me a while to find a way to keep the all the filters in one state, at the end I modified the IFilterResults interface to allow different values. This could be potentially breaking, since the onChangeCallback needs to support the new Date and DateInterval values, but I think it's worth it and if there are other values added in the future this can be updated too. Definitely the interfaces/typing have been the greatest challenge in the FilterBar feature.
 
 ### Reviewing/Testing steps 

 You can review this one in FilterBar story
 However since the data is sent through the sample_data, not the object in story book, changes in the object can only be tested running the project and changing the code. I have updated some commented selects to test the default values.
 
 Sorry for the inconvenience, since this has been delayed I didn't want to put more time in finding a way to handle the jp and eng data that can be updated from the storybook object
 
 Base
Screen Shot 2021-10-28 at 10 32 12

Selected

Screen Shot 2021-10-28 at 10 32 27


src/Filters/molecules/FiltersResults.tsx
```diff
@@ -61,13 +64,50 @@ const FilterLabelsGroup = styled.div`
...
 };
 
+const renderLabel = (item: IFilterItem | DateInterval | Date, icon?: string, filterName?: string, resultsDateFormat?: string) => {
+
+  let textLabel: string = "";
+  const isDateFormatValid = resultsDateFormat ? validateDateFormat(resultsDateFormat) : false;
+
+  if (filterName && isFilterItem(item)) {
+    textLabel = `${filterName}: ${item.text}`;
+  } else if (filterName && item instanceof Date) {
```
atomworks, line 93: Can we break these across a few lines. It’s longer so hard to read and the : within the values make it hard to see the one that is used in the ternary. I thought it was a nested ternary at first.

src/Filters/organisms/FilterBar.tsx
```diff
@@ -45,16 +53,16 @@ const createDropdownFilters = (
...
+  handleDropdownsSelected: (newValue: IFilterValue, filterId: string) => void
 ): IFilterDropdownExt[] => {
 
   const dropdownFilters: IFilterDropdownExt[] = [];
   dropdownsConfig.forEach(dropdown => {
     const filter = filtersValues.find(filter => filter.id === dropdown.id);
     if (filter) {
-      const selected = filter.selected;
-      const onSelect = (newSelection: IFilterValue) => { handleSelected(newSelection, filter.id); };
-      let disabled = getDisableValue(filtersValues, allowMultiFilter, filter);
```
atomworks, line 57: Having a || followed by a ternary in the same line is anpittle confusing. Is there a clearer way to define things here?

src/Filters/organisms/FilterBar.tsx
```diff
@@ -134,50 +188,89 @@ const createLabelResults = (
...
+      id,
+      type: 'dropdown',
+      selected: (isFilterItem(selected) || Array.isArray(selected)) ? selected : null,
+    };
+
+    newFilters.push(initialDropdown);
   });
+
+  datePickersConfig.forEach(({ id, initialValue, selected }) => {
+
```
atomworks, line 254: Do we need to explicitly set this to null?

src/Filters/organisms/FilterBar.tsx
```diff
@@ -339,11 +459,12 @@ const FilterBar: React.FC<IFilterBar> = ({
           showLessText,
         }}
         searchFilters={createSearchers(searchersConfig, filtersValues, allowMultiFilter, handleSearchers)}
-        dropdownFilters={createDropdownFilters(dropdownsConfigRef.current, filtersValues, allowMultiFilter, handleSelected)}
+        dropdownFilters={createDropdownFilters(dropdownsConfigRef.current, filtersValues, allowMultiFilter, handleDropdownsSelected)}
+        datePickFilters={datePickersConfig ? createDatePickers(datePickersConfig, filtersValues, allowMultiFilter, handleDatePickers) : undefined}
```
atomworks, line 463: Is ‘datePick’ correct or should this be pickers for consistency?

atomworks (changes requested):
Small things mostly. I’ll edit it once I’m in the office so the comments aren’t so rough!

src/Filters/molecules/FiltersResults.tsx
atomworks: Sorry, this should be line 94! (and also the same on 96)

src/Filters/molecules/FiltersResults.tsx
```diff
@@ -61,13 +64,50 @@ const FilterLabelsGroup = styled.div`
...
+    textLabel = isDateFormatValid && resultsDateFormat ? `${filterName}: ${format(item, resultsDateFormat)}` : `${filterName}: ${item.toDateString()}`;
+  } else if (filterName && isDateInterval(item)) {
+    textLabel = isDateFormatValid && resultsDateFormat ? `${filterName}: ${format(item.start, resultsDateFormat)} - ${format(item.end, resultsDateFormat)}` : `${filterName}: ${item.start.toDateString()} - ${item.end.toDateString()}`;
+  } else if (!filterName && isFilterItem(item)) {
+    textLabel = item.text;
+  } else if (!filterName && item instanceof Date) {
+    textLabel = isDateFormatValid && resultsDateFormat ? format(item, resultsDateFormat) : item.toDateString();
+  } else if (!filterName && isDateInterval(item)) {
+    textLabel = isDateFormatValid && resultsDateFormat ? `${format(item.start, resultsDateFormat)} - ${format(item.end, resultsDateFormat)}` : `${item.start.toDateString()} - ${item.end.toDateString()}`;
```
atomworks, line 102: Can we break these across a few lines. It’s longer so hard to read and the : within the values make it hard to see the one that is used in the ternary. I thought it was a nested ternary at first.

src/Filters/organisms/FilterBar.tsx
```diff
@@ -45,16 +53,16 @@ const createDropdownFilters = (
...
 ): IFilterDropdownExt[] => {
 
   const dropdownFilters: IFilterDropdownExt[] = [];
   dropdownsConfig.forEach(dropdown => {
     const filter = filtersValues.find(filter => filter.id === dropdown.id);
     if (filter) {
-      const selected = filter.selected;
-      const onSelect = (newSelection: IFilterValue) => { handleSelected(newSelection, filter.id); };
-      let disabled = getDisableValue(filtersValues, allowMultiFilter, filter);
+      const selected = Array.isArray(filter.selected) || isFilterItem(filter.selected) ? filter.selected : null;
```
atomworks, line 63: Having a '||' followed by a ternary in the same line is a little confusing. Is there a clearer way to define things here?

atomworks (changes requested):
A couple of comments from the previous review were on the wrong line so I'm re-adding them.

atomworks:
In regards to these two comments:

> DatePicker doesn't have a way to clear the current date beyond the results clear bar ( like a button of clear). Currently clearing can only happen from the Results bar (and unfortunately the only way I was able to achieve it is umounting the picker 😞 ) . Do you want to update the UI to have a clear button or something like the searchers?

That's a good idea. Let's not worry about it for now and put in on the UI Kit issues.


> I have updated allowMultifilter as true default. I update it this way to keep the name since I'm not sure if it was used already, so this update will potentially break only if the Dev was expecting not to add allowMultifilter={false}. I could have remove the prop and have one called singleFilterValue or something. What do you think is better?

Personally my preference would be to have every filter so it can be switched on or off via props so in the case of a limited set of filter rules it can be customised. So in that regard I'd say remove it entirely in favour of that. But I'm not sure how long that would take but I think if we're going to do it sooner is better than later due to the extra work required to upgrade.

src/Filters/FilterTypes.ts
```diff
@@ -1,9 +1,19 @@
...
+
+type IFilterItem = { text: string; value: string | number; }
+type IFilterValue = IFilterItem | IFilterItem[] | null;
+type IFilterType = 'search' | 'dropdown' | 'datepicker';
+
+// Type checking for IFilterItem
 // https://stackoverflow.com/questions/14425568/interface-type-check-with-typescript
-export const isFilterItem = (item: any): item is IFilterItem => {
-  if (item === null) { return false; }
+const isFilterItem = (item: any): item is IFilterItem => {
```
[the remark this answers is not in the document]
isacoder: In this case like we had a conversation this is a type guard, I'm expecting anything to determine if the received value/reference covers the expected.
https://www.typescriptlang.org/docs/handbook/advanced-types.html

src/Filters/atoms/FilterDropHandler.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  }, [onCloseCallback, openState.isOpen]);
+
+  useClickOutside(mainRef, handleClose);
+
+  const handleToggleOpen = useCallback((min_width: number, min_height: number) => {
+    if (!buttonWrapperRef.current) { return; }
+
+    const buttonRect = buttonWrapperRef.current?.getBoundingClientRect();
+    if (!buttonRect) { return; }
```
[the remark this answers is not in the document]
isacoder: I can remove the check in 125, I think it was auto completed, but basically line 123 will check that there is an assigned value and 125 will get the value of the rect and assign it to a variable. 

It might look like I'm overdoing it but It's necessary again to verify that we are able to get the rect.

src/Filters/molecules/DatePicker.tsx
```diff
@@ -211,29 +226,46 @@ const DayGuide: string[] = [
   "S", "M", "T", "W", "T", "F", "S"
 ];
 
-interface DateInterval {
+export const isDateInterval = (value: any): value is DateInterval => {
```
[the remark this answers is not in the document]
isacoder: Same case that isFilterItem, is a type guard function.
https://www.typescriptlang.org/docs/handbook/advanced-types.html

src/Filters/molecules/FiltersResults.tsx
isacoder: I have updated the logic to be easier to understand.

src/Filters/organisms/FilterBar.tsx
isacoder: I will update with identation

src/Filters/organisms/FilterBar.tsx
isacoder: Thanks for pointing out, I missed that.

isacoder:
I have updated Prop from allowMultifilter to SingleFilter in FilterBar to keep the logic of been default Multifilter and require to add something to make it singleFilter.

Just as a reminder I want to mention that any specific filter can be disabled directly in the specific configuration.

atomworks (approved):
All looks good. Sorry for the delay on the review!

PR #217, opened 2021-11-17
Fix missing display block in spans Label

### LINKED ISSUE #216: [Label] - Bring back the margins bellow labels.
### Description
Bug reported in missing styles on labels.

https://github.com/future-standard/scorer-ui-kit/pull/169/files

isacoder:
### Description

This PR fixes #216 - Missing bottom margin in label

 I made an update in Filter Bars to have Labels that could be right align but I missed a display:block property in the span that previously had, now is fixed and all is retro-compatible.

 ### Screenshoot
 
You can review this in Text Input component
Here you can see the Fix and how it's behaving currently.

Screen Shot 2021-11-17 at 10 24 21

### DIFF
--- a/src/Form/atoms/Label.tsx
+++ b/src/Form/atoms/Label.tsx
@@ -22,13 +22,13 @@ export const StyledLabel = styled.label<{ rightAlign: boolean }>`
 `;
 
 const LabelText = styled.span<{ rightAlign: boolean }>`
-${({ rightAlign }) => rightAlign
-    ? `
-      margin-left: 12px;
-      `
-    : `
-      display: block;
-      margin-bottom: 10px;
+  display: block;
+  ${({ rightAlign }) => rightAlign
+      ? `
+        margin-left: 12px;
+        `
+      : `
+        margin-bottom: 10px;
       `
   }
 `;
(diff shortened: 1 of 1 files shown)


src/Form/atoms/Label.tsx
```diff
@@ -22,8 +22,15 @@ export const StyledLabel = styled.label<{ rightAlign: boolean }>`
...
 
 const LabelText = styled.span<{ rightAlign: boolean }>`
-  margin-bottom: ${({ rightAlign }) => rightAlign ? '0px' : '10px'};
-  margin-left:  ${({ rightAlign }) => rightAlign ? '12px' : '0px'};
+${({ rightAlign }) => rightAlign
+    ? `
+      margin-left: 12px;
+      `
+    : `
+      display: block;
```
atomworks, line 30: Is there any reason why we can't just add the `display: block` as a standard rule, rather than based off `rightAlign` condition? Seems like it would be better to keep it consistent.

isacoder: No problem, I have updated display block outside of the condition.
works for both cases.

Right align
Screen Shot 2021-11-17 at 13 14 40

Common case with margin bottom.
Screen Shot 2021-11-17 at 13 15 02

atomworks (approved):
👍

PR #225, opened 2021-11-19
Fix submenu getting max-height 0 in language transition

### LINKED ISSUE #224: [Menu - Navigation Item] - Submenu is missing when translation
### Description
It was found that when the value of submenu changes and it's suddenly empty the information is not displayed again.

Screen Shot 2021-11-19 at 16 43 29 copy

isacoder:
### Description

 closes #224 - SubMenu missing after translation

There is a ref to the div of the submenu and it seems that it becomes 0 when it changes from one language to another.
I have added a condition to prevent to be updated to 0, but want to do beta test to verify.

Screen Shot 2021-11-19 at 16 44 27


JoshLipps (approved):
I am not sure I totally understand why this works. But if you have tested and it works i think this is G2G

PR #228, opened 2021-11-26
Initial global component dark theme

atomworks:
This PR introduces proper dark theming to global components and is formally the first step to being able to roll out a dark theme to the UI kit and for use in upcoming  projects.

Currently this is work in progress as I would like to tweak how some of the integrations are being used and polish some small details although it is essentially functional.


JoshLipps:
https://github.com/future-standard/scorer-ui-kit/releases/tag/v1.8.0-beta.6
Pushed to npm on this beta version.

JoshLipps (approved):
This looks good for initial pass to get the base theme set.

src/themes/dark/styles.ts
```diff
@@ -0,0 +1,49 @@
+import { colors } from './colors';export const styles = {
```
isacoder, line 1: same line, if possible change before merge

src/themes/light/styles.ts
```diff
@@ -1,4 +1,4 @@
-export const styles = {
+import { colors } from './colors';export const styles = {
```
isacoder, line 1: same line, if possible change before merge

isacoder:
Over all looks good, I understand some theming colors will stay ouside the zeppling scope and will be handle in js files however will get the values from the theme that way we can verify that all components will have css values from theme in a more obvious way.

src/themes/dark/styles.ts
atomworks: For some reason the line breaks my code generates don't show up here 😭 

If you look at the raw file you can see the breaks but not here for some reason. Tried a few things but no change so I think we may have to leave it for now.

atomworks: Wait... no, I screwed up. The new line breaks I try work... I just didn't compile my sketch plugin after I updated it and so I was just recompiling the same thing all the time.

JoshLipps:
This is missing many commits from the final branch used in beta.

JoshLipps:
I think this PR should be canceled in favor of the new final branch.

PR #231, opened 2021-12-01
Update for Date picker to receive text for titles

### LINKED ISSUE #229: [Datepicker] 'From' label hard coded but needs to be definable.
Currently if the Date Picker is set to Date Mode: `single` and Time Mode: `single` then the top left item says "From". However this needs to be changeable for different use cases.

At the moment we have a project that uses two fields to define a date range. When opening the To portion of the Datepicker, this value still reads From as it seems to be hard coded. 

We need a way to either allow the user to choose "To" or "From" for single Time Mode settings or perhaps use the defaults as we currently have but allow for these to be overridden with a string in the props. This may be useful if perhaps we want to label it something like "Reminder Time" for example.

You can recreate the current issue problems in Storybook easily. 
https://future-standard.github.io/scorer-ui-kit/storybook/?path=/story/filters-molecules--date-picker

isacoder:
### Description

closes #229 

[Datepicker] 'From' label hard coded but needs to be definable. 

 I have chosen the overriding option. Variables are descriptive but long.  I request your opinion on this, I could change it for title1 and tittle2 if you it's better.
 
I also updated the JSX parameters for the plugging perhaps I can do this every time I update a component and little by little it will become tidy.

 ### Screenshoots
 
Screen Shot 2021-12-01 at 12 32 01

JSX Before
Screen Shot 2021-12-01 at 12 30 59

JSX After

Screen Shot 2021-12-01 at 12 30 29

### DIFF
--- a/src/Filters/molecules/DatePicker.tsx
+++ b/src/Filters/molecules/DatePicker.tsx
@@ -251,12 +251,16 @@ export interface IDatePicker {
   dateMode?: DateMode
   timeMode?: TimeMode
   hasEmptyValue?: boolean
+  dateTimeTextUpside?: string
+  dateTimeTextDownside?: string
   updateCallback?: (data: DateInterval | Date) => void
 }
 
 const DatePicker: React.FC<IDatePicker> = ({
   dateMode = 'interval',
   timeMode = 'interval',
+  dateTimeTextUpside = 'From',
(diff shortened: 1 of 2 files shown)


atomworks (approved):
That's great, thank you!

PR #238, opened 2021-12-14
Fix: bug in tab component when onClick gets washed out by local onclick

JoshLipps:
### Description

Fix but in tab component where `onClick` is not passed down to tab component. 
Also exported Tabs context so we can make custom tab components that subscribe to the context.

 ### Reviewing/Testing steps

Test Tab with onClick


fti-dshelke (approved):

isacoder (approved):
LGTM 👍

PR #240, opened 2021-12-20
Show arrow Icon with line

fti-pvishwakarma:
### Description

The Arrow Icon with Line is required for Okinawa project's traffic counter settings.

Zeplin design: https://app.zeplin.io/project/616e925cbde941134b0e7cbd/screen/61a4ad9c97d4ac134c19227d
image

As discussed here the first approach is implemented https://fti-futurestandard.slack.com/archives/C02J9NMM2SH/p1639125571019200

The implementation is done for **LineUI** component only. Once the changes are approved then we will make similar changes for **LineUIVideo** and **LineUIRTC** components.

 ### Reviewing/Testing steps
Follow the steps given at :
https://github.com/future-standard/scorer-ui-kit/tree/feature/line-ui-arrow-icon#readme

When project runs, Open http://localhost:3000/scorer-ui-kit#/line in browser you will see

Example


src/LineUI/index.ts
```diff
@@ -19,6 +19,8 @@ export interface IPointSet {
   showOrientation?: boolean;
   styling?: string;
   areaName?: string;
+  centerIcon?: React.FunctionComponent | undefined;
+  centerIconPosition?: string;
```
fti-dshelke, line 23: Make it as :
`centerIconPosition?: 'up' | 'down';`

example/src/pages/LinePage.tsx
```diff
@@ -35,21 +35,40 @@ const Line: React.FC<{}> = () => {
...
+        name: 'DOWN',
+        points: [
+            {
+              x: 200,
+              y: 400
+            },
+            {
+              x: 400,
+              y: 400
+            }
```
atomworks, line 65: Related to other comments on implementation of up / down. Once that has been resolved you may want to flip these points so the arrow appears on the opposite side of the line.

src/LineUI/LineUnit.tsx
```diff
@@ -146,6 +148,15 @@ const LineUnit : React.FC<ILineUnitProps> = (props) => {
     x: (x2 + x1) / 2,
     y: (y2 + y1) / 2
   };
+
+  const centerIconCordinates = () => {
```
atomworks, line 152: There's a typo on Coordinates (add an 'o'). You may want to rename this though when renaming the prop as commented elsewhere.

src/LineUI/index.ts
```diff
@@ -19,6 +19,8 @@ export interface IPointSet {
   showOrientation?: boolean;
   styling?: string;
   areaName?: string;
+  centerIcon?: React.FunctionComponent;
+  centerIconPosition?: 'up' | 'down' | undefined;
```
atomworks, line 23: I think this can be omitted as per other comments on implementation.

src/LineUI/LineUnit.tsx
```diff
@@ -165,11 +176,23 @@ const LineUnit : React.FC<ILineUnitProps> = (props) => {
...
+        >
+          {label}
+        </LabelText>}
+      
+      {centerIconPosition && CenterIcon &&
+        <g transform={`translate(${centerIconCordinates().x},${centerIconCordinates().y}) rotate(${(180 / Math.PI) * Math.atan2(y2 - y1, x2 - x1) + (centerIconPosition === 'up' ? 0 : 180)}) scale(${unit * 1.5})`} >
+          <g transform={centerIconPosition === 'up' ? `translate(${Math.ceil(4-unit) > 2 ? 10 : 5 - unit}, -30)` : `translate(${(-25 + 1.5 * unit) < -21 ? -28 : -25 + 1.5 * unit}, -30)`}>
+            <CenterIcon height='20' width='20' />
+          </g>
+        </g>
```
atomworks, line 195: Related to the comments on dropping reliance on `UP` or `DOWN`, this can be simplified.

atomworks (changes requested):
Thanks for adding this feature. It looks like the more complex part of this you've got right although I have some notes on the overall implementation. At the moment the way this has been approached is very case specific so we'll want to revise this to ensure that it doesn't break functionality of existing uses and can remain flexible in the future.

To sum up everything, some of these have more details in the comments:
- [ ] 1. Rename `centerIcon` to `directionMark`.
- [ ] 2. Add a prop that is passed through `LineUI.tsx` and the example to turn the direction mark on and off.
- [ ] 3. The indicator should always be on the same side.
- [ ] 4. The icon you've included as an SVG is available in the default `Icon` set. Could you try this one instead of importing a new SVG?
- [ ] 5. Easy fix - The arrow is too big. (see screenshot)

More details if needed on this:
[1] This is just to help indicate more clearly the usage. I can see how this could be used for other things in it's current form but it's better to keep this more deliberate at this time.
[2] Rather then use `UP` or `DOWN` to toggle this feature, it's better to have it as a `boolean` value. Also with this, in the example you'll see some checkboxes for toggling various features. Could you add a checkbox for the directionMark too?
[3] I've added some ascii diagrams on a code comment for help make this more clear. 
[4] Could you use the `Icon` for the arrow potentially, then you can add the circle manually using styled components to make the wrapper `<circle .../>`
[5] Just needs to be a bit smaller!
image

I know there's a lot there but I think it's a lot simpler than it looks at first look. The main mechanisms you've got at work here look great which I know can be really difficult in something complexed like this. Thank you for all your efforts on this.

Please don't hesitate to reach out to me directly on Slack if you need any clarification.

example/src/svg/arrow-icon.svg
```diff
@@ -0,0 +1,8 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20">
+    <g id="Group_320" data-name="Group 320" transform="translate(9588 20667) rotate(180)">
+        <circle id="Ellipse_144" data-name="Ellipse 144" cx="10" cy="10" r="10" transform="translate(9568 20647)" fill="#5c7b9a" />
+        <g id="master-icon-arrows-up" transform="matrix(-0.951, 0.309, -0.309, -0.951, 10551.025, 20886.299)">
+            <path id="Path_719" data-name="Path 719" d="M849.491,518.638l5.91-3.01a.328.328,0,0,1,.441.143l3.011,5.909" fill="none" stroke="#fff" stroke-linecap="round" stroke-linejoin="round" stroke-width="3" fill-rule="evenodd" />
+        </g>
+    </g>
+</svg>
```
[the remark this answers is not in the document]
fti-dshelke: There are two Icons Back and Forward in the `Icons` component but they are not filled. Shall we modify where we are using it? which Icon do you prefer to use?
image
image


one more query:
I think if we are using Icon from ui-kit Icon module then we don't need to pass the directionMark icon though state, right?

isacoder (approved):
I have review the code and run the example and it's currently working.

After Leonard's agree with the changes this is good to merge.

src/LineUI/index.ts
```diff
@@ -19,6 +19,7 @@ export interface IPointSet {
   showOrientation?: boolean;
   styling?: string;
   areaName?: string;
+  isUpDirection?: boolean;
```
atomworks, line 22: `isUpDirection` shouldn't be something managed within the UI kit components. I've added a longer explanation to the main review comment.

src/LineUI/index.ts
```diff
@@ -64,7 +66,9 @@ export interface LineUIOptions {
     y: number;
   };
 
-  boundaryOffset? : number
+  boundaryOffset?: number,
+  showDirectionMark?: boolean,
+  isUpDirection?: boolean
```
atomworks, line 71: Remove.

src/LineUI/LineUnit.tsx
```diff
@@ -146,6 +152,113 @@ const LineUnit : React.FC<ILineUnitProps> = (props) => {
...
+          labelX = (lblMargin/2);
+          labelY = 20;
+          rotatefactor = -8;
+        }
+      }
+    }
+    rotate = isUpDirection ? rotate : rotate + 180;
+    labelRotate = isUpDirection ? labelRotate : labelRotate + 180;
+    return {x, y, rotate, labelX, labelY, labelRotate, rotatefactor};
+  };
```
atomworks, line 243: Remove any code that is manipulating the label position. Let's use the current behaviour for labels.

src/LineUI/LineUnit.tsx
```diff
@@ -146,6 +152,113 @@ const LineUnit : React.FC<ILineUnitProps> = (props) => {
...
+
+  const getDirectionMarkLine = () => {
+    const dmCordinate = directionMarkCoordinate();
```
atomworks, line 246: `dmCordinate` -> `dmCoordinate`

src/LineUI/LineUnit.tsx
```diff
@@ -75,6 +76,10 @@ const GrabHandleGroup = styled.g<{ showIndex: boolean, originalRadius: number, s
   `};
 `;
 
+const Circle = styled.circle`
```
atomworks, line 79: Can we give this a more descriptive name rather than just the element name. If it's for the direction mark background then perhaps just `DMCircle` is suitable.

atomworks (changes requested):
## General Notes
There has been a lot of code added here that doesn't seem to be in response to the previous review. This has made it harder to review as now I have to distinguish between code that has changed in response to my previous review and new features that have been added. 

I felt like we were close with the first PR but this one has run away with itself a little. This is made more difficult again by the longer wait for a re-review request which could have been shortened if focused on the the feedback. New features that aren't critical should be submitted in a separate PR later.

In the future, please add a comment when you re-request a review to let us know how you've responded to feedback and (in the case of new functionality being added) tell us that something has been added and why that was done.

## Direction Handling

For this PR, there is still a lot of specificity within this format that is product related but not generic enough for the UI Kit. The direction arrow itself is potentially useful for many projects but being tied so heavily to an `UP` and `DOWN` format is very product use case specific and couldn't be used outside of that. The purpose of the `(1)-^-(2)` diagram was to allow a consistency of direction in a generic way that is more flexible for other potential use cases. 

This means you don't need `isUpDirection` because that is a use case specific property that you can easily manage through the `name` field with in the logic of the project it is being used in. It shouldn't be present in this code.

## Label Positioning

I appreciate the effort in trying to tidy up how the label operates however it has added a lot of complex code that although well intentioned, isn't suitable. I know the current implementation isn't perfect but at this time doesn't have any significant problems to require it being changed and is being kept lightweight for ease of a more formal review later.

src/LineUI/index.ts
fti-bcarpenter: isUpDirection  is removed from code.

src/LineUI/index.ts
fti-bcarpenter: Done

src/LineUI/LineUnit.tsx
fti-bcarpenter: Code from line 163 to 241 is removed.

src/LineUI/LineUnit.tsx
fti-bcarpenter: Done

src/LineUI/LineUnit.tsx
fti-bcarpenter: Done

atomworks (approved):
Thanks for pushing this one through. Looking great. Thank you both for all your hard work on this one.

PR #243, opened 2021-12-24
Feature/svg mode for icons

atomworks:
### Description

As there was a need to use icons within an SVG (for Line UI purposes) we needed to add an option to make this SVG friendly as the `div` wrapper was invalid.

As part of this, to avoid repeating code I moved the color selection to the icon SVG prop as originally intended. I also refactored the story to make the Storybook Knobs control use the same naming as props to be consistent with our standard approach.

### Known issue.
The `useTheme()` hook needed to have a type for the variable. Currently as everyone is extremely busy preparing for a demo, I've just used `any` but would suggest an actual type we could use for this.

### DIFF
--- a/src/Icons/Icon.tsx
+++ b/src/Icons/Icon.tsx
@@ -25,11 +25,6 @@ export interface IconProps {
   weight?: 'light' | 'regular' | 'heavy' | 'strong'
   color?: ISvgIcons['color']
 }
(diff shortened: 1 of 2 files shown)


isacoder (approved):
Code looks right, I was able to switch between g and div on the storybook.

fti-dshelke (approved):

PR #257, opened 2022-01-20
Feature/main menu always collapsable

atomworks:
### Description

At larger screen resolutions, we pin the menu open. This makes sense for a lot of data based design patterns we've used it for but more full screen UI based designs (such as in map dashboard centric projects) the screen space is a lot more important and so we want to retain the option to keep the menu size streamlined. 

There is a new prop called `canAlwaysPin` to control this. 

Additionally, there is a small fix included that means when the menu pinning functionality isn't present, the help and support link sits at the bottom of the menu instead of hovering.


isacoder (approved):
works :D

PR #261, opened 2022-01-21
Fix default menu open to do what it claims

JoshLipps:
### Description

Allows UI to default close on large screen when always pin is enabled

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "scorer-ui-kit",
-  "version": "1.8.0-beta.7",
+  "version": "1.7.9",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "scorer-ui-kit",
-  "version": "1.7.9",
+  "version": "1.8.0-beta.7",
```
isacoder, line 3: Don't merge agains main if you are doing for beta, please 🙏

JoshLipps: fixed

isacoder (approved):
thanks for removing the package.json  beta

PR #268, opened 2022-01-27
FilterBar autocontrolled - WIP

### LINKED ISSUE #258: [Filter Dropdown]
### Description

FilterBars are not updating selecteValues of dropdowns when config changes

this is important when refreshing the page and keeping the query. param
Screen Shot 2022-01-20 at 18 36 59

Screen Shot 2022-01-20 at 18 37 21

isacoder:
### Description
Controlled value refactor 

fixes #258

### Implementation
- Having in mind that current implementations are using objects to keep the selections, I made the abstraction for receiving an object and managing the selections as controlled value from outside.

Known bugs.
- I moved the state for debounce to the input level, If the user deletes the value from the browser is not updated as empty 😞 

- Deleting  input value removes the selections of dropdowns 😞 

### Screenshoot
I made the test implementation in the table of examples to I could easily use the react-router and translations.

Screen Shot 2022-02-04 at 11 26 32

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-
 import {
   HashRouter as Router,
   Switch,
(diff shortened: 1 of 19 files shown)


isacoder:
Closing this PR since this code will not go to production

PR #269, opened 2022-02-01
Dark Mode toggle

JoshLipps:
### Description
Add Bottom menu in drawer
- Switch theme
- Language(modify)

Zeplin Design: https://app.zeplin.io/project/61f0fb56807607a416632e18/styleguide/components?coid=61f0fbbc052106a68cc358d4

### Implementation
Light theme(Top BAr):
image

Dark theme(Global UI):
image

### DIFF
--- a/src/Global/atoms/DrawerBottomMenu.tsx
+++ b/src/Global/atoms/DrawerBottomMenu.tsx
@@ -1,6 +1,6 @@
 import React, { FC } from 'react';
 import styled from 'styled-components';
-import { Icon } from '../..';
+import Icon from '../../Icons/Icon';
 
 const Container = styled.div`
   padding: 16px 10px 14px 23px;
@@ -44,7 +44,7 @@ interface IDrawerBottomMenu {
   icon: string,
   title: string,
   subTitle: string,
-  onClickCallback?: () => void 
+  onClickCallback?: () => void
 }
 
 const DrawerBottomMenu: FC<IDrawerBottomMenu> = ({ icon = 'Settings', title, subTitle, onClickCallback = () => {} }) => {
--- a/src/Global/atoms/DrawerBottomMenu.tsx
+++ b/src/Global/atoms/DrawerBottomMenu.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 import { Icon } from '../..';
 
 const Container = styled.div`
-  padding: 18px 10px 14px 23px;
+  padding: 16px 10px 14px 23px;
   display: flex;
   flex-direction: row;
   border-top: 1px solid ${({ theme }) => theme.colors.divider};
@@ -17,7 +17,7 @@ const Container = styled.div`
 const ColumnContainer = styled.div`
   display: flex;
   flex-direction: column;
-  padding-left: 11px;
+  padding-left: 14px;
   align-items: flex-start;
 `;
 
@@ -36,7 +36,7 @@ const SelectedValue = styled.div`
   font-weight: 500;
   letter-spacing: 0.29px;
   color: rgba(87, 87, 87, 0.5);
-  padding-top: 4px;
+  padding-top: 2px;
 `;
 
 interface IDrawerBottomMenu {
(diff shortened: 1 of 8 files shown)


isacoder (changes requested):
Hello @fti-dshelke  for taking this tasks in this confusing theming process @-@,
I started this review without description, @fti-dshelke if you could provide I would be helpful.

This update has no story in storybook, usually I wouldn't mind but this made very hard to review and by trying to update storybook to review discovered the missing parts.
TopBar is not passing the props to UserMenu related for the theming, neither text nor toggle so the button is not doing anything.

Another not functionality related fix is the color of the text does not look good on black theme.
I imagine the design didn't provide a value, you can use the username theme color for text, I think this will make it look better.

Screen Shot 2022-02-01 at 19 22 57


Screen Shot 2022-02-01 at 19 23 07

Please add the storybook update.
 Theming toggle can be added as a knob action, look at buttons story it's just a click that shows in the action section.

fti-dshelke:
> Hello @fti-dshelke for taking this tasks in this confusing theming process @-@, I started this review without description, @fti-dshelke if you could provide I would be helpful.
> 
> This update has no story in storybook, usually I wouldn't mind but this made very hard to review and by trying to update storybook to review discovered the missing parts. TopBar is not passing the props to UserMenu related for the theming, neither text nor toggle so the button is not doing anything.
> 
> Another not functionality related fix is the color of the text does not look good on black theme. I imagine the design didn't provide a value, you can use the username theme color for text, I think this will make it look better.
> 
> Screen Shot 2022-02-01 at 19 22 57
> 
> Screen Shot 2022-02-01 at 19 23 07
> 
> Please add the storybook update. Theming toggle can be added as a knob action, look at buttons story it's just a click that shows in the action section.

@isacoder Thanks for reviewing.
I will update the remianing tasks

isacoder:
Hi!

I was able to review the theme change and  get the trigger of the buttons when theme and language 

The only thing I see that needs to be fixed is the color of the text for visibility as requested.
Screen Shot 2022-02-08 at 17 25 03

fti-dshelke:
> Hi!
> 
> I was able to review the theme change and get the trigger of the buttons when theme and language
> 
> The only thing I see that needs to be fixed is the color of the text for visibility as requested. Screen Shot 2022-02-08 at 17 25 03

Done

src/Global/atoms/DrawerBottomMenu.tsx
```diff
@@ -0,0 +1,65 @@
...
+  letter-spacing: 0.29px;
+  color: ${({ theme }) => theme.typography.global.mainMenu.subheader.color};
+  padding-top: 2px;
+  opacity: 0.5;
+`;
+
+interface IDrawerBottomMenu {
+  icon: string,
+  title: string,
+  selectedValue: string,
```
isacoder, line 46: Hi Dipak, 
I realized that selectedValue is a string and is only providing the text to be displayed and not the actual value, it does not provide functionality.

Maybe in this case should be more like a subtitle of the DrawerBottomMenu

isacoder (changes requested):
Sorry for this late change request, just realized after thinking how to get the DarkMode Icon to see it.

perhaps instead of just the action for the story, might be better to add the state
const [isLightMode, setIsLightMode] = useState(true)

const onThemeToggle = () => {
   setIsLightMode((prev) => !prev)
}

isacoder (changes requested):
Hello!

I have look at the changes thanks for updating the naming of the variables.
I saw you took the pain of  trying to update the light mode/dark mode text :( but is not showing.
 
Screen Shot 2022-02-15 at 17 47 05


I think you need to do it directly in the component not with the text knob.
please update it there.

Also since you have updated before the Global UI story, please update this too there to keep it consistent.

fti-dshelke:
> I think you need to do it directly in the component not with the text knob. please update it there.

But we have prop (**selectedThemeText**) to update the text.

isacoder:
> > I think you need to do it directly in the component not with the text knob. please update it there.
> 
> But we have prop (**selectedThemeText**) to update the text.

I meant the Storybook component
Screen Shot 2022-02-15 at 18 20 04

fti-dshelke:
Updated the storybook for global UI
image

isacoder (approved):
Thanks for updating all the changes requested, this branch is ready to merge.

PR #272, opened 2022-02-08
useQueryParams Hook

isacoder:
### Description
This is a utility hook to manage multiple values of the query params without altering other query params that might be used from other components.

### Implementation
I made this hook as a helper for the refactor to fix the persistency issue in Results bar, however It didn't work so at least we can keep the hook.

This PR will not fix the issue of setting the values from query params.

### Screenshoot

Added a new page FilterBarTable to show the hook functionality.

Screen Shot 2022-02-08 at 14 40 30


atomworks (approved):
This looks good to me but if @JoshLipps can take a look at this one too, that would be great as he's got a bit more experience here that willl be good to have on this review.

JoshLipps (approved):
This looks good. When trying something similar i had trouble with logs of hidden re-renders, especially when multiple components in view were trying to manage params. Did you test with 2 at once in same view?

JoshLipps (approved):

isacoder:
This feature is still not ready since I have notice the issue Josh mentioned with multiple Components. I will not merge to main .

PR #290, opened 2022-03-18
Add alignment options for Action buttons

isacoder:
### Description

There was the request to have right align the action buttons on typetable.

I'm adding the capability of having alignment prop to generate this without loosing  previous compatibility.

### Screenshoots
Screen Shot 2022-03-17 at 23 53 59


isacoder:
@atomworks I have updated the right align as default like we talked over slack 😄

atomworks (approved):

PR #297, opened 2022-04-16
[Tab component] horizontal alignment for tab name when language is Japanese

fti-pvishwakarma:
### Description

In Support Site an issue was reported for selected and normal tab. When we clicks on tab name then horizontal alignment of selected tab and normal tab was not correct when selected language is Japanese. We have fixed this issue in this PR.
Ref: https://www.bugherd.com/projects/281466/tasks/42

image

 ### Reviewing/Testing steps

Follow the steps given at :
https://github.com/future-standard/scorer-ui-kit/

When project runs, Open http://localhost:3000/scorer-ui-kit#/tabs in browser you will see
image

### DIFF
--- a/example/src/pages/TabsPage.tsx
+++ b/example/src/pages/TabsPage.tsx
@@ -20,20 +20,16 @@ const Tab1Container = styled.div`
 const Tab2Container = styled.div`
   margin-top: 20px;
 `
-const LanguageButton = styled(Button)`
-  margin-top: 20px;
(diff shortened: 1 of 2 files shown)


fti-dshelke (approved):
Looks good to me

fti-bcarpenter (approved):
LGTM

isacoder (approved):
Thanks for the fix, works as expected

PR #302, opened 2022-04-20
Fix/thumbnail improvements

JoshLipps:
### Description
Table Thumbnail improvements

-  Fix the Table Thumbnail to not block on a bunch of image or movie url validations - this is not ideal in a list view
- Add retry behavior with exponential back off to retry  image for thumbnail
- Add Medial Modal loading retries as well and add media load fail state

image


isacoder (commented):
This works correctly, my only concern is that previous implementation will verify that the link is working to preven clicking when url is broken, now you can open the thumbnail even when link is not correct for the example of - Magic Edge Cloud

You can compare with UI Kit page - https://future-standard.github.io/scorer-ui-kit/#/table

Screen Shot 2022-04-27 at 20 18 17

JoshLipps:
> This works correctly, my only concern is that previous implementation will verify that the link is working to preven clicking when url is broken, now you can open the thumbnail even when link is not correct for the example of - Magic Edge Cloud
> 
> You can compare with UI Kit page - https://future-standard.github.io/scorer-ui-kit/#/table
> 
> Screen Shot 2022-04-27 at 20 18 17

This change is intentional. the previous implementation was pretty frustrating to use. Click with no feedback. I think it would be better to open with a proper fail state, compared to never ending loading.

src/Tables/atoms/TableRowThumbnail.tsx
```diff
@@ -84,65 +86,51 @@ interface IProps {
...
+    setImgSrc(image);
+  },[image]);
 
-  useEffect(() => {
-    if ( mediaUrl && mediaType) {
-      verifyPreview(mediaUrl, mediaType);
+  useEffect(()=>{
+    if(imgRef.current && imgRef.current.complete){
+      setShowImage(true);
     }
```
JoshLipps, line 118: This allows for a bug. When src ="" or is cached error this allows for broken images. I will attempt a better solution.

isacoder (approved):
Worked as expected, thanks for the improvement 🙏

PR #304, opened 2022-04-27
Update SelectField to prevent default value having placeholder style

isacoder:
### Description

This is a bug reported in zenhub 
https://app.zenhub.com/workspaces/platform-team-61b9b760a8454700107d1a50/issues/future-standard/scorer-surveillance-dashboard/726

Select Dropdowns - Default value appearing as placeholder / disabled option

I think bug this is causing a bit of confusion in regards to usability. We just need to make sure that if it's a selected value that is shows the style correctly. At the moment the default values that are loaded are displayed how only the no selection / placeholder state should appear. 

### Screenshot

Screen Shot 2022-04-26 at 23 48 43

### DIFF
--- a/src/Form/atoms/SelectField.tsx
+++ b/src/Form/atoms/SelectField.tsx
@@ -47,7 +47,7 @@ ${({ isCompact }) => isCompact && css`
     }
   `}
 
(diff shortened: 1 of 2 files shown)


atomworks (approved):
👍

PR #307, opened 2022-05-10
Filter dropdowns shouldn't paginate

isacoder:
### Description
https://app.zenhub.com/workspaces/platform-team-61b9b760a8454700107d1a50/issues/future-standard/scorer-surveillance-dashboard/648

Filter dropdowns shouldn't paginate

At the moment the dropdown filter is limiting the choices to 5 but there are no pagination controls to see any others and it relys on knowing how to search for the remaining tags to show them.

The filter should instead allow for scrolling of this area internally after 5-7 entries are included.

### Screenshoots
https://drive.google.com/file/d/1l-Dzpm5ZuEh704DBpTFgRDqaG0Eka4eh/view?usp=sharing

### DIFF
--- a/src/Filters/molecules/FilterDropdown.tsx
+++ b/src/Filters/molecules/FilterDropdown.tsx
@@ -55,6 +55,12 @@ const StyledFilterOption = styled(FilterOption)`
 const OptionList = styled.div`
   height: 180px;
   overflow-y: scroll;
+  ::-webkit-scrollbar {  /* Hide scrollbar for Chrome, Safari and Opera */
+    display: none;
+  }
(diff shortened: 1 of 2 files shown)


atomworks (changes requested):
One very small request... at the moment the list very tidily fits in the dropdown so you can't easily see there are more options.

So it looks like this:
image

But if the last item could be half cut off (more similar to this, but just half... this is a quick copy paste!) then we can see quickly it's a scrollable area.
image

isacoder:
Updated showing half as requested
Screen Shot 2022-05-26 at 17 25 40

atomworks (approved):

PR #312, opened 2022-05-25
Fix overlaping of titles and add hideAreaInDocTitle option

isacoder:
### Description

https://app.zenhub.com/workspaces/platform-team-61b9b760a8454700107d1a50/issues/future-standard/scorer-surveillance-dashboard/654

```
Page titles are stacking up!#654
As you can see from this screenshot of the back history, every time a new area is entered the page title is added to the start rather than replacing the previous area title.

```

### Implementation

I removed the state that is adding at the end the initial docTitle. 
I notice that when the title is empty it will add the delimiter then empty  `| `
When coming from another place the state is persisted so that is why is attached at the end.
Not sure which was the case scenario where this was necessary.

You can see in the after screen shot  how the "areaTitle" is attached at the end and it might not be ideal always (in surveillance project we use this title to come back to a different page). I added a hideAreaInDocTitle variable for this ocasions

### Screenshoots

Before can be check in examples

Screen Shot 2022-05-25 at 19 04 26

After 

Screen Shot 2022-05-25 at 19 04 20

### DIFF
--- a/src/Pages/atoms/PageTitle.tsx
+++ b/src/Pages/atoms/PageTitle.tsx
@@ -60,13 +60,13 @@ interface IProps {
   areaTitle?: string
   areaHref?: string
   updateDocTitle?: boolean
+  hideAreaInDocTitle? : boolean
 }
 
-const PageTitle : React.FC<IProps> = ({title, icon, areaTitle, areaHref, updateDocTitle = true}) => {
+const PageTitle : React.FC<IProps> = ({title, icon, areaTitle, areaHref, updateDocTitle = true, hideAreaInDocTitle=false }) => {
   // Set <title> attribute automagically.
 
-  useTitle(title, areaTitle || '', undefined, updateDocTitle);
-
+  useTitle(title, hideAreaInDocTitle ? undefined : areaTitle || '', undefined, updateDocTitle);
 
   return (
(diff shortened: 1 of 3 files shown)


atomworks (changes requested):
This looks like it removes `baseTitle` completely which breaks the intended pattern of titles. I can see why the current implementation had some issues with trying to use the `document.title` however this shouldn't be dropped entirely and could do with some kind of consideration.

The usage of `document.title` in this way was to create an easy way to get the application name for use in our structured titles. To remove this is somewhat of a breaking change (although not critically) so we might want to find a way to retain this capability while also providing a better way for developers to formally set this title once globally.

The page title structure should remain as follows:
`App Name` - Site's top level / home. For example, _Example App_
`Section Name | App Name` - First nested level, for example _Settings_
`Page Name | Section Name | App Name` - Second nested level, for example _General_

So a page title using those examples would be:
`General | Settings | Example App`

Just let me know if you have any questions.

isacoder:
Hi @atomworks
Thanks for the feedback, I have updated the code to keep "App Name" / base-title
Please review

Screen Shot 2022-06-02 at 18 07 18

atomworks (approved):

PR #296, opened 2022-04-15
[Menus] preserve keep-open/auto-hide state in menu

fti-pvishwakarma:
### Description
In Support site one issue was reported to menu's Keep Open and Auto Hide state. In this requirement it was expected that Keep Open / Auto-Hide state must be preserve while we navigate to another page or reload the page.

Ref: https://www.bugherd.com/projects/281466/tasks/44

Screenshot 2022-04-15 183719

### Reviewing/Testing steps

Follow the steps given at :
https://github.com/future-standard/scorer-ui-kit/
When project runs, Open http://localhost:3000/scorer-ui-kit#/globalUI in browser you will see

image

Click on KeepOpen  bottom and refresh menu.

PR #315, opened 2022-05-27
Fix bug in preserve keep-open/auto-hide state in menu

fti-pvishwakarma:
### Description

 This PR for fixing a Bug in the previous PR  [preserve keep-open/auto-hide state in menu](https://github.com/future-standard/scorer-ui-kit/pull/296) 

Bug found if pass props canAlwaysPin = true then if screen resolution width more than 1400px and refresh the page then it shows Auto-Hide and auto close menu, it's not expected behavior.

### Reviewing/Testing steps
Please follow the previous PR instructions. [LINK](https://github.com/future-standard/scorer-ui-kit/pull/296)

### DIFF
--- a/src/hooks/useMenu.ts
+++ b/src/hooks/useMenu.ts
@@ -43,7 +43,7 @@ const menuReducer = (state: IMenuState, action: IMenuActions) => {
 
       let isMenuOpen = (openValueStorage === 'true') || (openValueStorage === null && (!!action.data.defaultMenuOpen));
       let isMenuPinned = (openValueStorage === 'true') && !!action.data.canAlwaysPin;
-      let canPin = (action.data.desktopSize === 'xlarge') || !!action.data.canAlwaysPin;
+      const canPin = (action.data.desktopSize === 'xlarge') || !!action.data.canAlwaysPin;
 
       if (action.data.desktopSize === 'xxlarge' && action.data.canAlwaysPin === false) {
         isMenuOpen = true;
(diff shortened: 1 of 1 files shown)


src/hooks/useMenu.ts
```diff
@@ -37,23 +37,17 @@ const menuReducer = (state: IMenuState, action: IMenuActions) => {
...
-      let isMenuPinned = openMenu === 'true';
-      let canPin = false;
+      const openValueStorage = localStorage.getItem(window.location.hostname + '_isMenuOpen');
 
-      if (action.data.canAlwaysPin || (action.data.defaultMenuOpen && action.data.desktopSize === 'xlarge')) {
-        canPin = true;
-      }
+      let isMenuOpen = (openValueStorage === 'true') || (openValueStorage === null && (!!action.data.defaultMenuOpen));
+      let isMenuPinned = (openValueStorage === 'true') && !!action.data.canAlwaysPin;
+      let canPin = (action.data.desktopSize === 'xlarge') || !!action.data.canAlwaysPin;
```
isacoder, line 46: canPin is not edited further bellow, please change this from let variable to a const.

isacoder (changes requested):
I tested in comparison with current deploy and the problem does not persists :)

src/hooks/useMenu.ts
fti-pvishwakarma: Done

isacoder (approved):
Thanks for the update all works as expected

PR #325, opened 2022-06-09
New notification does not show without closing current notification

isacoder:
### Description
New notification does not show without closing current notification

https://www.bugherd.com/projects/288221/tasks/60
```
If an error occurs and the Save button is pressed before clicking on the "x" in the error pop-up, the pop-up will not appear again, despite the error condition.

In fact, the edits are not updated because of the error, but the user is not informed that an error has occurred.
```

### Implementation

The error I found was that the props when equal are not, refreshed.
I have added an id base in the date to have a prop that can trigger this "refresh" of values and animation.

Also added a clearNotification function since I notice that sometimes when "saving" we want to remove the notifications to have it ready for the new one relative to the view.

### Screenshoot

Video
https://drive.google.com/file/d/1IHIxSaiB_evvkb3NDtmwzzKSHplgYo-d/view?usp=sharing

Screen Shot 2022-06-09 at 20 43 24


src/context/NotificationContext.tsx
```diff
@@ -1,12 +1,19 @@
-import React, { useState, useCallback } from 'react';
+import React, { useState, useCallback, useMemo } from 'react';
 import Notification, { INotificationProps } from '../Alerts/atom/Notification';
 
+
+const uniqueID = (): string => {
+  return `${Date.now()}`;
```
JoshLipps, line 6: when many notices happen in a similar moment Date.now() can have collisions.  
image
 See here:
 
image


Alternatively we could just add some random number to date.now to get less collisions.  

https://developer.mozilla.org/en-US/docs/Web/API/Performance/now

src/Alerts/atom/Notification.tsx
```diff
@@ -131,7 +133,7 @@ const Notification : React.FC<INotificationProps> = ({type ='info', message, isP
     setDismiss(false);
     setSlideUp(false);
     setTextClicked(false);
-  },[message]);
```
JoshLipps, line 134: @atomworks I think here is where the ID is used.

src/context/NotificationContext.tsx
JoshLipps: Maybe this is just over engineering tho.

isacoder: I did gave a though on this but at the end I thought that maybe having the date will be useful in the future.

This article mentioned that if your "Trigger is user click" the date should be enough but also suggested some interesting ideas

https://dev.to/rahmanfadhil/how-to-generate-unique-id-in-javascript-1b13

I'll use this suggestion


```
const uniqueID = () =>
  String(
    Date.now().toString(32) +
      Math.random().toString(16)
  ).replace(/\./g, '');
  ```

isacoder:
> Looks good but a bit confused about the introduction of a unique ID to the messages. It looks like it's included in a few dependencies but isn't actually being used. I think I'm just missing the use case though.
> 
> What's it being used for?

Well it's a React thing. It was hard to catch for me... but it seems that since I'm using the same Notification Component and not generating a new one to display the alert and the properties were the same the State would not change and the new Alert will not "re-render" and trigger the animation of entering and leaving. 

Since the "on leaving" end of animation triggers the dismiss of the current one and moves to the next one in queue, the next alerts will not show, so I needed a unique prop.

```js
  <NotificationContext.Provider value={contextValue}>
      {activeNotification
        ? <Notification {...activeNotification} />
        : null}
      {children}
    </NotificationContext.Provider>
  );
  ```

src/context/NotificationContext.tsx
JoshLipps: This works, the issue is if 2 notifications are triggered by a user click at the same time.

JoshLipps (commented):
LGTM

atomworks (approved):

PR #327, opened 2022-06-13
[AlertBar] Add hideCloseIcon prop to AlertBar Component

### LINKED ISSUE #326: [AlertBar] Add hideCloseIcon prop to AlertBar Component
### 

In AlertBar we have found that there is no option(prop) available to hide or change that "close-icon" and a design is requiring it.

isacoder:
### Description

closes #326
- #326

### Implementation
Adding the hiding prop will not affect behaviour since dismiss button was just not rendering the bar but actually this component was "cleared" in the developers component by their own state.

### Screenshoot

**Common usage**
Screen Shot 2022-06-13 at 15 41 03

### hideCloseButton prop activated

Screen Shot 2022-06-13 at 15 40 54

### DIFF
--- a/src/Alerts/atom/AlertBar.tsx
+++ b/src/Alerts/atom/AlertBar.tsx
@@ -60,9 +60,9 @@ export interface IAlert {
 interface Props {
   type?: AlertType
   message?: string
-  hideCloseIcon?: boolean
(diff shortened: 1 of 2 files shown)


JoshLipps (approved):
:shipit:

PR #351, opened 2022-06-20
Feature/LineUI handle double arrow

fti-pvishwakarma:
### Description

This PR has enhancement in Line UI for the Direction Arrow feature (Feature required for Traffic Counter Project)
- Add Up and Down direction arrow on the same line
- Option to control the show and hide point handle of a specific line
- Fix the alignment issue for the line label to show it on center

Zeplin design:https://zpl.io/r1y9qOD
Camera - Configuration

The implementation is done for **LineUI** component only. Once the changes are approved then we will make similar changes for **LineUIVideo** and **LineUIRTC** components.

 ### Reviewing/Testing steps
Follow the steps given:
https://github.com/future-standard/scorer-ui-kit/tree/feature/line-ui-arrow-icon#readme

When the project runs, Open http://localhost:3000/scorer-ui-kit#/line in the browser you will see

Hide PointHandler Add  "showPointHandle: false" in State Data
Show Both Directions Mark  Add "showBothDirectionMark: true" in  State Data

localhost_3000_scorer-ui-kit(1440)

### DIFF
--- a/src/LineUI/LineSet.tsx
+++ b/src/LineUI/LineSet.tsx
@@ -229,14 +229,14 @@ const LineSet : React.FC<ILineSetProps> = ({ getCTM, boundaries, unit, size, lin
       styling={styling}
       lineMoveCallback={lineDragUpdate}
       lineMoveStartCallback={lineDragStart}
-      doubleDirectionMark={!!lineSetData.showDoubleDirectionMark}
+      bothDirectionMark={!!lineSetData.showBothDirectionMark}
     />
   );});
 
   return (
     <g>
       {lines}
-      {!lineSetData.hidePointHandle && handles}
+      {(lineSetData.showPointHandle ?? true) && handles}
       {points}
       <AreaLabel lineSetData={lineSetData} unit={unit} />
     </g>
--- a/src/LineUI/LineSet.tsx
+++ b/src/LineUI/LineSet.tsx
@@ -181,25 +181,26 @@ const LineSet : React.FC<ILineSetProps> = ({ getCTM, boundaries, unit, size, lin
     }
   }, [lineSetData, updateHandleAngles, handleUsesAngles]);
 
-  const handles = lineSetData.points.map(({x,y}, index) =>
-    <HandleUnit
-      key={index+lineSetId}
-      lineSetId={lineSetId}
-      rotate={lineSetData.rotate}
-      Icon={lineSetData.icon}
-      index={index}
-      unit={unit}
-      size={size}
-      useAngles={handleUsesAngles}
-      angle={handleAngles[index]}
-      x={x}
-      y={y}
-      moveEndCB={onLineMoveEnd}
-      moveCallback={handleMoveCallback}
-      options={options}
-      styling={lineSetData.styling}
-      readOnlyHandle={lineSetData.readOnly}
-    />
+  const handles = (lineSetData?.showPointHandle === undefined || lineSetData?.showPointHandle) &&
+    lineSetData.points.map(({ x, y }, index) =>
+      <HandleUnit
+        key={index+lineSetId}
+        lineSetId={lineSetId}
+        rotate={lineSetData.rotate}
+        Icon={lineSetData.icon}
+        index={index}
+        unit={unit}
+        size={size}
+        useAngles={handleUsesAngles}
+        angle={handleAngles[index]}
+        x={x}
+        y={y}
+        moveEndCB={onLineMoveEnd}
+        moveCallback={handleMoveCallback}
+        options={options}
+        styling={lineSetData.styling}
+        readOnlyHandle={lineSetData.readOnly}
+      />
   );
 
   const points = options.showPoint && (lineSetData.points.map(({x,y}, index) => <Point styling={lineSetData.styling||'primary'} key={index} r={unit} cx={x} cy={y} />));
@@ -237,7 +238,7 @@ const LineSet : React.FC<ILineSetProps> = ({ getCTM, boundaries, unit, size, lin
   return (
     <g>
       {lines}
-      {(lineSetData.showPointHandle ?? true) && handles}
+      {handles}
       {points}
       <AreaLabel lineSetData={lineSetData} unit={unit} />
     </g>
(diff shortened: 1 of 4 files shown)


src/LineUI/LineSet.tsx
```diff
@@ -229,13 +229,15 @@ const LineSet : React.FC<ILineSetProps> = ({ getCTM, boundaries, unit, size, lin
       styling={styling}
       lineMoveCallback={lineDragUpdate}
       lineMoveStartCallback={lineDragStart}
+      bothDirectionMark={!!lineSetData.showSmallDirectionMark}
```
fti-bcarpenter, line 232: It seems !! is wrong and also implementation of getting value of bothDirectionMark is wrong.
 
I think when we have **showDirectionMark** and **showSmallDirectionMark** both true then this **bothDirectionMark** should be true otherwise it should be false.

src/LineUI/LineUnit.tsx
```diff
@@ -93,19 +93,21 @@ interface ILineUnitProps {
   moveEndCB?: () => void;
   label?: string;
   styling?: string;
+  bothDirectionMark: boolean;
+  hideMoveHandle?: boolean;
```
fti-bcarpenter, line 97: Please change `hideMoveHandle` to `showMoveHandle` 
Also make sure you update the implementation where this variable is used.

src/LineUI/LineUnit.tsx
```diff
@@ -93,19 +93,21 @@ interface ILineUnitProps {
...
-  const { x1, y1, x2, y2, unit, lineMoveCallback, lineMoveStartCallback, options, lineSetId, label, styling='primary', moveEndCB=()=>{}} = props;
+  const { x1, y1, x2, y2, unit, lineMoveCallback, lineMoveStartCallback, options, lineSetId, label, styling = 'primary', moveEndCB = () => { }, bothDirectionMark, hideMoveHandle = true } = props;
   const { handleFinderActive, revealSetIndex, showMoveHandle, setIndexOffset, showDirectionMark} = options;
 
   const a = x1 - x2;
   const b = y1 - y2;
   const distance = Math.sqrt( a*a + b*b );
   //this distance 60 doesn't work now...
-  const hideGrabHandle = showMoveHandle === false || (showMoveHandle !== true && distance < 60);
+  const hideGrabHandle = hideMoveHandle ? (showMoveHandle === false || (showMoveHandle !== true && distance < 60)) : !hideMoveHandle;
```
fti-bcarpenter, line 110: Change hideMoveHandle variable as per previous comment

src/LineUI/LineUnit.tsx
```diff
@@ -93,19 +93,21 @@ interface ILineUnitProps {
...
   styling?: string;
+  bothDirectionMark: boolean;
+  hideMoveHandle?: boolean;
 }
 
 
 
 const LineUnit : React.FC<ILineUnitProps> = (props) => {
-  const { x1, y1, x2, y2, unit, lineMoveCallback, lineMoveStartCallback, options, lineSetId, label, styling='primary', moveEndCB=()=>{}} = props;
+  const { x1, y1, x2, y2, unit, lineMoveCallback, lineMoveStartCallback, options, lineSetId, label, styling = 'primary', moveEndCB = () => { }, bothDirectionMark, hideMoveHandle = true } = props;
```
fti-bcarpenter, line 103: Keep default value of bothDirectionMark = false

fti-bcarpenter (changes requested):

src/LineUI/LineSet.tsx
fti-pvishwakarma: Done

src/LineUI/LineUnit.tsx
fti-pvishwakarma: showMoveHandle is already available in the global option.

src/LineUI/LineUnit.tsx
fti-pvishwakarma: showMoveHandle is already available in the global option.

src/LineUI/LineUnit.tsx
fti-pvishwakarma: Done

src/LineUI/LineUnit.tsx
fti-bcarpenter: You can give different name to it but hideMoveHandle is opposite of global prop showMoveHandle which is confusing here.  You can change hideMoveHandle to showLineMoveHandle or something similar.

src/LineUI/LineSet.tsx
fti-bcarpenter: Its not done. Check comments again and fix it.

src/LineUI/LineUnit.tsx
fti-bcarpenter: You can give different name to it but hideMoveHandle is opposite of global prop showMoveHandle which is confusing here.  You can change hideMoveHandle to showLineMoveHandle or something similar.

fti-bcarpenter (changes requested):

src/LineUI/LineUnit.tsx
isacoder: I agree with @fti-bcarpenter  maybe it can be something like overrideShowMoveHandle

src/LineUI/LineUnit.tsx
isacoder: following up with the idea of the overrideShowMoveHandle  we can simplify this logic
// since the distance is not been accounted  now it can be removed if you prefer commenting the previous value as reference is fine
// showMoveHandle === false || (showMoveHandle !== true && distance < 60);

The logic I imagine that will be if either value is false it should be true to hide.

const hideGrabHandle = ( !showMoveHandle || !overrideShowMoveHandle)

Please consider and verify.

isacoder (changes requested):
Functionality wise seems to be the expected but implementation needs refinement.

src/LineUI/LineSet.tsx
```diff
@@ -229,13 +229,15 @@ const LineSet : React.FC<ILineSetProps> = ({ getCTM, boundaries, unit, size, lin
...
+      showSmallDirectionMark={lineSetData.showSmallDirectionMark}
+      overrideShowMoveHandle={lineSetData.showMoveHandle}
     />
   );});
 
   return (
     <g>
       {lines}
-      {handles}
+      {(lineSetData.showPointHandle ?? true) && handles}
```
isacoder, line 240: this is breaking storybook, 

please use || instead of ??

lineSetData.showPointHandle || true

Screen Shot 2022-07-07 at 21 55 08

isacoder (changes requested):
Just one small change, please

fti-bcarpenter (approved):

PR #353, opened 2022-06-23
[Filter Dropdown] Empty State - Implementation

### LINKED ISSUE #347: [Filter Dropdown] Empty State - Implementation
Should be an easy one!

The design is available here: https://zpl.io/NwGM0zN

isacoder:
### Description

- #347 

The design is available here: https://zpl.io/NwGM0zN

closes #347 

### Implementation

🙈  Noticed that the Search field and the empty results are 12px in the implementation and 14px in the design.

I don't remember why I re-wrote 12px on BasicInputSearch.
Do we leave it like is now or should I update this to 14px although we might find out the reason later..
or should I make this value variable/flexible for this case scenario?

### Screenshoot
Screen Shot 2022-06-23 at 17 48 35

### DIFF
--- a/src/Filters/molecules/FilterDropdown.tsx
+++ b/src/Filters/molecules/FilterDropdown.tsx
@@ -77,6 +77,18 @@ const SearchWrapper = styled.div`
   align-items: center;
 `;
 
+const EmptyResultText = styled.div`
+  display: block;
+  color: hsl(0, 0%, 56%);
+  margin-left: 12px;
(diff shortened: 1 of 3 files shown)


atomworks (approved):
👍

JoshLipps (approved):
LGTM. I am fine either way on the 12 vs 14 issue. I think 14 would be nice but I also worry about backwards compatibility.

PR #354, opened 2022-07-01
Actions Bar implementation with Wrapper switch

isacoder:
### Description

- #345

Design https://zpl.io/EmEDYqg

Selecting multiple rows to have an action.

### Implementation
I had to create in the story a component example to fix the modal situation since it needed to be wrap to be inside the provider. JSX is not very helpful in this case because the Component is shown as code to copy 😞 

### Screenshoot

Initial
Screen Shot 2022-07-01 at 21 28 21

Actions component
Screen Shot 2022-07-01 at 21 28 28

Modal
Screen Shot 2022-07-01 at 21 28 33

Many buttons (I copied the same for this screenshoot but code does not have it.
Screen Shot 2022-07-01 at 21 42 36


src/Misc/atoms/ActionsBar.tsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import styled from 'styled-components';
+import Button from '../../Form/atoms/Button';
+import ButtonWithIcon, { IButtonWithIcon } from '../../Form/atoms/ButtonWithIcon';
+
+const Container = styled.div``;
+const Title = styled.div`
+  font-family: ${({ theme }) => theme.fontFamily.ui};
+  color: hsl(207, 5%, 57%);
```
JoshLipps, line 9: Minor: Hard coded color?

src/Misc/atoms/ActionsBar.tsx
```diff
@@ -0,0 +1,97 @@
...
+const RightButtons = styled.div`
+  flex-shrink: 0;
+  align-self: flex-end;
+`;
+
+const SelectedResults = styled.div`
+  font-family: ${({ theme }) => theme.fontFamily.ui};
+  margin-top: 40px;
+  min-height: 19px;
+  color: hsl(207, 5%, 57%);
```
JoshLipps, line 36: Minor: Hard coded color?

src/Misc/atoms/ActionsBar.tsx
```diff
@@ -0,0 +1,97 @@
...
+  totalSelected?: number
+  totalAvailable?: number
+  finishCallback?: () => void
+}
+
+const ActionsBar: React.FC<IActionsBar> = ({
+  title = 'Actions:',
+  finishTextButton = 'Finish',
+  actionButtons = [],
+  selectedTemplate = 'Selected [SELECTED] of [TOTAL] Results',
```
JoshLipps, line 64: I wonder if this could actually be a template literal instead of using our own injection points

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates

JoshLipps (commented):
All in all this looks great. Made a couple comments. Template literal is just a Idea not a requirement.

atomworks (changes requested):
Just a couple of requests and a small code comment...

1. When selecting the button and changing between filter and action buttons, the table is jumping. Could we make sure that the height of the filters component and action buttons component is consistent to avoid this? For multi-line I think a jump will be natural but for single line cases it should be solid.
https://user-images.githubusercontent.com/808289/179995810-65c80a94-a08b-41c7-a626-daf9c046f401.mov

2. For the multi line button example in [this screenshot](https://user-images.githubusercontent.com/10409078/176897095-ad7e55a0-f70e-456e-b366-688b1353b4a6.png) the button is out of line with the rest. Can we move this to stay in line with the top line rather than the bottom line too?

src/Misc/atoms/ActionsBar.tsx
```diff
@@ -0,0 +1,97 @@
...
+`;
+
+const ButtonsWrapper = styled.div`
+  display: flex;
+  justify-content: space-between;
+`;
+
+const LeftButtons = styled.div`
+  button {
+    margin-right: 10px;
```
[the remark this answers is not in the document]
isacoder: Good catch!, I don't remember how that ended like that hehehe 😆

src/Misc/atoms/ActionsBar.tsx
isacoder: Well there was no value that matched the current css styles but this is a common value, perhaps @atomworks  can recommend the way to handle it. 

Screen Shot 2022-07-21 at 17 28 17

In any case I did verify it looks good in light and dark mode

Screen Shot 2022-07-21 at 17 31 05

Screen Shot 2022-07-21 at 17 31 52

src/Misc/atoms/ActionsBar.tsx
isacoder: same answer as above :)

src/Misc/atoms/ActionsBar.tsx
isacoder: Hi!

I might not understand correctly your question, I did read the link but couldn't understand what I'm not doing right. I think I'm doing template literals, although it might not be obvious

Screen Shot 2022-07-21 at 19 43 20

This selected value is the default template, the user can send any kind of sentence in any language while maintaining the [SELECTED] and [TOTAL] variables

The Developer can send something like

selectedTemplate = "Seleccionando [SELECTED] libros de [TOTAL]"

"Seleccionando 5 libros de 10"

The story has the example of English and Japanese


Screen Shot 2022-07-21 at 19 37 02

Screen Shot 2022-07-21 at 19 36 54

Sorry  I couldn't see what I need to improve, please provide and example to confirm this part or let's chat on slack to review it.

JoshLipps: I understand WHY there are interpolation points, I am just suggesting not reinventing a feature of the language. the prop passed in could be the template sting instead of a string with custom made injection points.

JoshLipps: It seems harder than expected to use template literals as variables. 
This solution's seems fine.

JoshLipps (approved):
I'm OK with this.

atomworks (approved):

PR #356, opened 2022-07-07
Feature/camera panel

### LINKED ISSUE #346: [Camera Panels] Design

isacoder:
### Description

closes #346
- #346

Design
https://zpl.io/llyq7Rv

### Implementation

I have reused the MediaBox since I was not sure this camera was for video or image but that can be managed by the user.

I tried to make it flexible, I think there is an area of opportunity to add the option that PanelMetaData could be replace by a customComponent, however let think this in the iterations of this feature.

I reused the values of IFeebackColor because it seems to be the similar colors but I could update to have any color received. 

I added epsilon for all titles since space seems limited

### Screenshoot

Screen Shot 2022-07-07 at 18 27 30

Screen Shot 2022-07-07 at 18 31 29

### DIFF
--- a/example/package-lock.json
+++ b/example/package-lock.json
@@ -27973,28 +27973,6 @@
           "resolved": "https://registry.npmjs.org/tapable/-/tapable-1.1.3.tgz",
           "integrity": "sha512-4WK/bYZmj8xLr+HUCODHGF1ZFzsYffasLUgEiMBY4fgtltdO6B4WJtlSbPaDTLpYTcGVwM2qLnFTICEcNxs3kA=="
         },
-        "terser": {
-          "version": "4.8.0",
-          "resolved": "https://registry.npmjs.org/terser/-/terser-4.8.0.tgz",
-          "integrity": "sha512-EAPipTNeWsb/3wLPeup1tVPaXfIaU68xMnVdPafIL1TV05OhASArYyIfFvnvJCNrR2NIOvDVNNTFRa+Re2MWyw==",
-          "requires": {
-            "commander": "^2.20.0",
-            "source-map": "~0.6.1",
-            "source-map-support": "~0.5.12"
-          },
-          "dependencies": {
-            "commander": {
-              "version": "2.20.3",
-              "resolved": "https://registry.npmjs.org/commander/-/commander-2.20.3.tgz",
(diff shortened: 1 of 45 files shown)


JoshLipps (approved):
This looks good, I wonder if how it will merge with the recent retry update to MediaBox, but i think this will be ok.

isacoder:
> This looks good, I wonder if how it will merge with the recent retry update to MediaBox, but i think this will be ok.
> 
> I also wonder how soon until someone askes for one that support WebRTC, but this can be a followup.


That change wasn't in the main, now it is 🙈 .. It breaks this branch, I will fix it.

isacoder:
Long story short, the only breaking conflict was the min-width, min-height that was added, so I just created props to send 

Now Both the Camera panels and the Media modal can have continue to have custom size.

Screen Shot 2022-07-28 at 20 36 47


Updated the text and close icon to mono since it was not very legible like Croptool required too, will be helpful for Dark Mode too.
Screen Shot 2022-07-28 at 20 36 29

atomworks (changes requested):
Sorry for the late feedback on this. The code looks good but there is an implementation thing I'd like to see so we have enough flexbility for a lot of the different variations.

The text area of the panel has many variations for the different projects. For flexibility I would like to see the `PanelMetaData` as a default component but if we can send a component via prop to replace this with something else, that would maximise the potential for use as that area could then be fully customised.

Also `PanelMetaData` I would like to make the location meta data (above the title) optional, and if so we vertically center the title. Also if the right hand part of this can be optionall too, then we just hide that if those props aren't provided.

Just drop me a message on Slack if you need any clarification.

isacoder:
### Updated MediaStream and PanelMetaData considering more flexibility hiding items and considering alignment.

**Custom Component**
Screen Shot 2022-07-29 at 16 54 28

**Sending empty strings but left subtitle**
Screen Shot 2022-07-29 at 15 54 38

**Without right items**

Screen Shot 2022-07-29 at 15 19 23

**Other options**
Screen Shot 2022-07-29 at 14 58 16


**MediaStream Notice and Icon flexibility**
Screen Shot 2022-07-29 at 17 57 20

atomworks (approved):

PR #362, opened 2022-07-22
Add icon prop to Notification Component

fti-skumar:
### Description

 - In `Notification` there is no option(prop) available to change that notification type "icon" on left and there is an requirement in the design.

### Implementation

- Added a prop: `icon` which can be used to change the type-icon.
- _Note: If `icon` prop is not specified then the component will follow default behavior that is the icon shown will depend on the type of notification._

### Screenshots

- **When icon is not specified**

image

- **When icon is specified**

image

### DIFF
--- a/src/Alerts/atom/Notification.tsx
+++ b/src/Alerts/atom/Notification.tsx
@@ -116,13 +116,14 @@ const MainMessage = styled.div`
 export type INotificationProps = {
   type: AlertType;
   message: string;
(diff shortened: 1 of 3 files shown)


isacoder (approved):
This for this update, it works as expected.
I appreciate that the story shows the empty icon variable and the others as select.

fti-skumar:
Thankyou @isacoder for the review.

PR #379, opened 2022-09-13
Feature/Line UI enhancements

fti-skumar:
### Description

This PR has updates for `LineUI` to provide support for customer requests. 
- Lines should be selectable on clicking them.
- Provide fix for overlapping of Line-label and `small-direction-arrow`
- Links of Bugherd issues
  - [Selectable Lines](https://www.bugherd.com/projects/289019/tasks/159)
  - [Overlapping issue](https://www.bugherd.com/projects/289019/tasks/157)

- For Selectable lines, Function binding is required with an parameter to `onLineClick` prop of LineUI which will be called when lines got clicked and `lineSetID` will be provided. We can use this **ID** to toggle `readOnly` and `showHandle` option of Lines
- For Overlapping issue, I kept previous implementation of line-label same, fix will only occur when we have `small-direction-arrow`.

 ### Reviewing/Testing steps

- To test it just run the example pages and goto line page. And click on any line and that line will be selected.
- To test overlapping issue with `small-direction-arrow` just rename the line will large text from the options in left menu and rotate the line in all direction and check if label and `small-direction-arrow` overlap or not.

### DIFF
--- a/example/src/pages/LinePage.tsx
+++ b/example/src/pages/LinePage.tsx
@@ -23,6 +23,7 @@ const Line: React.FC<{}> = () => {
   const [options, setOptions] = useState<LineUIOptions>({
     showSetIndex: true,
     pointIndexOffset: 1,
+    showPointHandle: true,
     showPointLabel: true,
     setIndexOffset: 1,
     showMoveHandle: true,
@@ -73,6 +74,7 @@ const Line: React.FC<{}> = () => {
               y: 389
             }
         ],
+        showPointHandle: true,
         showSmallDirectionMark: true,
         readOnly: false,
         styling: 'primary'
@@ -133,14 +135,20 @@ const Line: React.FC<{}> = () => {
   }, [state]);
 
   const selectLine = useCallback((lineId: number) => {
-    state.forEach((_, index) => {
-      dispatch({
-        type: 'UPDATE_SET_OPTIONS',
-        index,
-        data: {
-          showPointHandle: lineId === index,
-        }
-      });
+    const deselectLineIndex = state.findIndex((item) => item.showPointHandle);
+    dispatch({
+      type: 'UPDATE_SET_OPTIONS',
+      index: deselectLineIndex,
+      data: {
+        showPointHandle: false,
+      }
+    });
+    dispatch({
+      type: 'UPDATE_SET_OPTIONS',
+      index: lineId,
+      data: {
+        showPointHandle: true,
+      }
     });
   }, [state]);
 
--- a/example/src/pages/LinePage.tsx
+++ b/example/src/pages/LinePage.tsx
@@ -141,6 +141,7 @@ const Line: React.FC<{}> = () => {
       index: deselectLineIndex,
       data: {
         showPointHandle: false,
+        showMoveHandle: false
       }
     });
     dispatch({
@@ -148,6 +149,7 @@ const Line: React.FC<{}> = () => {
       index: lineId,
       data: {
         showPointHandle: true,
+        showMoveHandle: true
       }
     });
   }, [state]);
--- a/example/src/pages/LinePage.tsx
+++ b/example/src/pages/LinePage.tsx
@@ -132,6 +132,18 @@ const Line: React.FC<{}> = () => {
     });
   }, [state]);
 
+  const selectLine = useCallback((lineId: number) => {
+    state.forEach((_, index) => {
+      dispatch({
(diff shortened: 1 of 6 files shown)


example/src/pages/LinePage.tsx
```diff
@@ -132,6 +132,18 @@ const Line: React.FC<{}> = () => {
     });
   }, [state]);
 
+  const selectLine = useCallback((lineId: number) => {
+    state.forEach((_, index) => {
```
JoshLipps, line 136: Minor:
```suggestion
    state.some((_, index) => {
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/LineUI/LineUnit.tsx
```diff
@@ -88,6 +87,7 @@ interface ILineUnitProps {
   x2: number,
   y2: number,
   unit: number,
+  lineClickCallback: (lineSetId: number) => void,
```
JoshLipps, line 90:
```suggestion
  lineClickCallback?: (lineSetId: number) => void,
```

JoshLipps (approved):
Minor code changes otherwise seems fine.

example/src/pages/LinePage.tsx
fti-skumar: @JoshLipps I think we should use `forEach` as our purpose is to select the line which was clicked and deselect all other lines in the state. But with `some` we can only get Boolean value depending on the condition and the callback function will stop as soon as the condition returns `true`. And our purpose will not be fulfilled as some of the lines will not be deselected.

JoshLipps: ahh i see

fti-skumar: yes we just need to dispatch for only two lines one for selection and other for deselection, ok let me think how we implement this.

fti-skumar: @JoshLipps I had updated the code please review it again.

JoshLipps: If thats the case can we prevent the unnecessary dispatches?

fti-skumar: Yes I did, in current code, it only dispatch two times.

JoshLipps (approved):
LGTM

atomworks (changes requested):
At a quick glance these code changes look OK but I wanted to confirm the functionality in the browser before approval. It seems there is a Typescript related error being reported when I try to compile the code however so I'm unable to test it. 

I'm not sure how you would have compiled it to test it without seeing this but if you can go back and fix this issue first, then re-request a review and I will take a look at it when working. Thank you.

The error:

```
Error: /Users/leonard/Documents/Projects/scorer-ui-kit/src/LineUI/LineUIRTC.tsx(195,16): semantic error TS2741: Property 'onLineClick' is missing in type '{ key: number; onLineMoveEnd: () => void; lineSetId: number; lineData: IPointSet; getCTM: () => DOMMatrix | null; boundaries: IBoundary; unit: number; size: number; options: { ...; }; }' but required in type 'ILineSetProps'.
```

fti-skumar:
Thanks @atomworks I'll fix it.

fti-skumar:
@atomworks I had fixed this issue, please review it again.

example/src/pages/LinePage.tsx
```diff
@@ -132,6 +134,24 @@ const Line: React.FC<{}> = () => {
...
+        showPointHandle: false,
+      }
+    });
+    dispatch({
+      type: 'UPDATE_SET_OPTIONS',
+      index: lineId,
+      data: {
+        showPointHandle: true,
+      }
+    });
```
atomworks, line 152: Just one small update request... in the example, when a line is considered to not be selected, let's hide the handle too as the usability is a little confusing there (as you can drag the line but not reselect it with that part, so this proposal is easiest). 

```suggestion
    dispatch({
      type: 'UPDATE_SET_OPTIONS',
      index: deselectLineIndex,
      data: {
        showPointHandle: false,
        showMoveHandle: false
      }
    });
    dispatch({
      type: 'UPDATE_SET_OPTIONS',
      index: lineId,
      data: {
        showPointHandle: true,
        showMoveHandle: true
      }
    });
```

atomworks (changes requested):
Just one small request but otherwise I think we're good!

atomworks (approved):

PR #380, opened 2022-09-23
Fix/Table thumbnail modal min-height issue

fti-skumar:
### Description

- In Tables, modal for row thumbnail has `min-height: 300` which is more than a QVGA(320 x 240) image due to which an extra space were seen in the bottom. This issue was reported by customer in below bugherd link.
- [Modal Extra space issue](https://www.bugherd.com/projects/289019/tasks/196) please check its attached thumbnail snapshot.

### Considerations and Implementation
- I just reduced the `min-height` to `240px`.
- To verify and implement this, I used a QVGA image.

### Screenshots

- **White space in height of QVGA image popup**

spaceIssue

- **Fix**

spaceFix

### DIFF
--- a/src/hooks/useMediaModal.tsx
+++ b/src/hooks/useMediaModal.tsx
@@ -62,7 +62,7 @@ export const useMediaModal = () => {
       dismissCallback,
       retryLoading = false,
       retryLimit=5,
-      minHeight='240px',
+      minHeight='300px',
       minWidth='300px',
     } = mediaModal;
 
--- a/src/hooks/useMediaModal.tsx
+++ b/src/hooks/useMediaModal.tsx
@@ -62,7 +62,7 @@ export const useMediaModal = () => {
       dismissCallback,
       retryLoading = false,
       retryLimit=5,
-      minHeight='300px',
+      minHeight='240px',
       minWidth='300px',
     } = mediaModal;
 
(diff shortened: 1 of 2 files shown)


fti-bcarpenter (commented):
LGTM

src/hooks/useMediaModal.tsx
```diff
@@ -62,7 +62,7 @@ export const useMediaModal = () => {
       dismissCallback,
       retryLoading = false,
       retryLimit=5,
-      minHeight='300px',
+      minHeight='240px',
```
JoshLipps, line 65: While this seems OK, I wonder why we are just changing the default, seems like it only fixes this one case.  Also I wonder if actually just fixing the Thumbnail Modal is a more appropriate fix. See here:
https://github.com/future-standard/scorer-ui-kit/blob/29bcc63886fe6f08ca8850a0164cd97bc4e85ce9/src/Tables/atoms/TableRowThumbnail.tsx#L107

atomworks (commented):
As Josh mentioned I think this fix is better handled elsewhere so I'll just submit this review as a comment, so please re-request a review when needed.

If we need to potentially use this as a short term fix, we could do that but I'd prefer then to follow up quickly with the correct fix. If that's needed, please get in touch via Slack to discuss things here.

fti-skumar:
Sure thanks @JoshLipps , @atomworks , let me check this again.

src/hooks/useMediaModal.tsx
fti-skumar: @JoshLipps I had handled the issue from props as you suggested. And I reverted the change in `useMediaModal`. Please review it again.

JoshLipps (approved):

PR #389, opened 2022-12-07
WIP:Feature/enhancement typetable

fti-pfarne:
### Description
- This PR is enhancement in Type Table .

- When we click on thumbnail it will show a modal. but we can't change close button text of model. To overcome this issue I have added prop `copyText` . From help of this prop we can add translated text during other languages also and pass text with the help of `copyText` prop.

- Added a prop copyText to add text.

### Screenshot

### DIFF
--- a/src/Tables/organisms/TypeTable.tsx
+++ b/src/Tables/organisms/TypeTable.tsx
@@ -72,12 +72,14 @@ interface IProps {
   selectCallback?: (checked: boolean, id?: string | number) => void
   toggleAllCallback?: (checked: boolean) => void
   sortCallback?: (ascending: boolean, columnId: string) => void
+  closeText?:string
 }
 
 const TypeTable: React.FC<IProps> = ({
   columnConfig,
   selectable,
   rows = [],
+  closeText,
   hasStatus = false,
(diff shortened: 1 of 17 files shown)


src/Tables/organisms/TypeTable.tsx
```diff
@@ -156,7 +158,8 @@ const TypeTable: React.FC<IProps> = ({
                 columnConfig,
                 hasStatus,
                 hasThumbnail,
-                hasTypeIcon
+                hasTypeIcon,
+                closeText,
               }}
```
fti-bcarpenter, line 163: Comma after closeText is not needed.

Also put another snapshot by changing CLOSE into Japanese.

fti-bcarpenter (changes requested):

src/Tables/organisms/TypeTable.tsx
fti-pfarne: Done

fti-bcarpenter (approved):

atomworks (approved):
I wish we had a better way to localise this but I think that's another challenge. 

So, this looks good to me. Thank you.

PR #391, opened 2022-12-09
New Box Modal

fti-pfarne:
### Description
- This PR is for new Box Modal.

- This BoxModal design is same as the image modal which is shown on click on table thumbnail. This is a reusable component which we can use to show data/image in a modal.

-  I have develop a new Box Modal with same CSS and props as of our old UI kit modal.

- This modal work as a component in our code.

- Related Bugherd Requirement links
[Requirement](https://www.bugherd.com/projects/289019/tasks/254)

- Tested and confirmed this modal in TCE project on Camera Detail page as well as Configuration page.

- Created new component as `BoxModal.tsx `file.

### Screenshot

Screenshot from 2022-12-09 10-24-48


src/Modals/BoxModal.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const dismiss = useCallback(() => {
+    if(dismissCallback) {
+      dismissCallback();
+    }
+    onDismiss();
+  }, [onDismiss, dismissCallback]);
+
+  return (isOpen
+    ? ReactDom.createPortal(
+      <Container>s
```
fti-bcarpenter, line 114: s 
Is it a typo?

fti-bcarpenter (changes requested):

src/Modals/BoxModal.tsx
fti-pfarne: Done

fti-bcarpenter (approved):

atomworks (commented):
I'm not sure of the purpose of this new modal and why we're not using existing components for this. Also as there is no story created for this I'm wondering if this needs to be added to the UI kit or if it's something that should be added to your project directly.

Do you have more information regarding this?

Please re-request a review by the way once you respond, as I might miss it otherwise. You can also ping me on Slack. Thanks for you consideration.

fti-pfarne:
Hello @atomworks,

1. In old modal we are showing image but that image is not big as image on snapshot modal which will pop up in `TypeTable`. so to make image big same as `TypeTable` image modal I have done this new modal.

2. Yes I can add this directly in project, without using UI-Kit. We have bug in TCE for this, where I have to use Ui-Kit modal.

3. For old modal, It is difficult to pass frames continuously. This is a another reason to create this new modal.

4. To avoid new web socket connection, I have created this new modal.

fti-aagrawal:
Hi @atomworks 

As we discussed over call, main objective of this component is to solve an issue in TCE where we had a need of shared state between parent page and the Modal - for reusing websocket connection created in parent page.  So, it is more like a InlineModal or ProxyModal component, proxying modal like look and feel but actually drawn inline in the parent page itself so sharing of state becomes easier. 

I hope above background helps. 

As we discussed on call, to generalize this component for ui-kit, we probably need more considerations from you. So for now we will add this to TCE project directly which had the current need and at a suitable time later, we can consider again about bringing it to ui-kit for more general purpose use.

isacoder:
We have decided to close this PR, since the decision was not to add it to UI KIT for the time being

PR #392, opened 2023-01-03
Fix calendar component date

fti-pfarne:
### Description
- This PR is for calendar component changes.

- In current component it not possible to remove current hh:mm value and enter new value. To fix this issue I have added functionality to add values directly in input box with the help of keyboard numbers.

- I have added validation when we put wrong values in time input box.

- When we add wrong value in input box it shows red colour as border to the input box.

- I have added blur event, if in input we add `0` or empty string so when we click outside of input box it will show `00` value in input box.

- Related Bugherd Requirement links
[Bugherd Issue](https://www.bugherd.com/projects/216134/tasks/132)

- Tested this work in TCC project.

### Screenshot
Screenshot from 2023-01-09 10-36-04

Screenshot from 2023-01-03 18-36-23

Screenshot from 2023-01-03 18-36-34

### DIFF
--- a/src/Filters/atoms/DateTimeBlock.tsx
+++ b/src/Filters/atoms/DateTimeBlock.tsx
@@ -3,6 +3,7 @@ import styled, { css } from 'styled-components';
 import {endOfDay, format,isEqual,min,set } from 'date-fns';
 
 import Icon from '../../Icons/Icon';
+import { getFormatedTime } from '../../helpers';
 
 const Container = styled.div<{hide:boolean}>`
   display: flex;
@@ -118,10 +119,10 @@ const DateTimeBlock : React.FC<IProps> = ({
   setDateCallback = ()=>{},
 }) => {
 
-  const convertHours = (date.getHours()).toString();
-  const convertMinutes = date.getMinutes().toString();
-  const [displayHours, setDisplayHours] = useState<string>(convertHours === '0'?'00':convertHours);
-  const [displayMinutes, setDisplayMinutes] = useState<string>(convertMinutes === '0'?'00':convertMinutes);
+  const convertHours = (date?.getHours()).toString();
+  const convertMinutes = (date?.getMinutes()).toString();
+  const [displayHours, setDisplayHours] = useState<string>(getFormatedTime(convertHours) as string);
+  const [displayMinutes, setDisplayMinutes] = useState<string>(getFormatedTime(convertMinutes) as string);
 
   const setDateHours = useCallback(({target: {value}}: React.ChangeEvent<HTMLInputElement>) => {
     const hourRegex  = /^[0-1]{0,1}[0-9]{0,1}$|(^2[0-4])$/;
@@ -168,20 +169,14 @@ const DateTimeBlock : React.FC<IProps> = ({
     }
   },[date, allowAfterMidnight]);
 
-  const onBlurInputs = useCallback(() =>{
-    if(displayHours === '0' || displayHours === ''){
-      setDisplayHours('00');
+  const onBlurInputs = useCallback((blurValue: string, value) =>{
+    const convertedValue = getFormatedTime(blurValue);
+    if(value === 'Minutes'){
+      setDisplayMinutes(convertedValue as string);
+    }else {
+      setDisplayHours(convertedValue as string);
     }
-    if(displayMinutes ==='0' || displayMinutes ===''){
-      setDisplayMinutes('00');
-    }
-    if(displayMinutes.length === 1){
-      setDisplayMinutes('0' + displayMinutes);
-    }
-    if(displayHours.length === 1){
-      setDisplayHours('0' + displayHours);
-    }
-  }, [displayHours, displayMinutes]);
+  }, []);
 
   return (
     <Container hide={!hasDate && !hasTime}>
@@ -204,9 +199,9 @@ const DateTimeBlock : React.FC<IProps> = ({
             <Icon icon='Time' color='dimmed' size={14} weight='light' />
           </IconWrap>
           <InputWrap>
-            <InputValue onBlur={onBlurInputs} {...{allowManualTimeChange}} name='hours' type='number' value={displayHours} onChange={setDateHours} autoComplete='off' />
+            <InputValue onBlur={()=>onBlurInputs(displayHours, 'Hours')} {...{allowManualTimeChange}} name='hours' type='number' value={displayHours} onChange={setDateHours} autoComplete='off' />
             <TimeColon>:</TimeColon>
-            <InputValue onBlur={onBlurInputs} {...{allowManualTimeChange}} name='minutes' type='number' value={displayMinutes} onChange={setDateMinutes} autoComplete='off' />
+            <InputValue onBlur={()=>onBlurInputs(displayMinutes, 'Minutes')} {...{allowManualTimeChange}} name='minutes' type='number' value={displayMinutes} onChange={setDateMinutes} autoComplete='off' />
           </InputWrap>
         </Item>
       )}
--- a/src/Filters/atoms/DateTimeBlock.tsx
+++ b/src/Filters/atoms/DateTimeBlock.tsx
@@ -36,7 +36,7 @@ const IconWrap = styled.div`
   padding-top: 2px;
 `;
 
-const Input = styled.input<{ readOnly? : boolean }>`
+const Input = styled.input<{ readOnly? : boolean, checkTimeValidation?:boolean }>`
   ${({theme}) => css`
     font-family: ${theme.fontFamily.data};
   `}
@@ -45,19 +45,14 @@ const Input = styled.input<{ readOnly? : boolean }>`
 
   width: 100%;
   border: none;
-  border: transparent 1px solid;
+  border: ${({ checkTimeValidation }) => checkTimeValidation ? 'red 1px solid' : 'transparent 1px solid'};
   outline: none;
   flex: 1;
   justify-content: space-between;
   border-radius: 3px;
-
   &:focus, &:hover {
-    border-color: blue;
+    border-color: ${({ checkTimeValidation }) => checkTimeValidation ? 'red' : 'blue'};
   }
-
-  ${({readOnly}) => readOnly && css`
-    border-color: transparent;
-  `}
 `;
 
 const TimeColon = styled.div`
@@ -71,13 +66,8 @@ const InputWrap = styled.div`
   border-radius: 3px;
 
   &:focus-within {
-
     background: ${({theme}) => theme.colors.menu.passive};
     box-shadow: 0px 0px 0px 5px ${({theme}) => theme.colors.menu.passive};
-
-    ${Input} {
-      border-color: ${({theme}) => theme.colors.divider};
-    }
 
     ${TimeColon}{
       color: ${({theme}) => theme.colors.pureTop};
@@ -96,6 +86,7 @@ interface IProps {
   setDateCallback?: (date: Date) => void
   setTimeCallback?: (date: Date) => void
   allowAfterMidnight?: boolean,
+  checkTimeValidation?: boolean,
 }
 
 const DateTimeBlock : React.FC<IProps> = ({
@@ -103,6 +94,7 @@ const DateTimeBlock : React.FC<IProps> = ({
   title,
   hasDate,
   hasTime,
+  checkTimeValidation,
   date = new Date(),
   setDateCallback = ()=>{},
 }) => {
@@ -170,9 +162,9 @@ const DateTimeBlock : React.FC<IProps> = ({
             <Icon icon='Time' color='dimmed' size={14} weight='light' />
           </IconWrap>
           <InputWrap>
-            <Input name='hours' type='number' min='0' max={allowAfterMidnight ? 24: 23} value={displayHours} onChange={setDateHours} />
+            <Input {...{checkTimeValidation}} name='hours' type='number' min='0' max={allowAfterMidnight ? 24: 23} value={displayHours} onChange={setDateHours} />
             <TimeColon>:</TimeColon>
-            <Input name='minutes' type='number' min='0' max='59' value={displayMinutes} onChange={setDateMinutes} />
+            <Input {...{checkTimeValidation}} name='minutes' type='number' min='0' max='59' value={displayMinutes} onChange={setDateMinutes} />
           </InputWrap>
         </Item>
       )}
--- a/src/Filters/atoms/DateTimeBlock.tsx
+++ b/src/Filters/atoms/DateTimeBlock.tsx
@@ -180,6 +180,18 @@ const DateTimeBlock : React.FC<IProps> = ({
     }
   },[date, allowAfterMidnight]);
 
+  const onBlurInputs = useCallback(() =>{
+    if(displayHours === '0'){
+      setDisplayHours('00');
+    } else if(displayMinutes ==='0'){
+      setDisplayMinutes('00');
+    } else if(displayHours === ''){
+      setDisplayHours('00');
+    } else if(displayMinutes ===''){
+      setDisplayMinutes('00');
+    }
+  }, [displayHours, displayMinutes]);
+
   return (
     <Container hide={!hasDate && !hasTime}>
       <Label>{title}</Label>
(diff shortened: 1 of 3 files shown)


src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -142,13 +176,25 @@ const DateTimeBlock : React.FC<IProps> = ({
   useEffect(()=>{
     if(allowAfterMidnight && isEqual(date, endOfDay(date))){
       setDisplayHours('24');
-      setDisplayMinutes('00');
-    } else {
-      setDisplayMinutes(format(date, 'mm'));
-      setDisplayHours(format(date,'HH'));
+      setDisplayMinutes('0');
```
fti-skumar, line 179: I think this should be `00` .

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -142,13 +176,25 @@ const DateTimeBlock : React.FC<IProps> = ({
...
+    if(displayMinutes ==='0'){
+      setDisplayMinutes('00');
+    } 
+    if(displayHours === ''){
+      setDisplayHours('00');
+    }
+    if(displayMinutes ===''){
+      setDisplayMinutes('00');
+    }
+  }, [displayHours, displayMinutes]);
```
fti-skumar, line 196: You can use `||` operator here to combine these multiple if's.

src/Filters/molecules/DatePicker.tsx
```diff
@@ -359,22 +359,30 @@ const DatePicker: React.FC<IDatePicker> = ({
 
   const updateStartDate = useCallback((start: Date) => {
     const { end } = selectedRange ? selectedRange : TODAY_INTERVAL;
-    if (isAfter(add(start, { minutes: 1 }), end)) return;
+    if((isAfter(add(start, { minutes: 1 }), end))){
+      setCheckTimeValidation(true);
```
fti-skumar, line 363: Do we also need to return here if condition is true ? As previous version has return statement.

src/Filters/molecules/DatePicker.tsx
```diff
@@ -359,22 +359,30 @@ const DatePicker: React.FC<IDatePicker> = ({
...
+      setCheckTimeValidation(false);
+    }
     setSelectedRange({ start, end });
   }, [selectedRange]);
 
   const updateEndDate = useCallback((end: Date) => {
     const { start } = selectedRange ? selectedRange : TODAY_INTERVAL;
-    if (isAfter(add(start, { minutes: 1 }), end)) return;
+    if(isAfter(add(start, { minutes: 1 }), end)){
+      setCheckTimeValidation(true);
```
fti-skumar, line 373: If need a return then please add that here as well.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -126,6 +149,17 @@ const DateTimeBlock : React.FC<IProps> = ({
...
+    const isContainsSymbolN =  /^[0-9]+[0-9]*$/;
+    if(value === '000'|| value.length>2){
+      return;
+    } else if (Number(value)>59){
+      value = '59';
+    } else if (!isContainsSymbolN.test(value)) {
+      if(value !== ''){
+        return ;
+      }
+    }
```
fti-skumar, line 161: I think this whole validation checks can be optimized with a regular expression.
**RegEx**:  `^[0-5][0-9]$`
Please try this and find if it fails or not to satisfy all the conditions.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -96,22 +105,36 @@ interface IProps {
...
+    if(value === '000' || value.length>2){
+      return;
+    } else if (Number(value)>24){
+      value = '24';
+    } else if (!isContainsSymbolN.test(value)) {
+      if(value !== ''){
+        return ;
+      }
+    }
+    setDisplayHours(value);
```
fti-skumar, line 137: You can use RegEx here as well
**RegEx**:  `(^[0-1]{0,1}[0-9]$)|(^2[0-4]$)`

fti-skumar (commented):
PR looks good to me, just few small changes required.

src/Filters/molecules/DatePicker.tsx
fti-pfarne: Here we are showing red border for input box when this condition will TRUE. so no need to return.

src/Filters/molecules/DatePicker.tsx
fti-pfarne: Here we are showing red border for input box when this condition will TRUE. so no need to return.

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: Done

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: Done

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: Done

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: For initial I am showing `0` in every input. so I have use `0`. and I also use blur event. so when blur event runs it will show `00` in input.

fti-skumar: ok

src/Filters/molecules/DatePicker.tsx
fti-skumar: ok

src/Filters/molecules/DatePicker.tsx
fti-skumar: ok

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -96,22 +105,32 @@ interface IProps {
...
+  const [displayHours, setDisplayHours] = useState<string>(convertHours);
+  const [displayMinutes, setDisplayMinutes] = useState<string>(convertMinutes);
 
   const setDateHours = useCallback(({target: {value}}: React.ChangeEvent<HTMLInputElement>) => {
+    const isContainsSymbolN = /^[0-1]{0,1}[0-9]$|(^2[0-4])$/;
+    if(Number(value)>24){
+      value = '24';
+    } else if (!isContainsSymbolN.test(value) && value !== '') {
+      return;
+    }
```
fti-skumar, line 132: Please reconsider this checks, I think you only need RegEx check.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -126,6 +145,13 @@ const DateTimeBlock : React.FC<IProps> = ({
   }, [date, displayMinutes, setDateCallback]);
 
   const setDateMinutes = useCallback(({target: {value}}: React.ChangeEvent<HTMLInputElement>) => {
+    const isContainsSymbolN = /^[0-5]{0,1}[0-9]$/;
+    if(Number(value)>59){
+      value = '59';
+    } else if (!isContainsSymbolN.test(value) && value !== '') {
+      return;
+    }
```
fti-skumar, line 153: Please reconsider this checks, I think you only need RegEx check.

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: In regex we can't do this thing like..
if value is more that 24 like 55
it must add only 24 so I have added another `if`statement.
Please tell me your thoughts.

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: In regex we can't do this thing like..
if value is more that 59 like 88
it must add only 59 so I have added another if statement.
Please tell me your thoughts.

fti-skumar: I think regex is capable of excluding number greater than 59.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -96,22 +105,30 @@ interface IProps {
...
-
-  const [displayHours, setDisplayHours] = useState<string>(format(date, "mm"));
-  const [displayMinutes, setDisplayMinutes] = useState<string>(format(date,'HH'));
+  const convertHours = (date.getHours()).toString();
+  const convertMinutes = date.getMinutes().toString();
+  const [displayHours, setDisplayHours] = useState<string>(convertHours === '0'?'00':convertHours);
+  const [displayMinutes, setDisplayMinutes] = useState<string>(convertMinutes === '0'?'00':convertMinutes);
 
   const setDateHours = useCallback(({target: {value}}: React.ChangeEvent<HTMLInputElement>) => {
+    const isContainsSymbolN = /^[0-1]{0,1}[0-9]{0,1}$|(^2[0-4])$/;
```
fti-skumar, line 127: Please reconsider its name.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -126,6 +143,11 @@ const DateTimeBlock : React.FC<IProps> = ({
   }, [date, displayMinutes, setDateCallback]);
 
   const setDateMinutes = useCallback(({target: {value}}: React.ChangeEvent<HTMLInputElement>) => {
+    const isContainsSymbolN = /^[0-5]{0,1}[0-9]{0,1}$/;
```
fti-skumar, line 146: Please reconsider its name.

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: Done

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: Done

fti-skumar (approved):
Thanks, LGTM.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -96,22 +105,30 @@ interface IProps {
   setDateCallback?: (date: Date) => void
   setTimeCallback?: (date: Date) => void
   allowAfterMidnight?: boolean,
+  checkTimeValidation?: boolean,
```
fti-bcarpenter, line 108: checkTimeValidation ---> allowManualTimeChange

Please change this on all places

fti-bcarpenter (changes requested):

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: Done

fti-bcarpenter (approved):
Changes looks good.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -96,22 +105,30 @@ interface IProps {
...
   setDateCallback = ()=>{},
 }) => {
 
-
-  const [displayHours, setDisplayHours] = useState<string>(format(date, "mm"));
-  const [displayMinutes, setDisplayMinutes] = useState<string>(format(date,'HH'));
+  const convertHours = (date.getHours()).toString();
+  const convertMinutes = date.getMinutes().toString();
+  const [displayHours, setDisplayHours] = useState<string>(convertHours === '0'?'00':convertHours);
+  const [displayMinutes, setDisplayMinutes] = useState<string>(convertMinutes === '0'?'00':convertMinutes);
```
atomworks, line 124: We have similar logic being used to set values in `onBlurInputs`. Could be have a function that handles these cases and can be re-used more cleanly? At the moment the code repetition could be problematic if this behaviour is iterated again in the future.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -143,12 +165,24 @@ const DateTimeBlock : React.FC<IProps> = ({
...
+    }
+    if(displayMinutes ==='0' || displayMinutes ===''){
+      setDisplayMinutes('00');
+    }
+    if(displayMinutes.length === 1){
+      setDisplayMinutes('0' + displayMinutes);
+    }
+    if(displayHours.length === 1){
+      setDisplayHours('0' + displayHours);
+    }
```
atomworks, line 183: As mentioned in the previous code comment, it would be better to handle these in a unified function so it's easier to read and update later if needed.

atomworks (changes requested):
Just a small change request, mostly a minor refactoring.

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: Done

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: Done

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -118,10 +119,10 @@ const DateTimeBlock : React.FC<IProps> = ({
   setDateCallback = ()=>{},
 }) => {
 
-  const convertHours = (date.getHours()).toString();
-  const convertMinutes = date.getMinutes().toString();
-  const [displayHours, setDisplayHours] = useState<string>(convertHours === '0'?'00':convertHours);
-  const [displayMinutes, setDisplayMinutes] = useState<string>(convertMinutes === '0'?'00':convertMinutes);
+  const convertHours = (date?.getHours()).toString();
+  const convertMinutes = (date?.getMinutes()).toString();
```
atomworks, line 123: What happens if `date` is `undefined`? Currently it seems this prop is optional and then as this is used to set the default state of the `displayHours` and `displayMinutes`. However `getFormattedTime` currently needs `value` to be a string.

It seems like there is an error case here that needs fixing. I would have thought Typescript would be giving you an error for this but if I've misunderstood something, please let me know!

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -168,20 +169,14 @@ const DateTimeBlock : React.FC<IProps> = ({
     }
   },[date, allowAfterMidnight]);
 
-  const onBlurInputs = useCallback(() =>{
-    if(displayHours === '0' || displayHours === ''){
-      setDisplayHours('00');
+  const onBlurInputs = useCallback((blurValue: string, value:string) =>{
```
atomworks, line 172: `value` is being used here to to tell use if we're usings minutes or hours. However the variable name implies that it is a value (like a minute / hour) rather than what it is. Could you give these both names that best match their intention?

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -168,20 +169,14 @@ const DateTimeBlock : React.FC<IProps> = ({
...
+    const convertedValue = getFormattedTime(blurValue);
+    if(value === 'Minutes'){
```
atomworks, line 174: Having a string like this is fine but it would be better to keep it as lower case. As the options here are limited we might be able to get aways with `mins` or even just `m` so minutes.

atomworks (changes requested):
Some small change requests. Mostly just making sure we follow best practices for readability and usability for engineers. Thanks for your continued effort on this... it's nearly there!

src/helpers/index.tsx
```diff
@@ -81,6 +81,17 @@ const uniqueID = () =>
       Math.random().toString(16)
   ).replace(/\./g, '');
 
+const getFormattedTime = (value:string) =>{
+  if(value ==='0' || value ===''){
+    return '00';
+  }
+  if(value?.length === 1){
```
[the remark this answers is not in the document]
fti-pfarne: Done

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: I have add one extra condition in getFormattedTime() . When value is undefined it will return '00'

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: I have changed.
Minutes --> mins
Hour --> hours

src/Filters/atoms/DateTimeBlock.tsx
atomworks: The `getFormattedTime` is still set to require a string, so in the instance of undefined it's going to error rather than reach the new `else` state.

When you compile are there any errors from Typescript? I'm surprised it's not catching that one.

atomworks (commented):
One more detail, I've added the note to the currently open thread.

src/Filters/atoms/DateTimeBlock.tsx
fti-pfarne: Yes there is no error. I have checked.

atomworks (approved):

PR #395, opened 2023-01-09
Fix/broken image table thumbnail

fti-smaske:
Description
This PR is for No Image in thumbnail for table to avoid broken image and remove unnecessary console logs.

In case image has url but its not valid then it will show No Image in thumbnail.

Related Bugherd Requirement links
[Requirement](https://www.bugherd.com/projects/216134/tasks/56)

Considerations on Implementation
Tested this work in TCC project.

image


fti-skumar (approved):
LGTM

src/Tables/atoms/TableRowThumbnail.tsx
```diff
@@ -136,17 +149,44 @@ const TableRowThumbnail: React.FC<IProps> = ({ hoverZoom = true, image='', media
...
 
-
-
   const onLoad = useCallback(()=>{
     timeoutRef.current && clearTimeout(timeoutRef.current);
     timeoutRef.current = null;
     setShowImage(true);
   },[]);
 
+  const checkIfImageExists = (url: string, callback: (exists: boolean) => void) => {
```
fti-bcarpenter, line 158: callback is not a good name here. Use some different name like imageExistsCallback

fti-bcarpenter (changes requested):

src/Tables/atoms/TableRowThumbnail.tsx
fti-smaske: Done

fti-bcarpenter (approved):

atomworks (approved):
There was one thing I wasn't sure about so I've highlighted that in a comment for Josh to check. Other than that though, happy to approve.

PR #396, opened 2023-01-31
Filter: Fix filter last item crop issue

fti-rchaniya:
### Description
This PR consist a following bugfix:
All the filter dropdowns in the scorer-ui-kit, were having maximum 5 displayed options, among which last item was getting cropped as shown in below attached screenshot

**snapshot of ui having bug:**
image
image

So, as a fix of this bug, instead of increasing a max-height of OptionList from 162px to 170px, as per the suggested comment in the PR I have added a linear-gradient property to the option list if the options in the list are greater than 5 which looks proper as shown in below attached screenshot

**Here are some screenshots of the expected result.**
image
image
image
image
image

 ### Reviewing/ Testing Steps
 You can review this feature in this page
 https://future-standard.github.io/scorer-ui-kit/storybook/?path=/story/filters-molecules--filter-dropdown

### DIFF
--- a/src/Filters/molecules/FilterDropdown.tsx
+++ b/src/Filters/molecules/FilterDropdown.tsx
@@ -90,12 +90,12 @@ const EmptyResultText = styled.div`
   font-size: 12px;
 `;
 
-const Gradient = styled.div<{hasSearchOption: boolean}>`
+const Gradient = styled.div`
   position: absolute;
   right: 0px;
-  top: ${({ hasSearchOption }) => hasSearchOption ? '223px' : '149px'};
+  bottom: 9px;
   height: 25px;
-  background-image: linear-gradient(to bottom, rgba(246, 247, 249, 0) 1%, #F6F7F9 81%);
+  background-image: linear-gradient(to bottom, rgba(246, 247, 249, 0) 1%, #F6F7F9 120%);
   width: 100%;
 `;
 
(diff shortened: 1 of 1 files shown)


fti-bcarpenter (approved):
LGTM

fti-skumar (approved):

fti-smaske (approved):

atomworks (commented):
This was actually kind of a feature as it gave the user a way to visibily see there was something beyond the current view. 

It was a bit messy though... do you think instead of cropping this final one (as it was before) we could add a small white-to-transparent gradient to make it tidier?

fti-rchaniya:
As per the suggestion, instead of increasing a max-height of OptionList from 162px to 170px, I have added a linear-gradient property (linear-gradient(to bottom, rgba(246, 247, 249, 0) 1%, #F6F7F9 81%);) to the option list if the options in the list are greater than 5 which gives the user a way to visibly see there is something beyond the current view,  as shown in the attached screenshot in the description

src/Filters/molecules/FilterDropdown.tsx
```diff
@@ -89,6 +90,15 @@ const EmptyResultText = styled.div`
   font-size: 12px;
 `;
 
+const Gradient = styled.div<{hasSearchOption: boolean}>`
+  position: absolute;
+  right: 0px;
+  top: ${({ hasSearchOption }) => hasSearchOption ? '223px' : '149px'};
```
atomworks, line 96: Wouldn't this be simpler to do using `bottom` instead? That should negate the need for a distance variable and will mean that if we want to vary height in the future, the gradient will move with it.

atomworks (changes requested):
One small request on a line of CSS. Other than that, it's looking great... thank you!

src/Filters/molecules/FilterDropdown.tsx
fti-rchaniya: > Wouldn't this be simpler to do using `bottom` instead? That should negate the need for a distance variable and will mean that if we want to vary height in the future, the gradient will move with it.

Done. Thanks for the feedback.

atomworks (approved):

PR #402, opened 2023-03-20
On click over NoImage thumbnail, media modal should not open issue re…

fti-rchaniya:
**Description**
This PR has following bugfix in the TableRowThumbnail component:

If there is no image url or broken image url or distorted image url, then in all these cases, the thumbnail in the table showing No image as shown in below attached screenshot, else showing the exact image.
image

In case of No image thumbnail, somewhere onHover pointer was showing so somewhere not, same for some No image thumbnail, media modal was getting open as shown in below attached screenshot so somewhere not.
image

**Requirement**
Ideally if it is showing No image thumbnail, then pointer event should not be there on hovering over it and also a media modal should not get open.

1. Added showImage boolean state(use to check if the coming image is exists or not and consist boolean value based on the image url) to the onHover css to manage pointer event.
2. Added showImage boolean state to the handleModal function based on which a media modal will get open

**Note**- While using this component, user should not handle condition for "No Image" scenario as it is already handle in the component itself, no need to check if image exists or not

### DIFF
--- a/src/Tables/atoms/TableRowThumbnail.tsx
+++ b/src/Tables/atoms/TableRowThumbnail.tsx
@@ -7,7 +7,7 @@ import { NoImage } from '../../svg';
 
 type VideoAspects = '4:3' | '16:9';
 
-const Container = styled.div<{ hoverZoom?: boolean, aspect?: VideoAspects, mediaUrl?: string}>`
+const Container = styled.div<{ hoverZoom?: boolean, aspect?: VideoAspects, mediaUrl?: string, showImage?: boolean}>`
   position: relative;
   height: inherit;
   background: grey;
@@ -34,7 +34,7 @@ const Container = styled.div<{ hoverZoom?: boolean, aspect?: VideoAspects, media
   `}
 
   &:hover {
-    ${({ mediaUrl }) => mediaUrl && css`
+    ${({ mediaUrl, showImage }) => mediaUrl && showImage && css`
       cursor: pointer;
     `};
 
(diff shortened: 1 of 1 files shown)


fti-skumar (approved):
LGTM

fti-pfarne:
LGTM

fti-bcarpenter (approved):
LGTM

src/Misc/atoms/MediaBox.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useState, useCallback, VideoHTMLAttributes } from 'react';
 import styled, { css } from 'styled-components';
 import Spinner from '../../Indicators/Spinner';
 import { IMediaType } from '../../index';
-import Icon from '../../Icons/Icon';
+import { ReactComponent as LargeNoImage } from '../../svg/Large-No-image.svg';
```
fti-bcarpenter, line 5: Please change this to **NoImageBig.svg**

LargeNoImage  --> **NoImage**

fti-smaske (approved):

src/Misc/atoms/MediaBox.tsx
```diff
@@ -123,19 +123,21 @@ const MediaBox: React.FC<IMediaModal> = ({
...
         : <StyledImage
-            {...{ src, alt, onError, hasModalLimits }}
+            {...{ alt, onError, hasModalLimits }}
+            src={loadFailed ? '' : src}
             onLoad={handleLoad}
             isLoaded={loaded && !loadFailed}
           />}
       {(!loaded) && <LoadingOverlay><Spinner size='large' styling='primary' /></LoadingOverlay>}
-      {loadFailed && <LoadingOverlay><Icon icon='MissingImage' size={48} /></LoadingOverlay>}
+      {loadFailed && <LargeNoImage />}
```
fti-bcarpenter, line 140: NoImage

fti-bcarpenter (changes requested):

src/Misc/atoms/MediaBox.tsx
fti-rchaniya: done

fti-bcarpenter (approved):
LGTM

src/Misc/atoms/MediaBox.tsx
fti-rchaniya: done

atomworks (approved):

PR #404, opened 2023-03-20
DatePicker minute field roll over

fti-rchaniya:
### Description
This PR has following change in the minute input of DatePicker component:

 - When we keep on increasing minute value, it reaches to 59 and hold 59 when you continue increasing, we need roll over to 0 and start again from 0 once you increase after 59.
 - When we keep on decreasing minute value, it reaches to 0 and hold 0 when you continue decreasing, we need roll over to 59 and start again from 59 once you decrease after 0.
 
  **BH link**
  [Requirement](https://www.bugherd.com/projects/216134/tasks/195)

 ### Considerations on Implementation
 To achieve this, I'm checking the entered value and if it is greater than 59 then resetting value to 0 and if it is less than 0 then resetting it to value 59 and if both the condition does not meet then setting entered value as it is.

PR #405, opened 2023-03-21
DatePicker - 2 digit time format in the hours input field

fti-rchaniya:
### Description

This PR has following change in the Hour input of DatePicker component:

- While pressing UP arrow(incrementing) of the input field from 0 to 9, it shows a single digit value, ex- 0, 1, 2, 3, 4, 5, 6, 7, 8, 9  shown in below attached screenshot:

image

- While pressing UP arrow(incrementing) of the input field from 10 to 24 it shows 2 digit value
- Similarly while pressing DOWN arrow(decrementing) of the input field from 9 to 0 it shows single digit value
- Expected result is, for the single digit value from 0 to 9 it must show 2 digit representation such as 00, 01, 02, 03, 04, 05, 06, 07, 08, 09
- Also while entering value manually, if I enter 1 then it should show 01 and if i enter 11 then it must show 11.
- decimal values and exponential constant were allowing, which must be restricted

**Additional Requirement** 
- I have also implemented a rollover of the hours field value.
- Example: When the user increments the value from 23, it should roll over to 00. Similarly, when the user decrements the value from 00, it should roll over to 23.

- To achieve 2 digit time format, I'm padding 0 with the entered value as given below:
**setDisplayHours(value.padStart(2, '0'))** 

But I have handled this 0 padding conditionally, If I won't handle this conditionally then a user won't be able to delete entered value manually and enter a new value from keyboard because it will pad 00 if backspace hit occurs, and it will again raise below issue.
[BH](https://www.bugherd.com/projects/216134/tasks/132)

- We want 0 padding only when **UP/DOWN keyboard keys** or **UP/DOWN ui input box arrow** hit occurs, any keyboard entries other than **UP/DOWN keyboard keys** hit occurs then it should not pad 0.

- To make this happen, I used below state
const [isNonArrowKey, setIsNonArrowKey] = useState(false);

- Making isNonArrowKey = true only when **KeyboardEvent** occurs and **event.key** is not equal to"ArrowUp" or "ArrowDown"

- So based on isNonArrowKey state we padding 0 to the value as shown below:
**isNonArrowKey ? setDisplayHours(value) : setDisplayHours(value.padStart(2, '0'));**

 
**Here are some screenshots of the expected result.**

image

**Kindly go through attached Screencast for better clarity of implementation**

https://github.com/future-standard/scorer-ui-kit/assets/118800413/807ca74e-a5ae-4571-8e8f-faa8475649f1

**Important Note** -> 
I'm merging the changes of below PR in this PR only as there are some common functions or logic we are using in both PR.
https://github.com/future-standard/scorer-ui-kit/pull/404

### DIFF
--- a/src/Filters/atoms/DateTimeBlock.tsx
+++ b/src/Filters/atoms/DateTimeBlock.tsx
@@ -123,14 +123,14 @@ const DateTimeBlock : React.FC<IProps> = ({
   const convertMinutes = (date?.getMinutes()).toString();
   const [displayHours, setDisplayHours] = useState<string>(getFormattedTime(convertHours));
   const [displayMinutes, setDisplayMinutes] = useState<string>(getFormattedTime(convertMinutes));
-  const [isKeyboardEvent, setIsKeyboardEvent] = useState(false);
+  const [isNonArrowKey, setIsNonArrowKey] = useState(false);
 
   const handleKeyDown = useCallback((event: React.KeyboardEvent<HTMLInputElement>) => {
     const key = event.key;
     if (key === "ArrowUp" || key === "ArrowDown") {
-      setIsKeyboardEvent(false);
+      setIsNonArrowKey(false);
     } else {
-      setIsKeyboardEvent(true);
+      setIsNonArrowKey(true);
     }
   },[]);
 
@@ -139,7 +139,7 @@ const DateTimeBlock : React.FC<IProps> = ({
     if (!hourRegex.test(value)) {
       return;
     }
-    isKeyboardEvent ? setDisplayHours(value) : setDisplayHours(value.padStart(2, '0'));
+    isNonArrowKey ? setDisplayHours(value) : setDisplayHours(value.padStart(2, '0'));
     setDateCallback(
       min([
         endOfDay(date),
@@ -151,8 +151,8 @@ const DateTimeBlock : React.FC<IProps> = ({
         })
       ])
     );
-    setIsKeyboardEvent(false);
-  }, [date, displayMinutes, setDateCallback, isKeyboardEvent]);
+    setIsNonArrowKey(false);
+  }, [date, displayMinutes, setDateCallback, isNonArrowKey]);
 
   const setDateMinutes = useCallback(({target: {value}}: React.ChangeEvent<HTMLInputElement>) => {
     const minuteRegex = /^[0-5]{0,1}[0-9]{0,1}$/;
--- a/src/Filters/atoms/DateTimeBlock.tsx
+++ b/src/Filters/atoms/DateTimeBlock.tsx
@@ -125,20 +125,23 @@ const DateTimeBlock : React.FC<IProps> = ({
   const [displayMinutes, setDisplayMinutes] = useState<string>(getFormattedTime(convertMinutes));
 
   const setDateHours = useCallback(({target: {value}}: React.ChangeEvent<HTMLInputElement>) => {
-    let newVal = '0';
-    if(value.length === 1){
+    let newVal;
+    if(Number(value) > 24){
+      const minuteRegex = /^-?\d{1,2}$|^$/;
+      if (!minuteRegex.test(value)) {
+        return;
+      }
+      newVal = '24';
+    } else if(Number(value) < 0){
+      newVal = '00';
+    } else if(value.length === 1){
       if(value === '0'){
-        newVal = ('0');
+        newVal = ('00');
       } else {
         newVal = ('0' + value).slice(-2) ;
       }
     } else {
       newVal = value.slice(-2);
-    }
-    if(Number(newVal) > 24){
-      newVal = '24';
-    } else if(Number(newVal) < 0){
-      newVal = '0';
     }
     setDisplayHours(newVal);
     setDateCallback(
@@ -189,6 +192,14 @@ const DateTimeBlock : React.FC<IProps> = ({
     }
   }, []);
 
+  const onKeyDownChange = useCallback((event) => {
+    if(event.key==='.' || event.key==='e' || event.key==='-'){
+      event.preventDefault();
+    } else if (event.key==='Backspace'){
+      event.target.name === 'hours' ? setDisplayHours('') : setDisplayMinutes('');
+    }
+  },[]);
+
   return (
     <Container hide={!hasDate && !hasTime}>
       <Label>{title}</Label>
@@ -210,7 +221,7 @@ const DateTimeBlock : React.FC<IProps> = ({
             <Icon icon='Time' color='dimmed' size={14} weight='light' />
           </IconWrap>
           <InputWrap>
-            <InputValue onBlur={()=>onBlurInputs(displayHours, 'hours')} {...{allowManualTimeChange}} name='hours' type='number' value={displayHours} onChange={setDateHours} autoComplete='off' />
+            <InputValue onKeyDown={(e) => onKeyDownChange(e)} onBlur={()=>onBlurInputs(displayHours, 'hours')} {...{allowManualTimeChange}} name='hours' type='number' value={displayHours} onChange={setDateHours} autoComplete='off' />
             <TimeColon>:</TimeColon>
             <InputValue onBlur={()=>onBlurInputs(displayMinutes, 'mins')} {...{allowManualTimeChange}} name='minutes' type='number' value={displayMinutes} onChange={setDateMinutes} autoComplete='off' />
           </InputWrap>
--- a/src/Filters/atoms/DateTimeBlock.tsx
+++ b/src/Filters/atoms/DateTimeBlock.tsx
@@ -123,38 +123,61 @@ const DateTimeBlock : React.FC<IProps> = ({
   const convertMinutes = (date?.getMinutes()).toString();
   const [displayHours, setDisplayHours] = useState<string>(getFormattedTime(convertHours));
   const [displayMinutes, setDisplayMinutes] = useState<string>(getFormattedTime(convertMinutes));
+  const [isKeyboardEvent, setIsKeyboardEvent] = useState(false);
 
-  const setDateHours = useCallback(({target: {value}}: React.ChangeEvent<HTMLInputElement>) => {
-    let newVal;
-    if(Number(value) > 24){
-      const hourRegex = /^-?\d{1,2}$|^$/;
-      if (!hourRegex.test(value)) {
-        return;
-      }
-      newVal = '24';
-    } else if(Number(value) < 0){
-      newVal = '00';
-    } else if(value.length === 1){
-      if(value === '0'){
-        newVal = '00';
-      } else {
-        newVal = ('0' + value).slice(-2) ;
-      }
+  const handleKeyDown = useCallback((event: React.KeyboardEvent<HTMLInputElement>) => {
+    let value = event.currentTarget.value;
+    const key = event.key;
+    // const hourRegex  = /^[0-1]{0,1}[1-9]{0,1}$|(^2[0-3])$/;
+    // if (!hourRegex.test(value)) {
+    //   return;
+    // }
+    if (key === "ArrowUp") {
+      setIsKeyboardEvent(false);
+      // event.preventDefault();
+      // let hours = parseInt(value);
+      // setDisplayHours((hours + 1).toString().padStart(2, '0'));
+    } else if (key === "ArrowDown") {
+      setIsKeyboardEvent(false);
+      // event.preventDefault();
+      // let hours = parseInt(value);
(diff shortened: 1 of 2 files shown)


fti-smaske (approved):

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -125,16 +125,27 @@ const DateTimeBlock : React.FC<IProps> = ({
...
+      } else {
+        newVal = ('0' + value).slice(-2) ;
+      }
+    } else {
+      newVal = value.slice(-2);
+    }
+    if(Number(newVal) > 24){
+      newVal = '24';
+    } else if(Number(newVal) < 0){
+      newVal = '0';
```
fti-skumar, line 141: We can still add single '0'. Needs to change to '00'.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -125,16 +125,27 @@ const DateTimeBlock : React.FC<IProps> = ({
...
+    let newVal = '0';
+    if(value.length === 1){
+      if(value === '0'){
+        newVal = ('0');
+      } else {
+        newVal = ('0' + value).slice(-2) ;
+      }
+    } else {
+      newVal = value.slice(-2);
+    }
```
fti-skumar, line 137: We should use `padStart` function to pad 0, and this should be handled with keyboard events for specific up & down keys and arrow buttons in UI. If we handle it here then backspacing will also pad 0 and as user I cannot freely input anything without this padding.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -125,16 +125,27 @@ const DateTimeBlock : React.FC<IProps> = ({
...
+    if(Number(newVal) > 24){
```
fti-skumar, line 138: After adding single digit if I enter next which makes whole number greater than 24 then 24 will be explicitly added, in this scenario we should not allow user to enter, for ex- if I enter 04 then if I press 0, then it will become 24, instead 0 shouldn't be allowed there.

fti-skumar (changes requested):
Functionality wise its working correctly but, feels like user experience can be enhanced. Please check comments.

fti-skumar:
Also are we going to handle minute field in different PR ?

src/Filters/atoms/DateTimeBlock.tsx
fti-rchaniya: Done

fti-rchaniya: Please note that, to fix this I have updated logic that whenever user hit backspace event, it will clear both the digit in the hours input field.

src/Filters/atoms/DateTimeBlock.tsx
fti-rchaniya: We can handle 0 padding in the onKeyDown function when user press Up/Down arrows from keyboard, but in order to handle 0 padding onchange of the arrow buttons in UI then it will call the setDateHours onChange function and hence I have handled this 0 padding here and therefore when a user enters manually then also it triggers onChange event (setDateHours function) which pads 0 to the entered value

src/Filters/atoms/DateTimeBlock.tsx
fti-rchaniya: Done

fti-rchaniya:
> Also are we going to handle minute field in different PR ?

Yes, for minute field there is different PR given below:
[MinuteFieldChanges](https://github.com/future-standard/scorer-ui-kit/pull/404)

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -125,16 +125,30 @@ const DateTimeBlock : React.FC<IProps> = ({
   const [displayMinutes, setDisplayMinutes] = useState<string>(getFormattedTime(convertMinutes));
 
   const setDateHours = useCallback(({target: {value}}: React.ChangeEvent<HTMLInputElement>) => {
-    const hourRegex  = /^[0-1]{0,1}[0-9]{0,1}$|(^2[0-4])$/;
-    if (!hourRegex.test(value)) {
-      return;
+    let newVal;
+    if(Number(value) > 24){
+      const minuteRegex = /^-?\d{1,2}$|^$/;
```
fti-skumar, line 130: This should be names as `hourRegex`

fti-skumar (changes requested):

fti-skumar (approved):

fti-bcarpenter (approved):
LGTM

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -178,6 +192,14 @@ const DateTimeBlock : React.FC<IProps> = ({
     }
   }, []);
 
+  const onKeyDownChange = useCallback((event) => {
+    if(event.key==='.' || event.key==='e' || event.key==='-'){
+      event.preventDefault();
+    } else if (event.key==='Backspace'){
+      event.target.name === 'hours' ? setDisplayHours('') : setDisplayMinutes('');
+    }
```
atomworks, line 200: This is too aggressive. For example, if I want to type `12` but I accidently type `11`, it's expected that if I hit `backspace` once it will delete the last digit and then I can simply hit `2`. However, the user will then realise that it says `2` and not `12` and will have to start again, which will cause them furstration that we can more easily avoid here.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -178,6 +192,14 @@ const DateTimeBlock : React.FC<IProps> = ({
     }
   }, []);
 
+  const onKeyDownChange = useCallback((event) => {
+    if(event.key==='.' || event.key==='e' || event.key==='-'){
```
atomworks, line 196: I can see why you're stopping `.` from firing here, but I couldn't replicate `e` or `-` causing any problems in these fields at all. Are these needed for a reason I am not realising?

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -125,16 +125,30 @@ const DateTimeBlock : React.FC<IProps> = ({
...
+      newVal = '00';
+    } else if(value.length === 1){
+      if(value === '0'){
+        newVal = '00';
+      } else {
+        newVal = ('0' + value).slice(-2) ;
+      }
+    } else {
+      newVal = value.slice(-2);
     }
```
atomworks, line 145: Could you explain what the logic of this code is quickly? It's a bit difficult to follow and where I am following it, I'm unsure if it's the right place.

For example, the last work done in the area used a new helper called `getFormattedTime` to make sure the leading `0` was in place where needed. This however seems to reproduce that functionality again. On this current pass on the code, it seems like even that extra helper function wasn't needed as that code should already have been handled elsewhere in the system. 

Also, there are better place to enforce maximum and minimum values here if that was missing by using the browser native handling using attributes on the input element.

It also seems to be ignoring the `allowAfterMidnight` prop that would allow people to enter a time such as 25:00 as a way to say 01:00 + 1 day.

atomworks (changes requested):
I know this is a difficult component to work on although it seems that we're starting to replicate code and lose track of some important things that were already in place.

src/Filters/atoms/DateTimeBlock.tsx
```diff
@@ -123,13 +123,23 @@ const DateTimeBlock : React.FC<IProps> = ({
...
   const [displayMinutes, setDisplayMinutes] = useState<string>(getFormattedTime(convertMinutes));
+  const [isKeyboardEvent, setIsKeyboardEvent] = useState(false);
+
+  const handleKeyDown = useCallback((event: React.KeyboardEvent<HTMLInputElement>) => {
+    const key = event.key;
+    if (key === "ArrowUp" || key === "ArrowDown") {
+      setIsKeyboardEvent(false);
+    } else {
+      setIsKeyboardEvent(true);
+    }
```
fti-skumar, line 134: Here, Every event is an keyboard event so I think we need to reconsider name of this state something more explanatory.

fti-rchaniya: Done

fti-skumar (approved):
LGTM
Thanks for the update

fti-rchaniya:
> I know this is a difficult component to work on although it seems that we're starting to replicate code and lose track of some important things that were already in place.

@atomworks san, To begin with, I apologize for the delay in providing this PR update.

src/Filters/atoms/DateTimeBlock.tsx
fti-rchaniya: Absolutely agree your perspective and hence I have modified a code. 
Thanks

src/Filters/atoms/DateTimeBlock.tsx
fti-rchaniya: Updated code.

src/Filters/atoms/DateTimeBlock.tsx
fti-rchaniya: > Could you explain what the logic of this code is quickly?
@atomworks san, I have updated the logic for this, kindly review the updated one.

fti-rchaniya: > It also seems to be ignoring the allowAfterMidnight prop that would allow people to enter a time such as 25:00 as a way to say 01:00 + 1 day.

@atomworks san, I tried to understand and track the allowAfterMidnight prop's concept, but fails to understand. It seems that this props does not have any impact in the earlier code. 
I have attached a screencast of my findings, can you please go through it and help me to understand so that I can proceed in the correct direction to meet your expectation.

https://github.com/future-standard/scorer-ui-kit/assets/118800413/6abc4401-fc97-4158-bc08-e04613b74141

fti-rchaniya: @atomworks san, as per our recent conversation we had on slack, thread is shared below. We will create a separate PR for midnight issue and hence we can move forward to resolve the other points in this PR
[Slack Thread](https://fti-futurestandard.slack.com/archives/C01FYNR7D0D/p1699497665734719?thread_ts=1681108222.921939&cid=C01FYNR7D0D)
Thanks

atomworks (commented):
Where did the original request for the number rollover when use the arrow keys come from? Because rolling past midnight doesn't also iterate the date it causes an error because for example, the start time can become later than the end time. 

If the idea is someone wants to go from 11pm to 3am, then we'd need to increment the date. But before saying that the date should be incremented on the rollover it would be good to know where the original request came from. Thanks.

fti-rchaniya:
> Hey, I. have reviewed the functionality as described it's doing what the Bugherd required. However as Leonard mentioned incrementing from 0:59 mins to the next perhaps should also increment hours 1:00 and at 23:59 probably requires update of Day
> 
> ```
> 0:59 ->  key up -> 1:00
> 1:00 -> key down -> 0:59
> ```
> 
> A place where I have seen this kind of feature is on google maps route calculation with public transportation at depart at or arrive at. When you move key up it goes 04:59 -> 5:00 and at the same time when going down 4:00 -> 3:55
> 
> I see that in the example @h2suzuki was mainly concerned about the minutes increment and the hours increment was then reviewed and change added but not at the preferred level. The checking of hours after minutes increment that @atomworks mention is also important. Perhaps this could be follow up in Slack for transparency and to avoid confusion.
> 
> My recommendation in this topic is always add the bugherd/jira/github issue to the slack channel when requesting the PR review ( I would suggest in the same PR but since this is an open source repository, I don't think is the place for commenting private projects)
> 
> -- My opinion on another topic -- I have a concern about the breaking the allowAfterMidnight. This is a prop that is used for video time range ( imagine a video that 23:59 pm and continued until 24:00 ). This is for not loosing one second I've seen projects like "Traffic Counter Sass" rely on this functionality.
> 
> If we commit the code as it is we will need to provide a warning that this code has breaking changes for Calendar usage on the next release since the projects will need to add extra steps to avoid the error in their side.

Hello @isacoder ,

Thanks for your detailed review and feedback! I appreciate your thorough examination of the functionality. I agree with your observations regarding the incrementing from 0:59 mins to the next, and I see the relevance of considering hours as well.

Regarding your concern about breaking the allowAfterMidnight prop, I understand the importance of maintaining compatibility.  I've already had a discussion with @atomworks-san, and we've decided to plan a separate PR to address this concern.

Let's connect with @h2suzuki-san to discuss the possibility of postponing the PR related to his reported Bugherd, ensuring a more coordinated approach.

Thanks again for your collaboration and insights!

isacoder:
Closing this PR since requirements are covered on another PR - https://github.com/future-standard/scorer-ui-kit/pull/444

PR #406, opened 2023-03-22
Left sidebar >> Menu name should not be hidden in 1368X768 and 1024X720 resolutions

fti-rchaniya:
### Description

This PR has following change in the Left sidebar menu of the Global Ui component:

 - When the menu options exceeded, "Keep Open/Auto-Hide" footer text hides for some screen resolutions as shown in below attached screenshot
 
image

 - When Sub menus of the menu options exceeded then the sub menus are also hides as shown in attached screen shot
 
image

- As there is no vertical scroll bar a user cannot scroll down and see the hidden options or menus or footer 

- Added a vertical scroll bar to the menu options with min and max height by keeping a position of logo & text and Footer text fixed
- Also a scroll bar is hidden in all the browser as its visibility on ui does not seems correct.

Here is the screencast of the expected result.
https://www.awesomescreenshot.com/video/15820748?key=3e8aaf86b3e648da83a8b0924d1f2efe

As per the user readability, if there are more items then those items get hides and user cannot understand whether there is more menu or not, as shown in above screencast, hence to improve this we have added a linear-gradient at the bottom to show there are more items to scroll, please refer below screencast:
https://www.awesomescreenshot.com/video/16031733?key=ed935505a7338a8a74111b4f7b9f9e75

As there was more space showing between the Menu container and footer container, I have increased max-height of the Navigation container & decreased height of the Footer item container, as shown in below attached screenshot:
image

### DIFF
--- a/src/Global/organisms/MainMenu.tsx
+++ b/src/Global/organisms/MainMenu.tsx
@@ -66,7 +66,7 @@ const NavigationContainer = styled.div`
   }
   -ms-overflow-style: none;  /* IE and Edge */
   scrollbar-width: none; /* Firefox */
-  max-height: 400px;
+  max-height: 500px;
 
   &::after {
     background: linear-gradient(rgba(246, 247, 249, 0) 1%, #F6F7F9 120%) center bottom;
--- a/src/Global/organisms/MainMenu.tsx
+++ b/src/Global/organisms/MainMenu.tsx
@@ -49,12 +49,12 @@ const SVGObjectText = styled.object`
 const GradientContainer = styled.div`
   &::after {
     content: '';
-    background: radial-gradient(farthest-side at 50% 100%,rgba(246, 247, 249, 0) 1%, #F6F7F9 120%) center bottom; //Shodow bottom
+    background: linear-gradient(to bottom, transparent, #e5e5e5); //Shodow bottom
     height: 10px;
     display: block;
     position: absolute;
     margin-top: -10px;
-    width: 300px;
+    width: 100%;
   }
 `;
 
--- a/src/Global/organisms/MainMenu.tsx
+++ b/src/Global/organisms/MainMenu.tsx
@@ -194,21 +194,21 @@ const MainMenu: React.FC<IMenu> = ({ content, home = "/", logoMark, logoText, ke
               <LogoType>{logoText ? <SVGObjectText type='image/svg+xml' data={logoText} /> : <SvgLogoText />}</LogoType>
             </Logo>
             <GradientContainer>
-            <NavigationContainer>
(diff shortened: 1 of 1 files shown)


fti-smaske (approved):

fti-skumar (approved):

fti-skumar:
It works as expected but in some cases other menu items get hidden if one is expanded as shown in below snap

- image

We can have a shadow or something else to tell user that its scrollable.

fti-rchaniya:
> It works as expected but in some cases other menu items get hidden if one is expanded as shown in below snap
> 
> * image
> 
> We can have a shadow or something else to tell user that its scrollable.

Thanks for the feedback, suggested changes have been implemented

src/Global/organisms/MainMenu.tsx
```diff
@@ -46,7 +46,38 @@ const SVGObjectText = styled.object`
   max-width: 180px;
 `;
 
-const NavigationContainer = styled.div``;
+const GradientContainer = styled.div`
+  &::after {
+    content: '';
+    background: radial-gradient(farthest-side at 50% 100%,rgba(246, 247, 249, 0) 1%, #F6F7F9 120%) center bottom; //Shodow bottom
```
fti-skumar, line 52:
```suggestion
    background: linear-gradient(to bottom, transparent, #e5e5e5); //Shodow bottom
```
Just a suggestion,
Because current one is not even seem like menu items are behind it. You can also use a different one if that looks better.

**Before**

image

**After**

image

src/Global/organisms/MainMenu.tsx
```diff
@@ -46,7 +46,38 @@ const SVGObjectText = styled.object`
...
+    height: 10px;
+    display: block;
+    position: absolute;
+    margin-top: -10px;
+    width: 300px;
```
fti-skumar, line 57:
```suggestion
    width: 100%;
```

To fix overflow.

image

fti-skumar (approved):
Thanks LGTM

isacoder (changes requested):
Hello!

Thanks for working on this feature, navigation menu is a very tricky component.
Also thank you for adding a video of how this works on your environment.

I run this on my local on the resolution 1368X768 and is not working as the video. I see that the height in this case is 538px and the max height is 550px, probably related to that.

Screenshot 2024-01-22 at 23 06 58

https://drive.google.com/file/d/1YfzDjkYu8SbXN1qIE5ikoi7gQJeyr-cM/view?usp=sharing

atomworks:
This PR looks to be abandoned, so I am closing it. Please re-open if / when required and you are able to complete the work. Thank you.

PR #407, opened 2023-04-03
Added "Clear Files" button to Area Upload Manager to reset selected files

fti-rchaniya:
**Description**
This PR has following change in the Area Upload Manager component:

- While using this component, if an user wants to set selected files to null by an external triggered event such as close icon or button as shown in below attached screenshot, then the selected files does not reset to null due to the absence of prop which setFiles(null).
image

- Due to this, even if user click on close icon, and add files again so the expected behaviour is the selected file's count start from 0 which is not, it starts adding in the array where the previously added files are saved. 
- For more understanding, kindly refer below screencast:
https://www.awesomescreenshot.com/video/16183735?key=219730f4f27ff7f0ec403ed16e3e3611

- To achieve this, earlier we thought to add 1 Boolean state to the component which an user can use as a prop to setFiles(null) as shown in below attached screenshot.
image

- Instead of the above solution, we approaches another suggestion given in this [comment](https://github.com/future-standard/scorer-ui-kit/pull/407#pullrequestreview-1377390065).
- As per the above comment, instead of adding prop to trigger external event, we have added a "Clear Files" button in the component itself to clear selected files.
- Also "Clear Files" button will be only visible if the selected files count is more than 0, it is aligned parallelly with the "Select Files"/"Add More Files" button as shown in the below attached screenshot:
image

Screencast of the final implementation:
https://www.awesomescreenshot.com/video/16402979?key=eaec76b4966083b3e92b269a375fde97


src/Form/organisms/AreaUploadManager.tsx
```diff
@@ -105,6 +105,7 @@ interface IAreaUploaderManager {
   allowedFileTypes?: string[]
   customComponent?: ReactElement
   onChangeCallback?: (goodFiles: FileList, rejectedFiles: FileList) => void
+  clearState?: boolean;
```
fti-smaske, line 108: Change varaible name to `clearFiles`

fti-skumar (approved):
LGTM, just change the name as suggested by @fti-smaske

src/Form/organisms/AreaUploadManager.tsx
fti-rchaniya: Done, thanks for the feedback

fti-smaske (approved):

storybook/src/stories/Form/FileManagement/AreaUploadManager.stories.tsx
```diff
@@ -27,6 +27,7 @@ export const _AreaUploadManager = () => {
   const fileIcons = object('File Icons', defaultIcons)
   const showValue = action('Input Callback');
   const allowedFileTypes = object('allowedFileTypes', fileTypes);
+  let clearState = boolean('Clear selected Files', false);
```
fti-bcarpenter, line 30: This should be `clearFiles`. Isn't it?

storybook/src/stories/Form/FileManagement/AreaUploadManager.stories.tsx
```diff
@@ -39,6 +40,7 @@ export const _AreaUploadManager = () => {
       <AreaUploadManager
         {...{ selectFilesText, title, fileIcons, description, allowedFileTypes, addMoreFilesText }}
         onChangeCallback={inputCallback}
+        clearState={clearState}
```
fti-bcarpenter, line 43: This should be `clearFiles`. Isn't it?

fti-bcarpenter (changes requested):

fti-bcarpenter (approved):
LGTM

atomworks (changes requested):
Could we move the button to clear this in to the AreaUploadManager component itself? I think we can just show it all the time as it's useful. 

Just one of the grey buttons, small size and centered in the container would be great.

fti-rchaniya:
> Could we move the button to clear this in to the AreaUploadManager component itself? I think we can just show it all the time as it's useful.
> 
> Just one of the grey buttons, small size and centered in the container would be great.

Yes, we can. Can you please confirm me the place where we can place that button to clear files

fti-rchaniya:
> Could we move the button to clear this in to the AreaUploadManager component itself? I think we can just show it all the time as it's useful.
> 
> Just one of the grey buttons, small size and centered in the container would be great.

Thanks for the feedback, suggested changes have been implemented

atomworks (approved):

PR #408, opened 2023-05-09
FilterDropdown -> Truncating filter options into a single line Using CSS

fti-rchaniya:
### Description

This PR has following change in the FilterDropdown Ui component:

- As shown in the below attached screenshot, english language text is truncated to single(upto 28 characters) line by appending ..., but japanese text is truncated in 2 lines(that is also 28 characters), reason is below logic written to implement this:
**removeString.map((value: string) => ({ value: value, text: value !== null && value.length > 28 ? value.substring(0, 28) + '...' : value }));**

image

- In above logic we are truncating upto 28 characters by comparing its length, and that is the reason we can see difference in width for language english and japanese. Both taking 28 characters only but because of characters spacing is different in both the language, english characters comes in single line whereas japanse characters comes in 2 line.

**Solution for this is, achieve this by adding below css to the <Title /> of the FilterOption scorer-ui-kit component:**
const Title = styled.div`
  font-family: ${({ theme }) => theme.fontFamily.data};
  display: block;
  color: hsl(0, 0%, 56%);
  font-size: 14px;
  font-weight: 500;
  margin-left: 12px;
  user-select: none;
  white-space: nowrap;  
  overflow: hidden;  
  text-overflow: ellipsis;  
  max-width: 210px;
`;
In above solution, characters get truncate upto width 210px
 
**Screenshot of the expected ui :**
image

image

### DIFF
--- a/src/Form/atoms/FilterOption.tsx
+++ b/src/Form/atoms/FilterOption.tsx
@@ -1,4 +1,4 @@
-import React, {useState, useRef} from 'react';
+import React, {useRef} from 'react';
 import styled, { css } from 'styled-components';
 import { CheckMark } from '../../svg';
 import { IInputOptionsType } from '..';
@@ -160,12 +160,11 @@ const FilterOption: React.FC<IFilterOption> = ({
   const iconWeight: number = dimensions.icons.weights['regular'];
 
   const titleRef = useRef(null);
-  const [isOverflowing, setIsOverflowing] = useState(false);
 
   const handleMouseEnter = () => {
     const titleElement = titleRef.current as unknown as HTMLDivElement;
     if (titleElement) {
-      setIsOverflowing(titleElement.scrollWidth > titleElement.clientWidth);
+      titleElement.title = titleElement.scrollWidth > titleElement.clientWidth ? title : '';
     }
   };
 
@@ -183,7 +182,7 @@ const FilterOption: React.FC<IFilterOption> = ({
         </FakeCheckbox>
       )}
       {(optionType === 'radio') && <FakeRadioButton><FakeInnerRadio /></FakeRadioButton>}
-      <Title ref={titleRef} title={isOverflowing ? title : undefined} onMouseEnter={handleMouseEnter}>{title}</Title>
+      <Title ref={titleRef} onMouseEnter={handleMouseEnter}>{title}</Title>
     </Container>
   );
 };
--- a/src/Form/atoms/FilterOption.tsx
+++ b/src/Form/atoms/FilterOption.tsx
@@ -136,7 +136,7 @@ const Container = styled.div<{ disabled: boolean, selected: boolean }>`
 `;
 
 interface IFilterOption {
-  title: string
+  title: string | null
   value?: string | number
   optionType?: IInputOptionsType
   selected?: boolean
--- a/src/Form/atoms/FilterOption.tsx
+++ b/src/Form/atoms/FilterOption.tsx
@@ -12,7 +12,10 @@ const Title = styled.div`
   font-weight: 500;
   margin-left: 12px;
   user-select: none;
-  pointer-events: none;
+  white-space: nowrap;  
+  overflow: hidden;  
+  text-overflow: ellipsis;  
+  max-width: 210px;
 `;
 
 const FakeCheckbox = styled.div`
@@ -170,7 +173,7 @@ const FilterOption: React.FC<IFilterOption> = ({
         </FakeCheckbox>
       )}
       {(optionType === 'radio') && <FakeRadioButton><FakeInnerRadio /></FakeRadioButton>}
-      <Title>{title}</Title>
+      <Title title={title}>{title}</Title>
     </Container>
   );
 };
--- a/src/Form/atoms/FilterOption.tsx
+++ b/src/Form/atoms/FilterOption.tsx
@@ -136,7 +136,7 @@ const Container = styled.div<{ disabled: boolean, selected: boolean }>`
 `;
 
 interface IFilterOption {
-  title: string | null
+  title: string
   value?: string | number
   optionType?: IInputOptionsType
   selected?: boolean
--- a/src/Form/atoms/FilterOption.tsx
+++ b/src/Form/atoms/FilterOption.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useState, useRef} from 'react';
 import styled, { css } from 'styled-components';
 import { CheckMark } from '../../svg';
 import { IInputOptionsType } from '..';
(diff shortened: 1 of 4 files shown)


fti-skumar (commented):
I think we don't need these changes, if we handle text-truncate on ui-kit's component itself, rather than handling it on projects.
In that way we don't have to write additional handling on projects each time we use `FilterDropdown` component.
@fti-bcarpenter sir, please suggest on this.

fti-smaske (approved):

fti-smaske (approved):

fti-rchaniya:
> I think we don't need these changes, if we handle text-truncate on ui-kit's component itself, rather than handling it on projects. In that way we don't have to write additional handling on projects each time we use `FilterDropdown` component. @fti-bcarpenter sir, please suggest on this.



> I think we don't need these changes, if we handle text-truncate on ui-kit's component itself, rather than handling it on projects. In that way we don't have to write additional handling on projects each time we use `FilterDropdown` component. @fti-bcarpenter sir, please suggest on this.

@fti-skumar , yes we can do in that way.
But even if we handle text-truncate on ui-kit's component itself, still we need these changes. Because instead of string there will be HTMLElement assigned to the text attribute.

Only change will be happening is as below:
We will add Text-Truncate CSS at the ui-kit's component and instead of checking below condition  every where:
**let text = typeof item.text === 'string' ? item.text : item.text instanceof HTMLElement ? item.text?.textContent : '';**

We can simply, assign **item.text?.textContent** to the **text** key

@fti-bcarpenter kindly suggest above opinion.

fti-skumar:
> Because instead of string there will be HTMLElement assigned to the text attribute.

@fti-rchaniya 
Why ? Any specific reason for assigning HTMLElement, do we need to add any other feature to those texts ?

fti-rchaniya:
> @fti-rchaniya
> 
> > Because instead of string there will be HTMLElement assigned to the text attribute.
> 
> Why ? Any specific reason for assigning HTMLElement, do we need to add any other feature to those texts ?

Then how you will be truncate text without html element???

fti-skumar:
> Then how you will truncate text without html element???

You just need apply CSS on Text-div

Here is the solution

- Remove all changes
- Goto this [file](https://github.com/future-standard/scorer-ui-kit/blob/main/packages/ui-lib/src/Form/atoms/FilterOption.tsx)
- Apply ellipsis style on `<Title />` component.
- Also add `title` prop on `<Title />` component so that on hover we can get the tooltip

Let me know if you have any doubt on this.

fti-sbondar:
> white-space: nowrap; overflow: hidden; text-overflow: ellipsis; max-width: 220px;;

No, I'm clear about your given solution and that might work perfectly without adding this much of code. I think we can add text overflow css to the **Title** of the ui-kit's  FilterDropdown itself.

Just need confirmation from @fti-bcarpenter Sir

fti-rchaniya:
> > Then how you will truncate text without html element???
> 
> You just need apply CSS on Text-div
> 
> Here is the solution
> 
> * Remove all changes
> * Goto this [file](https://github.com/future-standard/scorer-ui-kit/blob/main/packages/ui-lib/src/Form/atoms/FilterOption.tsx)
> * Apply ellipsis style on `<Title />` component.
> * Also add `title` prop on `<Title />` component so that on hover we can get the tooltip
> 
> Let me know if you have any doubt on this.

No, I don't have any doubt. That will be work, we can simply add text overflow CSS to the **Title** of the ui-kit's FilterDropdown, in that way we can minimize implementation.
Just need confirmation from @fti-bcarpenter sir.

fti-rchaniya:
> > > Then how you will truncate text without html element???
> > 
> > 
> > You just need apply CSS on Text-div
> > Here is the solution
> > 
> > * Remove all changes
> > * Goto this [file](https://github.com/future-standard/scorer-ui-kit/blob/main/packages/ui-lib/src/Form/atoms/FilterOption.tsx)
> > * Apply ellipsis style on `<Title />` component.
> > * Also add `title` prop on `<Title />` component so that on hover we can get the tooltip
> > 
> > Let me know if you have any doubt on this.
> 
> No, I don't have any doubt. That will be work, we can simply add text overflow CSS to the **Title** of the ui-kit's FilterDropdown, in that way we can minimize implementation. Just need confirmation from @fti-bcarpenter sir.

Suggested implementation from @fti-skumar is confirmed from @fti-bcarpenter sir, and Updated PR. Done

fti-skumar (approved):

src/Form/atoms/FilterOption.tsx
```diff
@@ -170,7 +173,7 @@ const FilterOption: React.FC<IFilterOption> = ({
         </FakeCheckbox>
       )}
       {(optionType === 'radio') && <FakeRadioButton><FakeInnerRadio /></FakeRadioButton>}
-      <Title>{title}</Title>
+      <Title title={title}>{title}</Title>
```
fti-bcarpenter, line 176: In general showing title in FilterDropdown is not a good idea, but if any reason we need to show it then it should be conditional and should only shown for the options which are doesn't fit in the dropdown width.

**pointer-events: none;** and **title** should be conditional.

Do not show **title** for the FilterOption which are fit perfectly in dropdown.

Some code reference are available here
https://codepen.io/triple-j/pen/wadKQB
https://dirask.com/posts/JavaScript-check-if-text-overflow-ellipsis-is-active-1XB5Yp

fti-bcarpenter (changes requested):

src/Form/atoms/FilterOption.tsx
```diff
@@ -156,6 +159,16 @@ const FilterOption: React.FC<IFilterOption> = ({
 
   const iconWeight: number = dimensions.icons.weights['regular'];
 
+  const titleRef = useRef(null);
+  const [isOverflowing, setIsOverflowing] = useState(false);
+
+  const handleMouseEnter = () => {
+    const titleElement = titleRef.current as unknown as HTMLDivElement;
+    if (titleElement) {
+      setIsOverflowing(titleElement.scrollWidth > titleElement.clientWidth);
```
fti-bcarpenter, line 168: We can avoid using state and re-rendering with below code
```
if (titleElement) {
   titleElement.title=title;
}
```

fti-bcarpenter (changes requested):

src/Form/atoms/FilterOption.tsx
fti-rchaniya: Done, Thanks for feedback

isacoder (changes requested):
Hello!

Thanks for making the update.
I have run this on local.
The examples showed in the description are not implemented. 
My first request will be to have examples in the story.

After adding examples, I was not able to reproduce the hover functionality with the tooltip. I don't see the tooltip code, was this removed?

This PR is a bit confusing to review please update and ask the review again when ready.


Screenshot 2024-01-22 at 15 06 39

Screenshot 2024-01-22 at 15 06 45

atomworks:
This PR looks to be abandoned, so I am closing it. Please re-open if / when required and you are able to complete the work. Thank you.

PR #409, opened 2023-05-10
Filter dropdown should show vertical scroll if it has more item which are not visible in the list

fti-rchaniya:
### Description

This PR has following change in the FilterDropdown Ui component:

- As shown in the below attached screenshot, there are more items to scroll down, even though there is slight gradient at the bottom to notify user that there are more items but still it feels that there must be a vertical scroll bar in the FilterDropdown if items are more to scroll down
image
image

- To achieve this, I have added **overflow-y** property in the CSS conditionally as shown below -
**overflow-y: ${props => (props.hideVerticleScrollbar ? 'hidden' : 'scroll')};
<OptionList hideVerticleScrollbar={list.length < 5 || list.length === 5}>**

- In above code, if the items are greater than or equal to 5 then vertical scroll bar should be **hidden** else it should be **scrollable**.

Here are some screenshots of the expected result.
image
image
image

### DIFF
--- a/src/Filters/molecules/FilterDropdown.tsx
+++ b/src/Filters/molecules/FilterDropdown.tsx
@@ -34,16 +34,11 @@ const StyledFilterOption = styled(FilterOption)`
   letter-spacing: 0.2px;
 `;
 
-const OptionList = styled.div`
+const OptionList = styled.div<{hideVerticleScrollbar: boolean}>`
   max-height: 162px;
   min-height: 40px;
   position: relative;
-  overflow-y: scroll;
-  ::-webkit-scrollbar {  /* Hide scrollbar for Chrome, Safari and Opera */
-    display: none;
-  }
-  -ms-overflow-style: none;  /* IE and Edge */
-  scrollbar-width: none; /* Firefox */
+  overflow-y: ${props => (props.hideVerticleScrollbar ? 'hidden' : 'scroll')};
 
   ${StyledFilterOption} {
     height: 35px;
@@ -339,7 +334,7 @@ const FilterDropdown: React.FC<IFilterDropdown> = ({
             : (
               <ResultsContainer>
                 {hasOptionsFilter && <ResultCounter>{getResultText(searchResultText, visibleList.length, list.length)}</ResultCounter>}
-                <OptionList>
+                <OptionList hideVerticleScrollbar={list.length < 5 || list.length === 5}>
                   {(visibleList.length > 0)
 
                     ? visibleList.map((item: IFilterItem, index) => {


src/Filters/molecules/FilterDropdown.tsx
```diff
@@ -339,7 +334,7 @@ const FilterDropdown: React.FC<IFilterDropdown> = ({
             : (
               <ResultsContainer>
                 {hasOptionsFilter && <ResultCounter>{getResultText(searchResultText, visibleList.length, list.length)}</ResultCounter>}
-                <OptionList>
+                <OptionList hideVerticleScrollbar={list.length < 5 || list.length === 5}>
```
fti-skumar, line 337: I think its not needed, we have max-height in OptionList and that should be the only criteria for showing scroll-bar, let css decide whether to display scroll or not based upon max-height, you can use `overflow: auto;`.

fti-skumar (changes requested):

src/Filters/molecules/FilterDropdown.tsx
fti-skumar: Please also style the scroll-bar to match ui-kit's styling.
Here is one example which you can refer.
image

src/Filters/molecules/FilterDropdown.tsx
```diff
@@ -38,12 +38,22 @@ const OptionList = styled.div`
...
+  ::-webkit-scrollbar {
+    width: 6px;
+    height: 34%;
+  }
+
+  ::-webkit-scrollbar-thumb {
+    border-radius: 10px;
+    background: #44444487;
+    opacity: 0.5;
+    transform: translate(0px, 0px);
```
fti-skumar, line 54: This two properties seems to be ineffective, and not needed.

src/Filters/molecules/FilterDropdown.tsx
```diff
@@ -38,12 +38,22 @@ const OptionList = styled.div`
...
+    transition: background-color .3s;
```
fti-skumar, line 55: I think you want to apply some hover effect on **background** as you have used `transition`, but there is no such feature applied, so I think if you don't need it you can remove it.

fti-skumar (changes requested):

fti-skumar (approved):

src/Filters/molecules/FilterDropdown.tsx
```diff
@@ -38,12 +38,19 @@ const OptionList = styled.div`
...
+  /* Customize the scrollbar */
+  ::-webkit-scrollbar {
+    width: 6px;
+    height: 34%;
+  }
+
+  ::-webkit-scrollbar-thumb {
+    border-radius: 10px;
+    background: #44444487;
   }
```
fti-bcarpenter, line 53: Use thin scrollbar

fti-bcarpenter (changes requested):

src/Filters/molecules/FilterDropdown.tsx
```diff
@@ -38,12 +38,19 @@ const OptionList = styled.div`
...
-  overflow-y: scroll;
-  ::-webkit-scrollbar {  /* Hide scrollbar for Chrome, Safari and Opera */
-    display: none;
+  overflow-y: auto;
+  margin-right: 5px;
+
+  /* Customize the scrollbar */
+  ::-webkit-scrollbar {
+    width: 6px;
+    height: 34%;
```
fti-bcarpenter, line 47: IMO height is not needed. Please check.

src/Filters/molecules/FilterDropdown.tsx
fti-rchaniya: Done

src/Filters/molecules/FilterDropdown.tsx
fti-rchaniya: Done

src/Filters/molecules/FilterDropdown.tsx
```diff
@@ -34,22 +34,28 @@ const StyledFilterOption = styled(FilterOption)`
...
   position: relative;
   overflow-y: auto;
-  margin-right: 5px;
+  
+  /* Firefox supports the scrollbar-width property */
+  scrollbar-width: thin;
 
-  /* Customize the scrollbar */
+  /* Apply margin-right to scrollbar for Chrome and Safari */
+  ${({ isChromeOrSafari }) => isChromeOrSafari && `
```
fti-bcarpenter, line 47: It is ok to have 2px margin for all browser to avoid this condition.

fti-bcarpenter (changes requested):

src/Filters/molecules/FilterDropdown.tsx
fti-rchaniya: For Firefox, it is by default taking some right margin
to maintain uniformity i have added right margin to chrome and safari
Please check below attached screenshot If I remove right margin 
image
image

If I'll add right margin without any browser specific condition, then extra right margin will be added to the firefox, will it be okay?????
please check below attached screenshot

image
image

fti-rchaniya: Done

src/Filters/molecules/FilterDropdown.tsx
```diff
@@ -346,7 +353,7 @@ const FilterDropdown: React.FC<IFilterDropdown> = ({
             : (
               <ResultsContainer>
                 {hasOptionsFilter && <ResultCounter>{getResultText(searchResultText, visibleList.length, list.length)}</ResultCounter>}
-                <OptionList>
+                <OptionList isChromeOrSafari={isChromeOrSafari}>
```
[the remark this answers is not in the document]
fti-rchaniya: Done

fti-bcarpenter (approved):
LGTM

atomworks (approved):

PR #413, opened 2023-07-17
Add Callback and invocation for menu open close value.

fti-smaske:
### Description

This PR has code for getting value true or false wether menu is open or closed. 
Here onMenuToggle callback is used to pass value to parent component. If menu is closed it will pass false other wise true,  

This feature is required in Platform renovation project to make organisation switch which is placed on top bar making its position abosulte near side menu. Problem arrived when user open menu or hover on menu the switch hide behind the menu. to overcome this problem code need callback to get menu open/closed value.

Here are some screenshots of the expected result.

**1. When menu is open its passed **true** value**

image

**2. When Menu is closed its passed false value**

image

### DIFF
--- a/src/Global/index.ts
+++ b/src/Global/index.ts
@@ -61,6 +61,7 @@ export interface IMenuTop {
     autoHideText? :string
     defaultMenuOpen?: boolean
     canAlwaysPin?: boolean
+    onMenuToggle?: (isMenuOpen: boolean) => void
   }
 
   export interface ICustomDrawer {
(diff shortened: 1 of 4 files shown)


JoshLipps (changes requested):
See PR with changes that should be back ported if you want this on the 1.x branch.
https://github.com/future-standard/scorer-ui-kit/pull/420

atomworks:
As this PR seems to be abandoned, I am closing it.

PR #414, opened 2023-07-17
FilterDropDown -> Graded last filter option was not selectable

fti-rchaniya:
**Description**

This PR has following fix in the FilterDropDown component:

- With respect to the below referenced PR, when we fixed last item cropped issue for the FilterDropdown component, we added a gradient css property to make last item faded which was achieved by below css :
**const Gradient = styled.div`
  position: absolute;
  right: 0px;
  bottom: 9px;
  height: 25px;
  background-image: linear-gradient(to bottom, rgba(246, 247, 249, 0) 1%, #F6F7F9 120%);
  width: 100%;
`;**

- Due to above property on last item which consist **position: absolute;**, user was not able to select the last item.
- To fix above issue, I have added one css property called **pointer-events: none;** 
- Now, a user can select the last graded option from the list

[Reference](https://github.com/future-standard/scorer-ui-kit/pull/396)

**Implemented Ui Snapshot**
image

### DIFF
--- a/src/Filters/molecules/FilterDropdown.tsx
+++ b/src/Filters/molecules/FilterDropdown.tsx
@@ -97,6 +97,7 @@ const Gradient = styled.div`
   height: 25px;
   background-image: linear-gradient(to bottom, rgba(246, 247, 249, 0) 1%, #F6F7F9 120%);
   width: 100%;
+  pointer-events: none;
 `;
 
 const isValueSelected = (item: IFilterItem, selected: IFilterValue) => {


fti-skumar (changes requested):

fti-skumar:
I think we should hide gradient-element when we scroll to the bottom. Please also explore this, you need to compare dropdown element height and scroll position.

fti-rchaniya:
> I think we should hide gradient-element when we scroll to the bottom. Please also explore this, you need to compare dropdown element height and scroll position.

Done

fti-skumar:
@fti-rchaniya 
Even though I scrolled to the bottom, the gradient is not removed or hide.

image

fti-skumar:
Also please check this github warning.

image

fti-rchaniya:
> @fti-rchaniya Even though I scrolled to the bottom, the gradient is not removed or hide.
> 
> image

@fti-skumar 
If you will scroll in between middle then obviously you will see that gradient, if you will be scrolling fullest to the bottom then it will definetly hiding. Please verify again and also I have attached below video for your reference

https://github.com/future-standard/scorer-ui-kit/assets/118800413/4c0e04d2-42cd-4193-8d89-387c72f1fc7e

fti-rchaniya:
> Also please check this github warning.
> 
> image

Resolved

fti-skumar:
@fti-rchaniya ,

> If you will scroll in between middle then obviously you will see that gradient, if you will be scrolling fullest to the bottom then it will definetly hiding. Please verify again and also I have attached below video for your reference

Sorry, I didn't have the latest code that's why its not hiding.

fti-skumar:
@fti-rchaniya ,

When we have only 5 elements in drop down then its container max-height is not enough to cover all elements, I think max-height can be **175px** currently its **162px,** can you please check and use appropriate height.

image

fti-skumar (approved):
Thanks for the changes, it looks good to me

fti-bcarpenter (approved):
LGTM

atomworks (approved):

PR #415, opened 2023-07-26
ICustomDrawer -> modified interface type of counter to accept string or number 

fti-rchaniya:
### Description
This PR consist a following change in the ICustomDrawer interface-

1. In TopBar component, when the user showing a count of the notifications there might be possibility that count length gets increase to 4 or more as shown in below attached snap-
image

2. In order to get rid of such behaviour, we decided that if count exceeds **999** then we can show it as **999+** , as shown in below attached snap-
image

3. To achieve above, we need to concatenate a number with **+** sign, which will return a string 
4. But as **counter** in **ICustomDrawer** accepts number only it throws an error
5. Hence, we modified type of **counter** in **ICustomDrawer** as below:
**counter?: number | string**

Kindly go through below BH requirement link in order to get better clarity 

[BH Requirement](https://www.bugherd.com/projects/263632/tasks/325)

### DIFF
--- a/src/Icons/StatusIcon.tsx
+++ b/src/Icons/StatusIcon.tsx
@@ -12,7 +12,7 @@ const StatusCounter = styled.div<{ color?: IStatusDot }>`
   position: absolute;
   left: 14px;;
   top: -12px;
-  border-radius: 50%;
+  border-radius: 4px;
   height: 14px;
   min-width: 14px;
   padding: 2px;
@@ -45,14 +45,15 @@ interface IStatusIcon {
   icon: string
   status?: IStatusDot
   counter?: number
+  maxCounter: number
 }
 
-const StatusIcon: React.FC<IStatusIcon> = ({icon, status, counter}) => {
+const StatusIcon: React.FC<IStatusIcon> = ({icon, status, counter, maxCounter = 999}) => {
   return (
     <Container>
       {status && (counter === undefined)
         ? <StatusDot color={status} />
-        :  (counter === undefined) ? null : <StatusCounter color={status}>{counter}</StatusCounter>}
+        :  (counter === undefined) ? null : <StatusCounter color={status}>{counter > maxCounter ? `${maxCounter}+` : counter}</StatusCounter>}
       <Icon icon={icon} size={18} color='dimmed' />
     </Container>
   );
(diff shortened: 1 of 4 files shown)


fti-skumar (approved):

fti-bcarpenter (approved):
LGTM

atomworks (changes requested):
It would be better to add a new prop called `maxCounter` and have that as a default of `999` and then use some logic internally to set the `+` suffix behaviour.

As an additional request, would we change the background that is behind the number to have a border radius of 4px? At the moment, the intended circle shape doesn't work on longer values as it stretches so a fix bored radius would look cleaner. Thank you.

fti-rchaniya:
> It would be better to add a new prop called `maxCounter` and have that as a default of `999` and then use some logic internally to set the `+` suffix behaviour.
> 
> As an additional request, would we change the background that is behind the number to have a border radius of 4px? At the moment, the intended circle shape doesn't work on longer values as it stretches so a fix bored radius would look cleaner. Thank you.

Done
Below is the snap for ui change.
image

fti-skumar (approved):

storybook/src/stories/Global/GlobalUI.stories.tsx
```diff
@@ -367,7 +367,8 @@ const customDrawer: ICustomDrawer = {
   customComponent: MyCustomDrawer,
   icon: 'Add',
   status: 'danger',
-  // counter: 5,
+  counter: 1001,
+  // maxCounter: 1000,
```
fti-bcarpenter, line 371: Remove this if not needed.

src/Icons/StatusIcon.tsx
```diff
@@ -12,7 +12,7 @@ const StatusCounter = styled.div<{ color?: IStatusDot }>`
   position: absolute;
   left: 14px;;
   top: -12px;
-  border-radius: 50%;
+  border-radius: 4px;
```
fti-bcarpenter, line 15: Attach the snapshot with border-radius 4px in this PR.

fti-bcarpenter (changes requested):

storybook/src/stories/Global/GlobalUI.stories.tsx
fti-rchaniya: Done

fti-bcarpenter (approved):
LGTM

atomworks (approved):

PR #424, opened 2023-10-03
Copy Icon in userDrawerMeta component handled conditionally

fti-rchaniya:
**Description**
This PR has following enhancement in the **UserDrawerMeta** component:

1. Copy Icon showing inside **UserDrawerMeta** component in the **TopBar** ui-kit's component is present based on value of the prop **hasCopyIcon**. If **hasCopyIcon** is true then Copy Icon will be present in the component, other it will be hidden.
image

2. Based on below BH requierment, it was required to show that copy icon only when user hover over the **UserDrawerMeta**, to achieve this, we have added one more prop named **hasHoverCopyIcon**.

3. When value of **hasHoverCopyIcon** is true then the copy icon will be visible only when user hover over the **UserDrawerMeta**. If it is false, then the copy icon will be visible always irrespective of onHover event.

[BH Requirement](https://www.bugherd.com/projects/289019/tasks/278)

**Here is screencast of the expected result.**

https://github.com/future-standard/scorer-ui-kit/assets/118800413/64db1f40-c329-4533-b279-895e2ad464bc

### DIFF
--- a/src/Global/atoms/UserDrawerMeta.tsx
+++ b/src/Global/atoms/UserDrawerMeta.tsx
@@ -81,8 +81,15 @@ const TitleBox = styled.div`
   align-items: center;
 `;
 
-const IconBox = styled.div`
+const IconBox = styled.div<{hasHoverCopyIcon?: boolean}>`
   padding: 1px 5px 0 0;
+  ${({ hasHoverCopyIcon }) => hasHoverCopyIcon && css`
+    opacity: 0;
+    ${Container}:hover & {
+      opacity: 1;
+      cursor: pointer;
+    }
+  `}
 `;
 
 const CopyTextBox = styled.pre`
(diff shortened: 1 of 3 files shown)


fti-smaske (approved):

fti-pfarne (approved):
Looks Good

fti-bcarpenter (approved):
LGTM

atomworks (changes requested):
If there was a request to make these visible only on hover of the meta box then I think it's best just to go with that as a single option for now. Having a split of the functionality here adds a little more complication than is necessary for the intentions of this component, thank you.

Thanks for adding a video to this PR by the way, it really helped me out!

fti-rchaniya:
> If there was a request to make these visible only on hover of the meta box then I think it's best just to go with that as a single option for now. Having a split of the functionality here adds a little more complication than is necessary for the intentions of this component, thank you.
> 
> Thanks for adding a video to this PR by the way, it really helped me out!

You're welcome! I completely understand your preference for simplicity, and I agree that keeping the visibility of these elements tied to the hover functionality would streamline the user experience. Please review updated version.
Also, I'm glad the video was helpful!
Below is the updated video attached for the same.

https://github.com/future-standard/scorer-ui-kit/assets/118800413/4a559f4c-9910-44b7-b682-93502902b7e6

atomworks (approved):

PR #425, opened 2023-10-09
Percentage Slider -> Added prop to handle ValueTitle show/hide behaviour

fti-rchaniya:
**Description**
This PR has following enhancement in the **PercentageSlider** component:

1. In our current component, we have Value title label showing at the top of the slider bar, as showing in the below attached snapshot.
image

2. There is a chances to make it's visibility show/hide based on the Boolean value passed by an user.

3. Hence, I have added a boolean prop named **showValueTitle** which will handle visibility of an element based on the value provided.

4. If true then label will be shown else it will be hidden.

### DIFF
--- a/storybook/src/stories/Form/Input/PercentageSlider.stories.tsx
+++ b/storybook/src/stories/Form/Input/PercentageSlider.stories.tsx
@@ -48,7 +48,7 @@ export const _PercentageSlider = () => {
   const customTitle = boolean("Custom Title function",false);
   const showValue = action('Input Callback');
   const marks = object('Marks', exampleMarks);
-  const showValueTitle = boolean("Value title visible", false);
+  const showValueTitle = boolean("Show Value Title", false);
 
   // const step = number('Step', 1); // still fixing step option
   const handleUpdate = (value: number) => {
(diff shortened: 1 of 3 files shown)


fti-smaske (approved):

fti-pfarne (approved):
Looks good

storybook/src/stories/Form/Input/PercentageSlider.stories.tsx
```diff
@@ -48,6 +48,7 @@ export const _PercentageSlider = () => {
   const customTitle = boolean("Custom Title function",false);
   const showValue = action('Input Callback');
   const marks = object('Marks', exampleMarks);
+  const showValueTitle = boolean("Value title visible", false);
```
fti-bcarpenter, line 51: "Value title visible" --> "Show Value Title"

fti-bcarpenter (changes requested):

storybook/src/stories/Form/Input/PercentageSlider.stories.tsx
fti-rchaniya: Done

fti-bcarpenter (approved):

atomworks (changes requested):
Looks good. Only small request, can we change `showValueTitle` to just `showValue`? Thanks!

fti-rchaniya:
> Looks good. Only small request, can we change `showValueTitle` to just `showValue`? Thanks!

Certainly! I've made the necessary changes as requested. Please review the updated version.

atomworks (approved):

PR #440, opened 2024-03-06
Update missing htmlfor, name, id missing in examples

isacoder:
### Description

There was a problem on a project that showed warnings in chrome, when reported it was mention that the same error was in our code. The solution was to update the missing parameters of id and htmlfor in labels. 

This PR corrects this problem in the examples to avoid the warnings and avoid confusion.

### Before
Two occurrences are shown because missing id on small input and one that is always present from Storybook code
   
Screenshot 2024-03-06 at 11 37 18

### After
After the update there is stilll the Storybook code shown but not in the small input
Screenshot 2024-03-06 at 11 34 54

### Description

Testing just check the components in Example and Storybook

### DIFF
--- a/example/src/pages/LinePage.tsx
+++ b/example/src/pages/LinePage.tsx
@@ -213,35 +213,35 @@ const Line: React.FC<{}> = () => {
         </SidebarBox>
         <SidebarBox>
           <Label labelText='Show Point Handle' htmlFor='showPointHandle' >
-            <input type='checkbox' name='showPointHandle' checked={options.showPointHandle} onChange={toggleOptions('showPointHandle')}/>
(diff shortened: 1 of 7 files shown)


atomworks (approved):
Thanks for that. This change looks good. 

I'm wondering if there's an issue in how we're handling this in our components that we can improve moving forward.

At the moment we're passing the same value to `id` and `name` which doesn't seem as tidy as it could be. I know `name` only has to be unique to the form and `id` to the whole page, so while it's common they'll be the same there might be a need to have them differ. I don't think we need to solve this here but I just wanted to bring it up, we should consider how we can improve this as we begin more breaking changes.

isacoder:
Current flexibility allows this warning to occur. I agree we might consider to update components that are wrapped with label to avoid this warning. Will follow up in a different PR.

PR #444, opened 2024-03-20
Fixes for DatePicker -  DateTimeBlock

isacoder:
### Description

This PR is a refactor of the DateTimeBlock feature of the Date Picker.
The expected improvements are:

- 24 hours fix,  the issue with  the current code base is that allows to have non zero minutes and the callback will return 23 hours. (24:04 -> 23:04)
- Updated the logic in  Date range validation that colors the Time with red for edge cases
- Rolling minutes (Minutes can increment hour when pressing keyup after 59,  or decrement when pressing key down after 00 unless clock is at 00:00 or 23:59)

In this refactor I used current main 2.0 as base,  I will list things that were removed from 1.0 for easier review but mostly removed the logic added from this PR [#392](https://github.com/future-standard/scorer-ui-kit/pull/392)

- Consolidated back Input for Date and Time since the difference between Input and InputValue was the border. Current logic also does not work with Date validation only Time range.
 
- Removed 00 logic because I found that updating the display with the date-fns library with the useEffect prevented missing 0

There are currently two PR's open on the DatePicker, I tried to make this PR to cover the requested scenarios, since my changes affects those, I will add the developer to the review this PR.

[DatePicker - 2 digit time format in the hours input field](https://github.com/future-standard/scorer-ui-kit/pull/405)
[DatePicker minute field roll over](https://github.com/future-standard/scorer-ui-kit/pull/404)
 

 ### Reviewing/Testing steps

Run Storybook on local and verify DatePicker by adding different values and changing interval option in Time and Date

I made a spreadsheet with the scenarios I used to test this PR feel free to comment on it to add some more if necessary
https://docs.google.com/spreadsheets/d/1POe9uZxKXtLhQFF6DIV-RclUp7T7oXgSeSLlbYmE38g/edit?usp=sharing

Here is an example video 

[Demo video](https://drive.google.com/file/d/1N7I34uShxWPNaYSLWLiq9L5KpDFzaoJM/view?usp=drive_link)

### DIFF
--- a/src/Filters/atoms/DateTimeBlock.tsx
+++ b/src/Filters/atoms/DateTimeBlock.tsx
@@ -37,7 +37,8 @@ const IconWrap = styled.div`
   padding-top: 2px;
 `;
 
-const InputValue = styled.input<{ readOnly? : boolean, allowManualTimeChange?:boolean }>`
+
+const Input = styled.input<{ readOnly? : boolean, isTimeRangeValid: boolean, isTimeInput?: boolean }>`
   ${({theme}) => css`
     font-family: ${theme.fontFamily.data};
   `}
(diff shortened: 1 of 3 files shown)


atomworks (approved):
Code looks good and the interactions feel really smooth in the browser when I tried it. Thanks for your hard work on this!

PR #457, opened 2024-07-09
Pagination Component

isacoder:
### Description

This is a Pagination component that has been used in many projects but we haven't integrate it to the UI Kit Yet. 

[Most resent design](https://app.zeplin.io/project/666a7d9f00bcfa44320386ed/screen/666a823b65115fce4f058ad3)

### Testing
Run Storybook on local and Review Pagination story

Some interaction I have reviewed

**Correct Typing Values**

- Insert correct value and update by enter
- Insert correct value and update by click
- Insert correct value and leave

**Invalid Typing Value**

- Insert more than the value of pages and try to enter
- Insert more than the value of pages and try to click
- Insert 0 and try to enter

**Tab Interaction**

- Tab into arrows and move one page up by right arrow
- Tab into arrows and move one page down by left arrow
- Tab into arrows and move left but is page 1 and its disabled
- Tab into arrows and move right but is page limit and it’s should be disabled

### Screenshot
[Interaction video](https://drive.google.com/file/d/1pIn10BirWxdnfjdcCpWngbCaCjWyiP7O/view?usp=sharing)

Screenshot 2024-07-10 at 13 56 14

### DIFF
--- a/packages/example/src/theme/theme-variables.css
+++ b/packages/example/src/theme/theme-variables.css
@@ -1,5 +1,7 @@
 :root {
   --font-ui: Monorale, Hiragino Sans, "ヒラギノ角ゴシック", Hiragino Kaku Gothic ProN, "ヒラギノ角ゴ ProN W3", Roboto, YuGothic, "游ゴシック", Meiryo, "メイリオ", sans-serif;
+  --font-data: Lato, Hiragino Sans, "ヒラギノ角ゴシック", Hiragino Kaku Gothic ProN, "ヒラギノ角ゴ ProN W3", Roboto, YuGothic, "游ゴシック", Meiryo, "メイリオ", sans-serif;
+
   --font-size: 14px;
 
   --button-font-size: 14px;
@@ -12,6 +14,8 @@
 
   --button-icon-size: 14px;
   --button-icon-h-padding: 8px;
+
+  --input-height: 40px;
 }
 
 .button-small {
@@ -368,6 +372,13 @@
   --spinner-primary: var(--primary-6);
   --spinner-simple: var(--grey-a8);
 
+  --input-border-default: var(--grey-7);
+  --input-border-default-focused: var(--primary-7);
+  --input-border-disabled: var(--grey-7);
+  --input-border-valid: var(--success-7);
+  --input-border-invalid: var(--warning-8);
+  --input-border-processing: var(--primary-7);
+
 }
 
 .dark-theme {
@@ -636,4 +647,11 @@
   --spinner-secondary: var(--grey-8);
   --spinner-primary: var(--primary-6);
   --spinner-simple: var(--grey-a8);
+
+  --input-border-default: var(--grey-7);
+  --input-border-default-focused: var(--primary-7);
+  --input-border-disabled: var(--grey-7);
+  --input-border-valid: var(--success-7);
+  --input-border-invalid: var(--warning-8);
+  --input-border-processing: var(--primary-7);
 }
--- a/packages/example/src/theme/theme-variables.css
+++ b/packages/example/src/theme/theme-variables.css
@@ -371,14 +371,6 @@
   --spinner-secondary: var(--grey-8);
   --spinner-primary: var(--primary-6);
   --spinner-simple: var(--grey-a8);
-
-  --input-border-default: var(--grey-7);
-  --input-border-default-focused: var(--primary-7);
-  --input-border-disabled: var(--grey-7);
-  --input-border-valid: var(--success-7);
-  --input-border-invalid: var(--warning-8);
-  --input-border-processing: var(--primary-7);
-
 }
 
 .dark-theme {
@@ -647,11 +639,4 @@
   --spinner-secondary: var(--grey-8);
   --spinner-primary: var(--primary-6);
   --spinner-simple: var(--grey-a8);
-
-  --input-border-default: var(--grey-7);
-  --input-border-default-focused: var(--primary-7);
-  --input-border-disabled: var(--grey-7);
-  --input-border-valid: var(--success-7);
-  --input-border-invalid: var(--warning-8);
-  --input-border-processing: var(--primary-7);
 }
--- a/packages/ui-lib/src/Form/atoms/SelectField.tsx
+++ b/packages/ui-lib/src/Form/atoms/SelectField.tsx
@@ -43,7 +43,7 @@ const StyledSelect = styled.select`
   }
 `;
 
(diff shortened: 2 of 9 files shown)


packages/ui-lib/src/Misc/molecules/Pagination.tsx
```diff
@@ -0,0 +1,351 @@
...
+
+      case 'invalid':
+        return 'var(--input-border-invalid, #f66)';
+
+      case 'default':
+      default:
+        return 'var(--input-border-default, --grey-9)';
+    }
+
+  }, []);
```
atomworks, line 270: I'd prefer to stick with just using the direct color references for text colours. If we need to have aliasing here, it shouldn't be linked with the border color aliases.

packages/ui-lib/src/Misc/molecules/Pagination.tsx
```diff
@@ -0,0 +1,351 @@
...
+      case 'processing':
+        return 'var(--input-border-processing, #4db5ff)';
+        break;
+
+      case 'invalid':
+        return 'var(--input-border-invalid, #f66)';
+
+      case 'default':
+      default:
+        return 'var(--input-border-default, --grey-9)';
```
atomworks, line 267: `grey-9` is suitable here but the `--input-border-default` is the wrong color for this and should be removed.

packages/ui-lib/src/Misc/molecules/Pagination.tsx
```diff
@@ -0,0 +1,351 @@
...
+    selectWidth = '60px',
+    selectDisabled = false,
+    itemsOptions = [],
+    onPageChange,
+    onItemsChange,
+  } = props;
+
+  const [fieldState, setFieldState] = useState<string>('default');
+  const [pageValue, setPageValue] = useState<string>(activePage.toString());
+  const [disableGO, setDisabledGo] = useState<boolean>(parseInt(pageValue) > totalPages && fieldState !== '' ? false : true);
```
atomworks, line 158: This is totally just me being picky, but can be update `disableGO` to `disableGo` for consistency? If you skip this I won't be mad but that case mismatch just jumps out at me 😄

packages/ui-lib/src/Misc/molecules/Pagination.tsx
```diff
@@ -0,0 +1,351 @@
...
+  20%, 40%, 60%, 80% { transform: translateX(5px); }
+  100% { transform: translateX(0); }
+`;
+
+const InputContainer = styled.div<{ borderColor?: string, shouldShake: boolean }>`
+  height: var(--input-height, 40px);
+  animation: ${({ shouldShake }) => (shouldShake ? shakeAnimation : 'none')} 150ms 2 linear;
+  -moz-animation: ${({ shouldShake }) => (shouldShake ? shakeAnimation : 'none')} 150ms 2 linear;
+  -webkit-animation: ${({ shouldShake }) => (shouldShake ? shakeAnimation : 'none')} 150ms 2 linear;
+  -o-animation: ${({ shouldShake }) => (shouldShake ? shakeAnimation : 'none')} 150ms 2 linear;
```
atomworks, line 74: I think we can drop browser prefixes here. I'm pretty sure compiler adds these for properties that still require it.

packages/ui-lib/src/Misc/molecules/Pagination.tsx
```diff
@@ -0,0 +1,351 @@
...
+        >
+          <Fragment>
+            {itemsOptions.map(({ value, textValue }, index) =>
+              <option key={index} value={value}>{textValue}</option>
+            )}
+          </Fragment>
+        </SelectField>
+      </ItemsSelectWrapper>
+      <PageLabel htmlFor='goButton'>{pageText}</PageLabel>
+      <InputContainer borderColor={getTextColor(fieldState)} shouldShake={shouldShake}>
```
atomworks, line 317: This is using `getTextColor` for `borderColor` which is why I think we've got some mix ups that I've commented on elsewhere. I think we either need to separate them or replace them with something else.

atomworks (changes requested):
Thanks for the work on this. I just have a few minor changes to request although if we need to, we can release this and then follow up afterwards.

There are some questions too but it's mostly related to either layout or code cleanliness.

packages/ui-lib/src/helpers/index.tsx
```diff
@@ -84,7 +84,7 @@ const uniqueID = () =>
 
   // https://stackoverflow.com/questions/30314447/how-do-you-test-for-nan-in-javascript
   const isNotNumber = (value: string) => {
-    const intValue = parseInt(value);
+    const intValue = Number(value);
```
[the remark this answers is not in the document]
isacoder: The stack overflow comment is still relevant, since the parse method is not covered of how to parse text but the comparison with itself to verify is not a number. The difference between previous implementation is that I didn't realized before that parseInt will do an effort to remove letters if there is at least one number, so parsing '11a' will be '11'. However Number() fuction is more strict and will perform better in this case where the input allows non numeric values. I only learned this because this input is not type number, it allows all text. 

Input type number shows the arrows like the datepicker. I could change this input to number only, will require extra css to hide the arrows though.

packages/ui-lib/src/Misc/molecules/Pagination.tsx
isacoder: Agreed

packages/ui-lib/src/Form/atoms/SelectField.tsx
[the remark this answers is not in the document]
isacoder: The situation here is that the Label  Component has margin-bottom.  This has been the standard for all the Form components. I'm not sure if we will continue to have this setting for all form components with the CSS external update . It was intended just for internal usage, but I understand that this opens an opportunity to be misused. I will do a workaround and remove the prop.

packages/example/src/theme/theme-variables.css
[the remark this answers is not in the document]
isacoder: I will just keep it in the ui-lib for now.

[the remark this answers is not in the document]
isacoder: Since this requirement is beyond the scope of this requirement and is not a one-liner change I will follow up in a different PR [Using Styled Components GlobalStyle for Variable theming ](https://github.com/future-standard/scorer-ui-kit/pull/458)

packages/ui-lib/src/helpers/index.tsx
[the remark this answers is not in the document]
atomworks: I feel like this function has some fundamental problem but at the same time, it makes enough sense and we haven't had any bugs reported from it.

Your explanation of the change though is really clear though, thank you for that. Looks good.

atomworks (approved):

PR #458, opened 2024-07-17
Using Styled Components GlobalStyle for Variable theming

isacoder:
### Description

This is a suggestion for handling the CSS variables by using GlobalStyles from Styled components

I have created an example page of Alerts that have a modification in the project from a GlobalStyled css and Local div wrapper that modifies the variables as well.

- This will cover the the requirement of having only one file for the CSS in the project.
- There is a conversation about installing Sass for internal usage, this will also will remove that requirement.
-  Lastly the another benefit will be to do seamless update of the variables without requiring the developer users to update their local variables css file.

 ### Reviewing/Testing steps
Run Example project and see the Custom Alerts Page

Screenshot 2024-07-17 at 14 28 19


packages/example/src/pages/LinksPage.tsx
```diff
@@ -17,6 +17,7 @@ const LinksPage : React.FC = () => {
     <div><Link to={`/login`}>Login</Link></div>
     <div><Link to={`/tabs`}>Tabs Example</Link></div>
     <div><Link to='/globalUI'>Global UI</Link></div>
+    <div><Link to={`/customalert`}>Custom Alert Page</Link></div>
```
atomworks, line 20: Might be better labelling this something like "component theme override example" just to let people know the main purpose here is to demonstrate that theme overriding.

atomworks (approved):
There's one small comment in there about the name of the example but other than that, happy to approve. Thanks!

PR #459, opened 2024-07-18
Icon color selection switched to CSS Variables

atomworks:
### Description

This updates icon color selections to use CSS variables instead of the older json based system. The older JSON stays in place for now so not to break project integration, however a deprecation warning has been added to the console.

From here, the color reference is the name of the color variable without the `--` prefix. For example, `grey-9` or `primary-6`.

- Updates all references done via `theme.colors.icons` to `var(--example-12)` css vars.
- Updates the `ISvgIcons` type so the color is simply a `string`.
- Added a deprecation `console.warn` message for use of the old references.

This work is required for the upcoming layouts feature but is being sent seperately for ease of review.

### DIFF
--- a/packages/ui-lib/src/Form/atoms/FilterOption.tsx
+++ b/packages/ui-lib/src/Form/atoms/FilterOption.tsx
@@ -47,7 +47,7 @@ const CheckMarkWrapper = styled.div`
       stroke: transparent;
     }
     [fill] {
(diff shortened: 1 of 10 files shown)


isacoder (approved):
Thank you for updating the theme variables, there are still a lot of theme to remove but will get there one file at the time 😄 

All seems working well, let's see how it's used in production.

PR #460, opened 2024-07-19
Page Header Iteration

isacoder:
### Description

This PR adds functionality to the Page Header component

 - Page title has different settings for subtitle and  icon color and position.
 - Page header can include tags and elements in the right section by sending a custom component.
 - Tags:  size can be set with tagSize prop as `compact` and `default` but still supports previous implementation with size. Tags now can have no border with noBorder. Default size is 12px instead of 14px to reflect the new design rework.
 - ButtonsBundle: New component with flex-wrap column of buttons and buttonWithIcon. 

https://www.figma.com/design/5pU39qhggEXCHM5uVl7Sr5/SCORER---UI-Kit-Library

- Match with layout update.

- Current implementation has left and right  sections aligned with flex but there is no wrap for smaller devices. This could be improved. I'm guessing the buttons would wrap bottom or if there is a better suggestion please advice.

 - ButtonsBundle: New component with flex-wrap column of buttons and buttonWithIcon. My intention was to add buttonWithLoading too but current style was misaligning the loading spinner since there was no scope for this task decided to not include it for now.

 ### Reviewing/Testing steps
Review Page Title and Page Header in storybook local or GlobalUIPage or Layouts in Example

### Screenshot
Screenshot 2024-08-01 at 15 27 28

[Storybook Interaction Video](https://drive.google.com/file/d/1RaQQ41S7DcF5VQhMMmopNluIT3flrDsw/view?usp=drive_link)


packages/ui-lib/src/Form/molecules/ButtonsBundle.tsx
```diff
@@ -0,0 +1,45 @@
...
+  gap: 8px;
+  button {
+    text-wrap: nowrap;
+  }
+`;
+
+export type IButtonType = 'default' | 'icon-button'
+
+export interface IBButton extends IButtonProps {
+  id: string
```
atomworks, line 19: I believe `ID` here should be optional rather than required.  

Looking at our components, it's not a mandatory prop for anything moving through the different levels. If I've missed something though, please let me know.

packages/ui-lib/src/Form/molecules/ButtonsBundle.tsx
```diff
@@ -0,0 +1,45 @@
...
+  icon?: string
+  iconPosition?: 'left' | 'right'
+  text: string
+}
+
+export interface IButtonsBundle {
+  buttons : IBButton[]
+}
+
+const ButtonsBundle : React.FC<IButtonsBundle>= ({buttons}) => {
```
atomworks, line 30: Would you be opposed to renaming this to `ButtonStack`? I feel that would be a slightly clearer name descriptively.

packages/ui-lib/src/Form/molecules/ButtonsBundle.tsx
```diff
@@ -0,0 +1,45 @@
...
+  flex-direction: column;
+  gap: 8px;
+  button {
+    text-wrap: nowrap;
+  }
+`;
+
+export type IButtonType = 'default' | 'icon-button'
+
+export interface IBButton extends IButtonProps {
```
atomworks, line 18: `IBButton` could we lengthen this name? It's a little confusing as this is Button Bundle but the B initial appears before Button and so looks a little like a typo.

(Maybe this changes anyway if we rename `ButtonsBundle`)

packages/ui-lib/src/Pages/molecules/PageHeader.tsx
```diff
@@ -1,28 +1,78 @@
...
   areaTitle?: string
   icon?: string
+  iconPosition?: IIconPos
   introductionText?: string
   updateDocTitle?: boolean
-  hideAreaInDocTitle? : boolean
+  hideAreaInDocTitle?: boolean
+  tagList?: IHeaderTag[]
+  areaTitleBottom?: boolean
+  customRight?: React.ReactNode | React.FC | ReactElement;
```
atomworks, line 45: For certain layouts previously I believe I've used `[place]Content` for this kind of prop. In the SplitLayouts I used the naming convention `[place]Area` as it's a more nested thing.

Could we unify the naming with one of these for consistency? Thank you

atomworks (changes requested):
Thanks for all your work on these. I've made handful of comments, the majority of them are in relation to small things like naming or general code layout (which are more optional for approval).

The main thing I think is relating to the size of the icon and how it is placed when above. This work made me realise that since updating some icon related components in the design side, the page header design hadn't been updated to match it. Sorry about this.

If you've got any questions, just drop me a message!

packages/ui-lib/src/Misc/atoms/Tag.tsx
```diff
@@ -47,43 +63,52 @@ export const TagWrapper = styled.div<{ hoverColor:ISvgIcons['color']; enableHove
   `};
 `;
 
+export type TypeTagSize = 'compact' | 'default'
```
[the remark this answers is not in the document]
isacoder: Well the third option that is not so obviously is "undefined".
The implemented logic cares for the scenario this  value is not sent to respect the previous versions where the size of the tag was the same as the size of text. 

Do you think it should go to legacy and only keep default and compact options?

packages/ui-lib/src/Misc/atoms/Tag.tsx
```diff
@@ -47,43 +63,52 @@ export const TagWrapper = styled.div<{ hoverColor:ISvgIcons['color']; enableHove
...
-    ?  (
-      <StyledLink to={linkTo}>
-        {renderTag()}
-      </StyledLink>)
-    : renderTag()
+      ? (
+        <StyledLink to={linkTo}>
+          {renderTag()}
+        </StyledLink>)
+      : renderTag()
```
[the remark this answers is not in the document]
isacoder: I usually think the ? and the : are the separators to pay attention in ternary operations. I don't mind updating it though. I will update.

packages/ui-lib/src/Form/molecules/ButtonsBundle.tsx
isacoder: The reason is to assign the key and  prevent the warning that React gives when you are rendering a list of items. 
https://react.dev/learn/rendering-lists. 

I want to delegate the control of the id's to the Dev user in case they need it.

packages/ui-lib/src/Pages/molecules/PageHeader.tsx
isacoder: I'll update to rightContent

packages/ui-lib/src/Pages/atoms/PageTitle.tsx
```diff
@@ -54,28 +70,34 @@ const AreaLinkTitle = styled(Link)`
...
   return (
-    <Container>
+    <Container areaTitleBottom={areaTitleBottom} >
       {areaTitle && areaHref
         ? <AreaLinkTitle to={areaHref}>{areaTitle}</AreaLinkTitle>
         : areaTitle && <AreaTitle>{areaTitle}</AreaTitle>}
       <Title>{title}</Title>
       {icon ?
-        <IconContainer><Icon size={24} color='dimmed' {...{icon}} /></IconContainer>
+        <IconContainer iconPosition={iconPosition}><Icon size={32} color={iconColor} {...{icon}} /></IconContainer>
```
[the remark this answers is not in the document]
isacoder: I have updated to 24 , it looks right with the changes of the design.

packages/ui-lib/src/Pages/atoms/PageTitle.tsx
```diff
@@ -5,15 +5,31 @@ import Icon from '../../Icons/Icon';
...
+    ` position: absolute;
+      top: 50%;
+      transform: translateY(-50%);
+      left: -45px;
+      bottom: 0;
+    `:
+    ` position: absolute;
+      top: -8px;
+      transform: translateY(-100%);
+      left: 0px;
```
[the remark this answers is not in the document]
isacoder: Sorry I did miss the fluid intention, thank you for pointing it out. I have fix it.

packages/ui-lib/src/Misc/atoms/Tag.tsx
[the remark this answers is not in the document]
atomworks: Thanks for the update on that, I appreciate it.

For me, I'm used to seeing that opening `(` and from there I'm looking down the same column for the closing `)` before I check for that `:`, so having it laid out like that makes it so much easier to spot for me. 🤓

packages/ui-lib/src/Form/molecules/ButtonsBundle.tsx
atomworks: Thank you for the link. I've got a few extra thoughts on this that I'd like to chat about quickly, I'll send you a message in Slack about it - likely though there won't need to be any other changes on here.

packages/ui-lib/src/Misc/atoms/Tag.tsx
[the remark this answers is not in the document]
atomworks: As we have the prop set with the `?` as `tagSize?: TypeTagSize`, I don't think we need the `undefined` here as it should be clear enough.

While I'd like to just switch it to a boolean as those are the only two size options planned, it seems that as it would be unnecessarily breaking so maybe we can circle round to it when get to a point where we can more actively break and refactor things.

atomworks (changes requested):
Thanks for this, it's coming along really well. 

I've got a couple of requests and

- [x] In the button stack, the button size is too big currently.
- [ ] When the icon in the header moves to the top position, the button stack should stay aligned with the title rather than the icon.

I've also made a follow up comment regarding just removing that `undefined` from the type as we have the `?` in the props.

And I'll send you a quick message regarding the button `id` bits as I want to add some context to my thinking. That said, I expect the code is good to go here.

packages/ui-lib/src/Form/molecules/ButtonsBundle.tsx
isacoder: Updated with key={id || `button-stack-${index}`} as we agreed in chat.

atomworks (approved):
👍

PR #464, opened 2024-07-29
Add missing props from Select in Pagination

isacoder:
### Description

There were 2 fixed values for id and defaultValue in the Select of the Pagination component.
Update of default Value is necessary for local storage of the user selection, this fix will enable that option.

 ### Reviewing/Testing steps

Run Pagination component in Storybook and test different values in the itemsDefaultValue

### Screenshots 
Screenshot 2024-07-30 at 8 42 26


atomworks (approved):
👍

PR #468, opened 2024-08-08
Fix to layout breaks on examples and stories.

atomworks:
Removes the flex from the `#root` selector which was breaking the layout of some examples and stories. It was added as part of a fix to the layouts not spanning the full height of the container and backgrounds being cut off when scrolling, however it seems that this flex on #root was redundant.

### DIFF
--- a/packages/example/src/style.ts
+++ b/packages/example/src/style.ts
@@ -4,17 +4,8 @@ import { BaseStyles } from 'scorer-ui-kit';
 const GlobalStyle = createGlobalStyle`
   ${BaseStyles};
 
-  // Custom changes for examples usage.
-  html, body, #root {
-    height: 100%;
-  }
-
   body {
     background: var(--grey-3);
-  }
-  
-  #root {
-    display: flex;
   }
 `
 export default GlobalStyle;
--- a/packages/example/src/style.ts
+++ b/packages/example/src/style.ts
@@ -3,5 +3,18 @@ import { BaseStyles } from 'scorer-ui-kit';
 
 const GlobalStyle = createGlobalStyle`
   ${BaseStyles};
+
+  // Custom changes for examples usage.
+  html, body, #root {
+    height: 100%;
+  }
+
+  body {
+    background: var(--grey-3);
+  }
+  
+  #root {
+    display: flex;
+  }
 `
 export default GlobalStyle;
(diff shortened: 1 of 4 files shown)


isacoder (changes requested):
Over all this fix looks better than the current version, but I see at least two examples that show visible differences. 

All the tables and components we talked about look better with this configuration but I see that the split layout is not full height in examples although it is storybook. 
The  PTZ example also shows the sidebar cropped. 

Some how I feel more confident in making a release with this fix than without it, but please take a look to those cases and let me know what you think.

I'm also adding some examples of the 3 configurations of layouts and how is affecting the behaviour of the examples. This are very particular cases of "fake sidebars", so maybe it's not that important but you have more experience with out layout system. Please take a look and share your thought on it.


This is the version of Line release/v2.1.1 shows the scroll bar

Screenshot 2024-08-09 at 10 49 03

The current version in main does not have it

https://future-standard.github.io/scorer-ui-kit/#/line
Screenshot 2024-08-09 at 11 01 41

The fix also doesn't show the scroll back.

Screenshot 2024-08-09 at 11 18 13


PTZ release/v2.1.1 

Screenshot 2024-08-09 at 11 09 05

PTZ current main
Screenshot 2024-08-09 at 11 22 25


PTZ PR Fix 

Screenshot 2024-08-09 at 11 15 39

atomworks:
This latest update does the following:
* Moves the body background to the `BaseStyles` helper (which is probably where it should have been already).
* Adds some CSS changes to the examples `styles.ts` to restore scrolling.
* Also removes the background gradient for this to avoid background bug.

As our UI work is mostly intended to work in our GlobalUI module, the example shows how you can set up your code if you're working outside of that context.

The background gradient was a nice to have for these but isn't essential for examples. To have this working properly here would take additional examples refactoring that I feel is time best spent elsewhere at the moment.

isacoder (approved):
This update partially fixes the changes in layout after the [SplitLayout update ](https://github.com/future-standard/scorer-ui-kit/pull/465)

There is still some fixing for the older projects related to the Sidebar in Examples
Full review is on this [document](https://docs.google.com/spreadsheets/d/1-PK3ptKepoM03gUauTNGgWJ7BmFogyZxqZMcPbwtsi4/edit?usp=sharing)

I'll be approving and merging this branch to improve current main but we will need to follow up with the pages that require fixing on a different PR.

PR #477, opened 2024-09-19
Filter Button Basic Design  Feature

isacoder:
### Description

New design for Filter button without border and no background.

This update assumes that the rest of the style in FilterDropdown stays the same and only the Button has transparent background color and border.

I considered having a boolean variable since there is only default and plain design but I though this could open possibilities to have other styles for the Filter.

Removed old theme variables in FilterButton.

 ### Reviewing/Testing steps

 ### Screenshot

Screenshot 2024-09-19 at 20 36 44

Screenshot 2024-09-19 at 20 36 52


atomworks (changes requested):
Changes look good. Only on thing I'd ideally like to revise...

At the moment, am I correct in thinking that the default style is the older one and that `plain` is the newer one? As the designs will use the newer pattern moving forward, it would be great to have that as the default and make the older one an optional `legacy` option.

isacoder:
> Changes look good. Only on thing I'd ideally like to revise...
> 
> At the moment, am I correct in thinking that the default style is the older one and that `plain` is the newer one? As the designs will use the newer pattern moving forward, it would be great to have that as the default and make the older one an optional `legacy` option.

I have updated 'Plain' design value to 'Basic' as we commented on slack.
This design is the value that will be used in SortButton as the base design and the "default" value will continue to be the default for all the other filterButtons.

atomworks (approved):

PR #480, opened 2024-09-25
Feature/refactor global components

atomworks:
The main work here refactors the top bar to the new slightly slimmer height and layout. However as part of this I updated some other global components and associated code. This includes a lot of work to port from legacy variables to new CSS based approach so we can retire the older styles.

I've gone through the changes and listed everything I am aware of having changed:
* TopBar height reduced to match newer design pattern.
* Background gradient revised to better match designs.
* Lato Italic font variation added (was previously used but missing)
* Fixed maxWidth on PageHeader
* Updated IntroductionText to match newer style.
* Revised GlobalUI example content in Storybook and Examples
* Moved various styles out of legacy theme, revising and updating to CSS vars where relevant
* Added simple animation to TopBar drawer buttons
* Updated MainMenu to use CSS vars

## Dangers / Breaking Changes
As some of the legacy variables have been removed, there is a slight risk that some of the projects that use the UI Kit might be referencing these. That said, most of those removed are fairly nested component styles and shouldn't be being used for anything.

For color variables, those should throw and easy to identify error and can be replaced easily. For that reason I would prefer to risk these changes causes mild disruption. I am open to hear reasoning against this.


isacoder (approved):
Although legacy variables were removed from the style file. I think there was a good update to follow up. I see nothing preventing this merge. 

Thank you for all this updates, I will merge now so you can continue with updates.

PR #481, opened 2024-09-27
Feature/tooltips (Ref based)

atomworks:
Exploring using `ref` based component interaction in place of `id` based. This updates  the main component and the example use in both the Storybook text hover code and the Example's table's status dots.

I experience one of the things you were warning about with the use of refs in the context such as the table example. I believe the solution I came to is suitable.

If you choose not to merge this, please feel free to close it (with some attention to the below). Thank you!

---

_As it was a minor update, I've pushed a commit to the main PR for those color updates. So hopefully the changes here will merge across with no impact or can be forgotten if this is rejected._

~~There is also a color update which should be updated in the `feature/tooltips` branch, even if this PR is rejected. It's easy to update manually as it's about 6 lines changed and just the css var name. This text in the background looking like it's above the tooltip arrow.~~

### DIFF
--- a/packages/example/src/components/StatusComponent.tsx
+++ b/packages/example/src/components/StatusComponent.tsx
@@ -36,12 +36,7 @@ interface IStatusBundle {
 export const StatusComponent: React.FC<IStatusBundle> = ({ statusList }) => {
 
   const statusRefs = useRef<(React.RefObject<HTMLDivElement>)[]>([]);
-
-  React.useEffect(() => {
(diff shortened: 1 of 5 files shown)


isacoder (approved):
My only concern is that the tooltip ref type definition tooltipFor: React.RefObject<HTMLElement>
Not sure of the limitations it might bring. 
I think it's better to review it after been used, we can always expand it.

Thanks for working in this update I will merge to my branch.

PR #482, opened 2024-10-01
Lato / Monorale font fixes

atomworks:
# Problem
There were a handful of font related issues. This included font-weights being reported wrongly from Zeplin and causing mistmatches; issues with missing italic font variations; and font letter spacing in the browser being misaligned with the equivelant spacing in the designs.

# Changes 
* Added a Fonts story to show variations (an help check in against designs)
* Swapped out Lato for a variable font variant, providing all variations in an efficient package.
* Changed `text-rendering` to `geometricPrecision` so that typography exactly matches weights in the design.
* Updated some global UI components to heavier font-weight

# Update Advisories
- When updating the UI kit, because font files need to be manually updated in the project there should be minimal impact to the existing font rendering in the project. However there may be minor difference in UI Kit components.
- Font weights should be checked in your project when updating as some many have become to thin or too bold. Usually this just needs a minor increment in font-weight and while it does take code updates, search and replace should provide an easy method to update.
- If you do not wish to update to the new fonts update in your project, you may wish to add a css override to set `text-rendering: optimizeLegibility` to help avoid fonts being too thin.


isacoder (approved):
Thank you for making this review of fonts. I will follow up in release notes with the steps to verify fonts on the projects.

PR #483, opened 2024-10-01
Updated Example landing page

atomworks:
This one was always a functional page but never got iterated. When viewing it on dark mode, the default link colors also became impossible to see when they had been visited before. So I decided it was time to give it a quick polish!

As part of this, I also renamed `LinksPage` to `Index` because it always took me longer than it should to find what file this was when editing it.

Lastly, the link to Storybook was setup mainly for Github pages. However it didn't work on localhost. As we have to run storybook as a seperate process locally anyway, I just hide this section it we're on localhost.

image


isacoder (approved):
Hello Leonard, All looks good to me 😄 

Thanks for updating the examples page.
It's very refreshing and adding information will make it useful for showcase.

PR #488, opened 2024-10-07
Button refactors

atomworks:
The main purpose of this is to update the buttons to the new designs, remove legacy variables and also ensure that older layout bugs are fixed. Here's a list of changes:

- Button colors have been updated.
- Sizes, including padding and text-sizes have been updated and is variable managed.
- A lot of variables new variables are introduced for button themes.
- New hover states and active states
- Icon button has a loading option which switches the icon for the spinner
- Removed legacy integration
- Drop shadows are added as an optional prop
- Stories are updated

Other fixes include:
* Easing in and out types were the wrong way around. Confirmed it and flipped the values and implementations of these.


isacoder (approved):
Thanks for this update! 

Buttons is a very large feature with many variables for design and size, and you have managed to add them all. Thank you for revising the animations. These can be tricky to get right, and I'm happy that you did it because you have a lot of experience with animations.

We already talked on Slack about the missing update for the Split button that will be added in different commits.

I have only one comment about the variables for shadows being in a not very understandable place. In general, I think the code in the Colors file is starting to be hard to manage. There are many separate `.light-theme{`} and `.dark-theme{} `instances, as well as global definitions, and the combined` .light-theme{}  .dark-theme{}` with different variables.

I hope we can consolidate this to improve clarity for anyone looking to add or modify variables in the future.

You mentioned previously that you were planning to create new files for buttons - perhaps we can consider doing that in a separate PR.

Since there are no major concerns in this update I will approve.

isacoder (approved):
Hello Leonard,

Thanks for fixing the  size class for the split buttons.
I will approve this PR we can follow up the comments and improvements of the split button in a different PR.

PR #495, opened 2024-10-15
Pagination activePage prop not updating value

isacoder:
### Description

 This is a fix request for the Pagination Component that was not updating value from property.
 
 The specific scenario requested is:
 
 ```
Suppose if user previously choose 10 items per page, and then goes to last page and thereafter he changes page size again to 20, in that case total pages count decreases but the current selected page does not change as shown in snapshot below and its false to show such case in UI so I think activePage prop is not bind with its passed value due to which we get this situation.
 ```
 
 
  ### Considerations in the implementation
I have updated the code to update  pageValue if activePage is updated.
Page Value is an temporary state to check the validity of input of the user.
activePage is been validated under the user input standards.

 Updated default value of total pages from 199 to 1, seems like a more reasonable default value 😄 
 
 ### Reviewing/Testing steps
 
 Storybook -> Misc- > Pagination
 New ActivePage Prop has been added to Pagination Story, I didn't find a way to update this knobs variable when the user updates the value by input :/ 

However testing should be performed manually by updating TotalPages to a lower Value then sending a new value of Active Page
 
 
Screenshot 2024-10-15 at 14 24 19

Screenshot 2024-10-15 at 14 51 05

Screenshot 2024-10-15 at 14 51 28

Screenshot 2024-10-15 at 14 51 45

### DIFF
--- a/packages/storybook/src/stories/Misc/molecules/Pagination.stories.tsx
+++ b/packages/storybook/src/stories/Misc/molecules/Pagination.stories.tsx
@@ -18,7 +18,7 @@ const ItemsOptions = [
 ]
 
 export const _Pagination = () => {
-  const controlledActivePage = number('Active Page', 21)
+  const controlledActivePage = number('Active Page', 1)
   const [currentPage, setCurrentPage] = useState(controlledActivePage);
   const buttonOnClick = action('Page changed');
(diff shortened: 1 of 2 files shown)


atomworks (approved):
👍

PR #499, opened 2024-10-24
Added Filename and a home button to examples

atomworks:
This feature adds two minor but useful improvements to Examples.

1. The index has a label on now with the filename of the source code it uses.
2. The individual examples also have a small label in the bottom center for this.
3. Next to this label in examples is a small home button to get back to the top level.

This is intended to help reduce friction when sharing examples and helping engineeers find the code to reference.

There is also a few minor changes to color and border on the index to help is be more readable in dark mode.

image
image


isacoder (approved):
Thanks for the update.
Everything looks good.

PR #500, opened 2024-10-24
Fixes: Checkbox, Radio and FilterButton color fixes

atomworks:
This fixes the following:

* Updates checkboxes and radio buttons to match the theme and using new CSS variables.
* Also updates the fake versions of these in the filter dropdowns.
* Removed unused CSS variables.
* Fixed small arrow icon hover color issue in the FilterButton
* Filter Dropdowns translucent background is a bit more opaque to better handle busy backgrounds


isacoder (approved):
Thanks for this rework, all looks right.

PR #503, opened 2024-10-29
Add disableCheckbox option to rows

isacoder:
### Description

This is an update request for TypeTable component.
Currently the checkboxes in the rows can't be disabled by developer side.
This update will allow to disable them while still checking the states for the allChecked Checkbox

I decided to use the logic if at lease one checkbox in rows is disable the checkAll/ unCheckAll will be disable too.
Attaching the reasoning of this logic
Screenshot 2024-10-29 at 14 42 57

I updated the example toggles to use previous as base prevent useEffect forever cycle when updating.

 ### Reviewing/Testing steps
 
Test on storybook `Tables -> Molecules -> LoadingTable`  local using disable 1st checkbox 

Screenshot 2024-10-29 at 14 47 33

### DIFF
--- a/packages/storybook/src/stories/Tables/molecules/LoadingTable.stories.tsx
+++ b/packages/storybook/src/stories/Tables/molecules/LoadingTable.stories.tsx
@@ -107,6 +107,7 @@ const dataRows : ITypeTableData = [
 export const _LoadingTable = () => {
   const isLoading = boolean("IsLoading", true);
   const emptyTable = boolean("Show Empty Table", true);
+  const disableOneCheckbox = boolean('Disable 1st Checkbox', false);
   const emptyTableTitle = text("emptyTableTitle","No Data Available");
   const emptyTableText = text("emptyTableText", 'There is currently no data');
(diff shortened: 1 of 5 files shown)


atomworks (approved):
Code looks good and works well in Storybook, thank you.

I would like to keep the (de)select all option active and just exclude anything that has been disabled. That said it would be useful to keep this as something the developer can actively choose disable for contexts where it makes more sense. Where this feature is needed though at the moment, it makes sense to still allow for select all.

That said, if you want to get this released first so it can be used in projects and follow up with that other change after, that seems like a good plan to me but I shall leave it up to your choice.

isacoder:
> (de)select all option active

Thanks for mentioning.
Currently the logic does not allow it easily. 
`Check All` is internally controlled, but individual checkboxes are controlled from outside. 
If I make that change It will require to give full responsibility of the `check All` logic to the developer and would be breaking changes for previous projects. 

I will merge and if any project requires this functionality we can review the scenario.

PR #505, opened 2024-10-29
Submenu height fix.

atomworks:
This completes the work from the hotfix of PR #504. It includes:

- Fixes to the mobile menu heights, removing the previous minHeight implementation and using a new `mobileMenu` prop to define the difference.
- Removes the `loading` state that was used to track when each `NavigationItem` had checked the height of its submenus. As it didn't appear to be used for anything else practical outside of these, it's been removed along with the ready callback which was part of this check.
- This has also been removed from the `mobileMenu`

Once this has been reviewed and you're happy with the menu functionality, we can deploy this. There is rush though as with the hotfix version of this.


isacoder (approved):
Thanks for the follow up update and the clean up, all looks good in mobile too.

PR #508, opened 2024-11-07
Datepicker  Date Range availability

isacoder:
### Description

This PR introduces the ability to restrict the selectable date range in the Date Picker component as requested.

```
The date picker needs two additional props:

- Upper selectable date limit
- Lower selectable date limit

`In the DatePicker UI this will restrict` the selectable date range with dates outside that range being un-selectable. This feature is only required for date and not for time ranges.

As a stretch goal, if there is time please also consider adding the following options:

- Only future dates selectable (today and onward)
*This would be for use cases such as scheduling a task.*
- Only past dates selectable (today and past)
*The use case for this is commonly viewing data logs.*
```

To enable this functionality, a new optional prop, `availableRange`, of type `DateRange`, has been added.

The `DateRange` interface is similar to `DateInterval`, as both include a start and end date. However, DateRange allows either start or end to be set to null, providing more flexibility for scenarios without strict date limits.

Only future dates selectable : Set the `start` date to today’s date and `end` to `null`.
Only past dates selectable: Set `start` to `null` and `end` to today’s date.

[Design](https://www.figma.com/design/5pU39qhggEXCHM5uVl7Sr5/SCORER---UI-Kit-Library?node-id=3271-28053&m=dev) updates from Figma

Disabled Previous and Next Month :  0.5 opacity  and show cursor: not-allowed;
Disabled Days: Color is grey-6 and show cursor: not-allowed, will not change background when hovered.

Disabled selected range updated

Screenshot 2024-11-12 at 15 32 21

 ### Reviewing/Testing steps

 Run storybook on local Filters -> Molecules -> DatePicker
 
 Check select the prop: Send Available Range
 Update prop:  Available Range
 
These scenarios to disable parts of the DatePicker component were considered and tested during implementation:
 
- Validate Months caring for different Years.
- Validate Previous Month
- Validate Next Month
- Validate days inside range

### DIFF
--- a/packages/storybook/src/stories/Filters/molecules/DatePicker.stories.tsx
+++ b/packages/storybook/src/stories/Filters/molecules/DatePicker.stories.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import styled from 'styled-components';
-import { object, select, text } from '@storybook/addon-knobs';
+import { boolean, object, select, text } from '@storybook/addon-knobs';
 import { action } from '@storybook/addon-actions';
 import { DatePicker, FilterDropdownContainer } from 'scorer-ui-kit';
 
@@ -29,14 +29,12 @@ start.setDate(0);
 start.setDate(start.getDate() - 20);
 const end: Date = new Date();
 end.setDate(1);
-end.setDate(end.getDate() + 20);
+end.setDate(end.getDate() + 70);
 
 const datesRange = {
   start: start,
   end: end
 }
-
-
 
 export const _DatePicker = () => {
   const language = select('Language', { English: 'en', Japanese: 'ja' }, 'ja');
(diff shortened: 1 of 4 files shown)


packages/storybook/src/stories/Filters/molecules/DatePicker.stories.tsx
```diff
@@ -24,29 +24,58 @@ const exampleCallback = <T extends Function>(fn: T): T => {
...
+  const initialValueObj = object('Initial Value', initialValue);
   const dateMode = select('Date Mode', { single: 'single', interval: 'interval' }, 'interval');
   const timeMode = select('Time Mode', { off: 'off', single: 'single', interval: 'interval' }, 'interval');
   const dateTimeTextUpper = text('Date Time Text Upper', 'From');
   const dateTimeTextLower = text('Date Time Text Lower', 'To');
   const timeZoneTitle = text('Time Zone Title', 'Timezone');
   const timeZoneValueTitle = text('Time Zone Value', 'JST');
   const updateCallback = action('Date / Time Updated');
+  const sendRange = boolean('Send Available Range', false);
+  const availableRangeDates = object('Available Range', datesRange);
```
atomworks, line 62: Can we enable this by default and make the default date range something that would be more visible without needing pagination? 

I think as one of our most common use cases is recently data, having a period that is 2 weeks before until today's date would be good if we're keeping the date dynamic.

atomworks (changes requested):
Looks great, thank you. Just one small request regarding the preset data in the story but otherwise looks like it's ready!

atomworks (approved):

PR #510, opened 2024-11-11
Add Toggle Button with Story

isacoder:
### Description

This PR is a request to have a button with 2 states with the  design of the filter button.

```
This is primarily for the use of alternating between grid and list views using the button in filter bar pattern. As it often just changes between one of two values, it is quicker instead to toggle instead of opening the dropdown.

We may wish to use this same button for other dual state options, such as showing annotations on thumbnails so we should have text and icons for both the A and B states.
```

This ToggleButton is just a wrapper of the FilterButton using the IToggleOption[] values to create it.

 ### Reviewing/Testing steps
Review on Storybook on Filters -> Atoms -> FilterButton

Screenshot 2024-11-11 at 15 04 00 

Screenshot 2024-11-11 at 15 03 25

### DIFF
--- a/packages/storybook/src/stories/Filters/atoms/ToggleButton.stories.tsx
+++ b/packages/storybook/src/stories/Filters/atoms/ToggleButton.stories.tsx
@@ -1,0 +1,120 @@
+import React, { useCallback, useState } from "react";
+import { ToggleButton } from "scorer-ui-kit";
+import { boolean, object, select, text } from "@storybook/addon-knobs";
+import { action } from "@storybook/addon-actions";
+import styled from "styled-components";
+
+export default {
+  title: 'Filters/atoms',
+  component: ToggleButton,
+  decorators: []
+};
+
+const layoutOptions = [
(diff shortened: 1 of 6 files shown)


atomworks (changes requested):
Just a couple of small things here...

In the `basic` design, there is 20px of padding to the right. It should be 4px or it will start to cause issues when integrated with the layout. That doesn't seem to come up in this code so it might be an issue on FilterButton `basic` too.

Screenshot 2024-11-19 at 16 37 01

--- 
The second is, even though I wasn't thinking about adding a toggle to the filter buttons, if we're going to keep consistency and have a `default` and `basic` state for this, we should represent that in the examples for filters somewhere. 

The only real use case I can think of where it might be useful is on systems where we have devices that are linked with external servers and when that external server deletes a resource we mark its paired resource internally as for deletion after a set time. In those products we tend to just hide those marked for deletion but want to toggle them on at time. That's a data thing, so it makes sense to have it in the filter rows rather than where we have sorting and layout.

atomworks (approved):

Issue #453, opened 2024-05-14
[UX BUG] IMG overflow in Modal


isacoder:
### Description

There is an error on modal related to the No IMG svg and the image that should be there.
Since both elements compete the wrapping is around the bigger object.
Probably the solution is to render only one.

![Screenshot 2024-05-14 at 11 46 26](https://github.com/future-standard/scorer-ui-kit/assets/10409078/78f161b3-fe06-4274-a02c-d1709e0fce3e)

Error found in v1 and v2

isacoder:
Fixed by PR #513

PR #513, opened 2024-11-15
Issue #453 Fix - Broken Media Modal Layout

atomworks:
This issue is a fix for #453 

Basic fix... this issue was visible because the placeholder image being used was failing. For this reason, it was a two part fix.

1. Added the placeholder image to repo so it didn't rely on a third party site - this fixed the image not loading.
2. When in the broken state, the image component was still occupying space with the alt text. This is now conditionally hidden until an image loads successfully. 

Additionally, due to the reliance of third party placeholder images being out of our control, I replaced all the ones I could find with local images instead.


isacoder (approved):
Thank you for updating the placeholder images the code looks good and ready to merge This is a great improvement but I'm not sure this one solves the problem entirely to close the issue.

The first time I notice was on a media box that was displayed in an Project where when the page was resized the black background would appear. I could not find the example. I'm sorry for not adding it to the issue at that time. 


I tried to replicate it for explaining it but couldn't do it quite right, but I remember than when I was debugging it there where 2 images and one had the with different than the other because one was fluid and the other one was fixed.  In this example shows that at least that behaviour persists. 

[Live example of the background](https://drive.google.com/file/d/1cSp75C3BY7Y_glO4zPwIV-7oI4DbmMq5/view?usp=sharing)

I'm approving this PR because I think is good for the UI Kit but I would like to know about your opinion if no other update will be needed to close the issue.

isacoder:
> Thank you for updating the placeholder images the code looks good and ready to merge This is a great improvement but I'm not sure this one solves the problem entirely to close the issue.
> 
> The first time I notice was on a media box that was displayed in an Project where when the page was resized the black background would appear. I could not find the example. I'm sorry for not adding it to the issue at that time.
> 
> I tried to replicate it for explaining it but couldn't do it quite right, but I remember than when I was debugging it there where 2 images and one had the with different than the other because one was fluid and the other one was fixed. In this example shows that at least that behaviour persists.
> 
> [Live example of the background](https://drive.google.com/file/d/1cSp75C3BY7Y_glO4zPwIV-7oI4DbmMq5/view?usp=sharing)
> 
> I'm approving this PR because I think is good for the UI Kit but I would like to know about your opinion if no other update will be needed to close the issue.

Reviewed with Leonard, live on chat. I think this implementation covers the issue will open a new one if a similar error happens.

PR #515, opened 2024-11-20
Temporary fix for having jsx in Storybook8 with one man army package

isacoder:
### Description

This is a potential temporary fix for the Storybook update to support JSX code in stories.

Since addDecorator was deprecated and the [JSX addon does not work without it](https://github.com/storybookjs/addon-jsx), this branch implements an alternative solution using a third-party addon: [storybook-addon-source](https://github.com/mihkeleidast/storybook-addon-source).

- The current example only showcases Alert Bars because this JSX addon only works with controls. As a result, the Alert Bars story has been updated to use controls.
- To fully adopt this addon, all component stories will need to be upgraded to use controls.

 ### Reviewing/Testing steps
Run Storybook with Node 18 or above.
Verify that the Alert Bars story displays correctly with Source informamtion

Screenshot 2024-11-20 at 1 13 46


JoshLipps:
can we cancel this?

isacoder:
> can we cancel this?

Yes, I agree in closing this one @JoshLipps 

 Improving the documentation should be the follow up instead of fixing this jsx plugin

PR #518, opened 2024-11-26
Breadcrumbs have Home icon added to top level

atomworks:
This simply adds the home icon to the top level of the breadcrumb. It is set to on by default as the top level is usually intended to be the home reference but this can be optionally switched off if the project chooses to treat this differently. This is done with the `showHomeIcon?: boolean` prop. 

At the moment no projects are known to be using this feature so this change isn't deemed to be breaking or disruptive. The first project this has been requested for will require this icon, though it is not critical.

New home icon can be seen here:
image


isacoder (approved):
All looks good, thanks for this update.

PR #520, opened 2024-11-27
Select component - Refactor and icon

atomworks:
The Select has been refactored to remove the old theming, link it to the new CSS theme and also add a new icon feature. 

- Refactored to new theme code
- Added `fieldState` to match `input`
- Slight adjustment in height
- Added new optional icon to start of select
- Updated story with new variant for the icon.

As there is a new variable for the input compact height, the `BasicSearchInput` has also been updated to use this.

This update should not create any breaking changes. If the size adjustment of the select field causes problems in existing projects, it can be fixed a custom CSS override on the `--input-height` and `--input-compact-height` variables.

Preview:
https://github.com/user-attachments/assets/f96d4330-c0b6-405e-8b79-7e800241db0a

_Note, there is a bug in the preview with icon colors on disabled. This has been fixed._

### DIFF
--- a/packages/storybook/src/stories/Form/SelectField.stories.tsx
+++ b/packages/storybook/src/stories/Form/SelectField.stories.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import styled from 'styled-components';
-import { text, object, boolean } from "@storybook/addon-knobs";
+import { text, select, object, boolean } from "@storybook/addon-knobs";
 import { action } from '@storybook/addon-actions';
-import { SelectField, PageHeader, SelectWrapper} from 'scorer-ui-kit';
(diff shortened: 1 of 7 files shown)


isacoder (changes requested):
Hello Leonard, 

Thanks for the design and css variables update. I have review the code and check the live version on local. 

The code looks well over all, the only problem is the missing dependencies in the SelectField component rederSelect callback that were detected by GitHub. Please update.

Screenshot 2024-11-28 at 8 02 08

atomworks:
This has been fixed. As the `iconColor` function was a dependency, it also needed to become a useCallback as otherwise it was reported as a problem due to it "changing on every render".

PR #524, opened 2024-12-04
Datepicker Dot Markers

isacoder:
### Description

This is a new Feature request for the Datepicker

`
A new date picker feature where a list of dates or date ranges can be provided which will then place a dot on the date picker to signify, for example, that there is content available to view within that range.
`

Moved examples data to new data helper file for DatePicker examples.
I decided to add go with the dates array for content.

 ### Reviewing/Testing steps
Test in Storybook

Filters -> Molecules -> DatePicker

Screenshot 2024-12-05 at 13 48 13

Filters -> Organism -> FilterBar

Screenshot 2024-12-05 at 13 48 43

### DIFF
--- a/packages/ui-lib/src/Filters/molecules/DatePicker.tsx
+++ b/packages/ui-lib/src/Filters/molecules/DatePicker.tsx
@@ -530,16 +530,19 @@ const DatePicker: React.FC<IDatePicker> = ({
             return (
               <CalRow key={index}>
                 {days.map((day, index) => {
+                  const dayState = cellState(day, selectedRange);
+                  const isTodayValue = isToday(day);
+
                   return (
                     <CalCellB
                       key={index}
                       disabled={isDayOutOfRange(day, availableRange)}
                       onClick={() => onCellClick(day)}
-                      state={cellState(day, selectedRange)}
+                      state={dayState}
                       thisMonth={isSameMonth(day, focusedMonth)}
-                      isToday={isToday(day)}>
+                      isToday={isTodayValue}>
                       {format(day, "d")}
-                      <ContentDot hasContent={dayHasContent(day, contentDays)} state={cellState(day, selectedRange)} isToday={isToday(day)}/>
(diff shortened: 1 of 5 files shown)


packages/ui-lib/src/Filters/molecules/DatePicker.tsx
```diff
@@ -195,6 +195,29 @@ const CalHCell = styled(CalCell)`
...
 `;
 
+const ContentDot = styled.div<{ hasContent: boolean, state?: CellStates, isToday?: boolean, }>`
+  position: absolute;
+  left: 18px;
+  bottom: 5px;
+  width: 4px;
+  height: 4px;
+  border-radius: 50%;
+  background-color: var(--info-10);
```
atomworks, line 205: Can we update this to `--primary-11` please.

packages/ui-lib/src/Filters/molecules/DatePicker.tsx
```diff
@@ -195,6 +195,29 @@ const CalHCell = styled(CalCell)`
...
+
+  ${({ state }) => (state === 'single' || state === 'start' || state === 'end' || state === 'inside') && css`
+    background-color: var(--white-12);`
```
atomworks, line 208: For `inside` can we make this `--primary-12` please. The other states should remain `--white-12`, thank you.

atomworks (changes requested):
Looking great... just three things, from small to big. Depending on your thoughts, the third might not need any work right now.

1. I'd like to revise the dot colors slightly - I'll leave the changes in the code comments.
2. While the dot is in the right place, the number itself needs to go up by 1px.
3. --The date array approach looks great though I'm concerned that it's a little messy to have required in this field again. I know we have it in the date range fields themselves but as I'd like to revise and remove that in the future when we can start to make breaking changes, I'd rather avoid it for now. What are your thoughts?--

Thank you!

isacoder:
I have updated the colors, I will review with @atomworks  points 2 and 3 to understand better these requirements before updating the PR.

Screenshot 2024-12-06 at 15 31 52

atomworks (approved):
Looks ready to go, thank you!

Issue #525, opened 2024-12-04
Pagination label broken on the page length select


atomworks:
See screenshot. This seems to be part of the refactoring work I did on the Select and Label components, discovered to be an issue in 2.6.7

![image](https://github.com/user-attachments/assets/1fb76e0b-055a-4859-8bbf-c5c1bdae7d16)

isacoder:
Fixed in #526

PR #526, opened 2024-12-05
Bugfix/525 Pagination fix with Label refactor

atomworks:
This fix was primarily to fix issue #525, however it also includes related fixes and refactoring to the `Label` component to bring that more in line with the flex approach we are taking more now. This in turn will make it more easily used among other components which should but don't yet use it.

The main change to be aware of is the addition of the new `direction` prop which will be used to handle where the label is in relation to the field is it wrapping.

`column` is the default. It is used when the label is above, such as for text inputs.
`row` is used when the label should be to the left of the input. For example, the pagination controls do this.
`row-reverse` is used when the text goes after the input, such as in the case of radio buttons.

## Deprecations and Risks
The following two props have been marked for deprecation however will default the value needed to replicate the required behaviour, along with a console warning to advise on how to fix it moving forward.

While there is no expected cases this should break, please be cautious when updating to make sure there are no unexpected errors.

As the props are fairly uniquely named, it should be easy to search your project for instances that need fixing:
`SelectField` prop `isSameRow` should be replaced by `direction='row'`
`Label` prop `rightAlign` should be replaced by `direction='row-reverse'`

## Other Changes
There have been updates to some of the colors in the `Pagination` component to bring them more in to line with recent input refactors and theming.

## Relating to 525 - Pagination Fix
Here is a screenshot of the restored and refactored Pagination component.
image

### DIFF
--- a/packages/storybook/src/stories/Form/Input/RadioButton.stories.tsx
+++ b/packages/storybook/src/stories/Form/Input/RadioButton.stories.tsx
@@ -31,17 +31,17 @@ export const _RadioButton = () => {
   return (
     <Container>
       <OptionsWrapper>
-        <Label htmlFor={'option1'} labelText={'Option 1'} rightAlign>
+        <Label htmlFor={'option1'} labelText={'Option 1'} direction='row-reverse'>
           <RadioButton {...{ name, disabled, currentChecked }} id={'option1'} value='option1' onChangeCallback={handleChange} />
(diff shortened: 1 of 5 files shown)


isacoder (approved):
Thank you for making this fix quick and working with the middle improvements. 

Code looks good and I think the breaking changes and deprecations are worth it.
This is approved.

Issue #511, opened 2024-11-14
Circular Dependency Issue (with FilterDropdown)


atomworks:
Currently when compiling on my local environment I'm seeing that we recently started to get the following circular depency issue. It doesn't seem to be stopping compile or causing problems but should be addressed sooner rather than later.

```
Circular dependency: src/Filters/index.ts -> src/Filters/molecules/SortDropdown.tsx -> src/Filters/index.ts
Circular dependency: src/Filters/index.ts -> src/Filters/molecules/SortDropdown.tsx -> src/Filters/index.ts
```

PR #527, opened 2024-12-09
Fix circular dependency error for  FilterButtonDesign type

isacoder:
### Description

There was an error introduction when `FilterButtonDesign` type was added on the Filters index in the refactor of SortDropdown. https://github.com/future-standard/scorer-ui-kit/issues/511

`
Circular dependency: src/Filters/index.ts -> src/Filters/molecules/SortDropdown.tsx -> src/Filters/index.ts
Circular dependency: src/Filters/index.ts -> src/Filters/molecules/SortDropdown.tsx -> src/Filters/index.ts
`

The circular dependency was fixed by moving the type from index to the Filter types files, all files that use `FilterButtonDesign` type were updated.

 ### Reviewing/Testing steps

Due the nature of the error, this can only be seen on the logs for the hot reload of ui-lib package.
Run the project on local and review the circular dependency error is not present.

Screenshot 2024-12-09 at 11 57 55

### DIFF
--- a/packages/ui-lib/src/Filters/FilterTypes.ts
+++ b/packages/ui-lib/src/Filters/FilterTypes.ts
@@ -67,6 +67,8 @@ interface IFilterDropdownConfig {
   name?: string
 }
 
+type FilterButtonDesign = 'default' | 'basic'
+
 export {
   isFilterItem
 };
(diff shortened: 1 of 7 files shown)


atomworks (approved):
That's great, thank you. It'll be nice to not have to see that issue in the terminal anymore!

PR #528, opened 2024-12-09
Post-Merge Label and Switch conflict fix

atomworks:
This fixes the conflict that arose with `Switch` after it was merged along with additional `Label` refactors. There are two parts.

1. Made the `direction` prop optional, as originally intended.
2. Removed the old `sameRow` implementation from `Pagination` and updated it to the new `direction` prop.


isacoder (approved):
Reviewed pagination live and continue to works as intended.
Thank you for working in this fix quickly.

PR #530, opened 2024-12-10
UtilityHeader label props and refactor

atomworks:
This iteration was made to allow the labels for the back and share buttons to be overridden. Whilst doing this and adding to the story I realised we quickly began to overload the props we were sending and things were getting messy. This becomes more problematic, the deeper that components are nested. For this reason I decided to refactor this early while it is only being used in one project and we can limit the impact of breaking changes.

## How to Update to New Props
The following props have been removed and will need to be changed: `backLink`, `showShareLink` and `shareLink`

They have been replaced with two new object props: `back` and `share` which contain everything that is needed.

Instead of `backLink`, use:
```
back = {
  link: '/example',   // This replaces backLink
  label: 'Back'       // Optional but intended for localisation overrides
}
```

Instead of `shareLink` and `showShareLink` you can do the follow. Not that `showShareLink` is not required as it is optional and will default to `true`.

```
share = {
  link: true,             // This replaces shareLink
  label: 'Back'           // Optional but intended for localisation overrides
  copiedLabel: 'Copied.'  // Shown after is clicked. Optional but intended for localisation overrides
}
```

Both of these new props also have a `show` property you can set. As these are defaulted to `true` by default, they are not needed and mostly intended for use if you need to disable and or conditionally disable the parts more conveniently. 

The back button will also not show if the `link` property is missing.

### DIFF
--- a/packages/example/src/pages/Layouts.tsx
+++ b/packages/example/src/pages/Layouts.tsx
@@ -21,11 +21,10 @@ const Layouts: FC = () => {
 
   const ExampleContent : IHeaderContent = {
     UtilityHeaderOptions: {
-      backLink: '/',
+      back: { label: "Back", link: "/" },
       breadcrumbs: [{text:'Examples', href:'/'},{text:'Two', href:'#2'},{text:'Three', href:'#3'},{text:'Four', href:'#4'},{text:'Five', href:'#5'}],
       showBreadcrumbs: true,
-      showShareLink: true,
-      shareLink: 'http://www.example.com/'
+      share: { show: true, label: "Share", link: "https://www.example.com", copiedLabel: "Copied" }
     },
     PageHeaderArea: <PageHeader
     title='Welcome'
--- a/packages/storybook/src/stories/Global/GlobalUI.stories.tsx
+++ b/packages/storybook/src/stories/Global/GlobalUI.stories.tsx
@@ -70,11 +70,10 @@ const buttonList : IButtonStack[] = [
 
 const ExampleContent : IHeaderContent = {
   UtilityHeaderOptions: {
-    backLink: '/',
+    back: { label: "Back", link: "/" },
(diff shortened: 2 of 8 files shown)


isacoder (changes requested):
Hello Leonard,

Thanks for following up with this fix. The refactor breaking changes look necessary and I don't think there will be any issues with that.

I had to make some local updates to run this fix, so I'm not sure the result of the live review is fair. I see some issues with the passed "shared" label not been updated.

Please review the comments and let me know if you have any questions.

atomworks:
Thanks for the review @isacoder and apologies for it being in such a rough state, I'm not sure how I missed that string of problems when testing as it was certainly working when I did. Perhaps I made the error of a hasty last minute rename which didn't get the same attention.

I've just pushed some changes that should resolve all that. Considering the issue of defaults and how it was being handled, it felt like the best pattern in the end would be just to break down those components they could have there prop defaults managed at that level.

As a thought on that, and maybe this is just more of a note for me than anything, perhaps anytime we use an object in the future to pass through a set of props for something nested, that probably means the nested thing should be it's own component at some point. In the original PR, those things were part of the component, which is why it got a bit messy. It makes a lot more sense now.

isacoder (changes requested):
Thanks for the update, everything is looking good and almost ready to merge.

However I discovered that I made a mistake in my comment about removing the $variables.

This is an scenario where they are necessary.
I was looking at the console while running the local version and noticed some warnings about props. Since the UtilityHeaderBack is using Link Component from the React Router Package the way props are defined is important. You can either follow their advice and have them to be lowercase or use $ at the beginning. That should remove the warning.

Screenshot 2024-12-17 at 21 29 28

atomworks:
In the end, as it's a good indicator for a prop that's effecting a styled component (and so will have a visual impact, rather than just a data change), it's useful to see and doesn't create a weird contrast to all our other camel case props.

isacoder (approved):
Thanks for the refactor and polish of UtilityHeader. It's great to have this for the final release of the year!

PR #540, opened 2025-01-10
Fix to have colon conditional in Toggle Button

### LINKED ISSUE #539: `:` is not conditionally shown in ToggleButton
Currently the `:` appears regardless of the props provided.

The `:` should only show if the first part is provided. We should also remove the space that is before it.

isacoder:
### Description

 This is a fix update for Toggle button, that will make `:` sign conditional to receiving category label.

Fixes #539

Screenshoot
Screenshot 2025-01-09 at 21 57 47

Screenshot 2025-01-09 at 21 57 55

### DIFF
--- a/packages/ui-lib/src/Filters/atoms/ToggleButton.tsx
+++ b/packages/ui-lib/src/Filters/atoms/ToggleButton.tsx
@@ -23,7 +23,8 @@ const ToggleButton: React.FC<IToggleButton> = ({ options, categoryLabel, selecte
 
   return (
     <FilterButton icon={options[selectedIndex].icon} onClick={() => onToggleCallback(selectedIndex)} {...{design}} {...props}>
-      {`${categoryLabel} : ${options[selectedIndex].text}`}
+      { categoryLabel && `${categoryLabel} : `}
+      {`${options[selectedIndex].text}`}
     </FilterButton>
   );
 };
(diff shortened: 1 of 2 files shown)


packages/ui-lib/src/Filters/atoms/ToggleButton.tsx
```diff
@@ -23,7 +23,8 @@ const ToggleButton: React.FC<IToggleButton> = ({ options, categoryLabel, selecte
 
   return (
     <FilterButton icon={options[selectedIndex].icon} onClick={() => onToggleCallback(selectedIndex)} {...{design}} {...props}>
-      {`${categoryLabel} : ${options[selectedIndex].text}`}
+      { categoryLabel && `${categoryLabel} : `}
```
atomworks, line 26: Can we remove the whitespace between the label and the `:`, so it sits against it, for example: "Layout: Grid"

atomworks (changes requested):
Just one small request, otherwise this looks good.

packages/ui-lib/src/Filters/atoms/ToggleButton.tsx
isacoder: Thanks for catching this one :D!

atomworks (approved):

PR #541, opened 2025-01-11
Datepicker, cancel and apply buttons

isacoder:
### Description

Updated the DatePicker component to include "Apply" and "Cancel" buttons.

This update introduces the following new props to DatePicker:

```ts
  cancelText = 'Cancel',
  applyText = 'Apply',
  hasApply = false,
  onToggleCallback = () => { },
  onCancelCallback = () => { },
  onApplyCallback = () => { },
```

- This update is backward compatible for both `DatePicker` and `FilterBar`, meaning no changes are required unless DropdownDatePicker is used as a standalone component.

- `DatePicker` prop `hasApply` is false ensuring no UI changes for existing implementations.
- In `DropdownDatePicker`, `hasApply` defaults to true. To maintain the previous behavior, set `hasApply` to `false`. If the new behaviour is desired the developer needs to implement the `onCancelCallback` and `onApplyCallback` on their side to handle the `selected` value.

I have made a document of the use cases for Apply, Cancel and Toggle 
https://docs.google.com/spreadsheets/d/1POe9uZxKXtLhQFF6DIV-RclUp7T7oXgSeSLlbYmE38g/edit?usp=sharing

The mayor change updates were in `Toggle` and `Click Outside`

**Before:**

**Toggle Close:** will send updated value
**Click Outside:** Closes Dropdown without update

**Now**

**Toggle Close**: Does not send update and keeps the current value even if it's not apply. When reopened it will show.
**Click Outside:** Does not closes the Dropdown

### Reviewing/Testing steps
Please Review in Storybook under Filters -> Organisms -> Filterbar
[Demo Video](https://drive.google.com/file/d/11cLXMDOCOpWC0Y1O1a1EoGvID_lKynOa/view?usp=drive_link)

**Updated**
Screenshot 2025-02-06 at 15 24 59

**Invalid values. disabled Apply button**

Screenshot 2025-02-06 at 15 25 09

**New selection**
Screenshot 2025-02-06 at 15 25 16
Screenshot 2025-02-06 at 15 25 21

No selection disabled Apply
Screenshot 2025-02-06 at 15 25 28

### DIFF
--- a/packages/storybook/src/stories/Filters/organisms/FilterBar.stories.tsx
+++ b/packages/storybook/src/stories/Filters/organisms/FilterBar.stories.tsx
@@ -203,6 +203,8 @@ export const _FilterBar = () => {
       dateTimeTextUpper: language === 'english' ? 'From' : 'から',
       dateTimeTextLower: language === 'english' ? 'To' : 'まで',
       timeZoneTitle: language === 'english' ? 'Timezone' : '時間帯',
+      cancelText: language === 'english' ? 'Cancel' : 'キャンセル',
+      applyText: language === 'english' ? 'Apply' : '完了',
       lang: language === 'english' ? 'en' : 'ja',
       selected: InitialSelectedDate,
       availableRange: datesRange,
--- a/packages/storybook/src/stories/Filters/organisms/FilterBar.stories.tsx
+++ b/packages/storybook/src/stories/Filters/organisms/FilterBar.stories.tsx
@@ -121,7 +121,7 @@ export const _FilterBar = () => {
   const hasShowMore = boolean('Has Show More', true);
   // valid formats - https://date-fns.org/v2.25.0/docs/format
   const resultsDateFormat = text('Results date format', 'yyyy-MM-dd HH:mm');
-  const datePickerHasApply = select('Datepicker has Apply', { true: true, false: false }, true);
+  const datePickerHasApply = boolean('Datepicker has Apply', true);
 
   // Sent to checkbox in TableRow via Table component.
   const selectCallback = useCallback((checked: boolean, id?: string | number) => {
--- a/packages/storybook/src/stories/Filters/organisms/FilterBar.stories.tsx
+++ b/packages/storybook/src/stories/Filters/organisms/FilterBar.stories.tsx
@@ -114,7 +114,6 @@ export const _FilterBar = () => {
   const [data, setData] = useState<ITableSampleData[]>(dataInitialState);
   const [rows, setRows] = useState<ITypeTableData>(rowMaker(dataInitialState));
   const [filters, setFilters] = useState<IFilterResult[]>([]);
-
   /**
  * Story interaction section
  */
@@ -122,6 +121,7 @@ export const _FilterBar = () => {
   const hasShowMore = boolean('Has Show More', true);
   // valid formats - https://date-fns.org/v2.25.0/docs/format
   const resultsDateFormat = text('Results date format', 'yyyy-MM-dd HH:mm');
+  const datePickerHasApply = select('Datepicker has Apply', { true: true, false: false }, true);
 
   // Sent to checkbox in TableRow via Table component.
   const selectCallback = useCallback((checked: boolean, id?: string | number) => {
@@ -208,7 +208,8 @@ export const _FilterBar = () => {
       lang: language === 'english' ? 'en' : 'ja',
       selected: InitialSelectedDate,
       availableRange: datesRange,
-      contentDays: dataContentDays
+      contentDays: dataContentDays,
+      hasApply: datePickerHasApply
     }
   ]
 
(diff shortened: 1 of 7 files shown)


packages/ui-lib/src/Filters/atoms/FilterDropHandler.tsx
```diff
@@ -84,77 +84,105 @@ interface IFilterDropHandler {
...
+    useClickOutside(mainRef, clickOutsideClose);
+
+    const handleToggleOpen = useCallback((minWidth: number, minHeight: number) => {
+      if (!buttonWrapperRef.current) { return; }
+
+      const buttonRect = buttonWrapperRef.current.getBoundingClientRect();
+      if (!buttonRect) { return; }
+      const position: IOpenPos = getDropPosition(buttonRect, minWidth, minHeight);
+
+      onToggleOpenCallback(!openState.isOpen);
```
isacoder, line 149: I corrected the indentation of this file so it's a bit confusing, there are 3 updates in this file

I have corrected the logic here on the handleToggleOpen, previously was sending the previous state of openState, now I'm sending the updated state. This is reflected in the FilterBar update that sends the callback when isOpen is false.

onClickOutsideClose  previously named handle close it will review noCloseOnClickOutside, since the FilterDropdownHandler only closes onClickOutside.

I have updated the Component to have forward reference to add the cancelClose functionality from outside.

packages/storybook/src/stories/Filters/organisms/FilterBar.stories.tsx
```diff
@@ -114,14 +114,14 @@ export const _FilterBar = () => {
...
   const [filters, setFilters] = useState<IFilterResult[]>([]);
-
   /**
  * Story interaction section
  */
   const singleFilter = boolean('Single Filter', false);
   const hasShowMore = boolean('Has Show More', true);
   // valid formats - https://date-fns.org/v2.25.0/docs/format
   const resultsDateFormat = text('Results date format', 'yyyy-MM-dd HH:mm');
+  const datePickerHasApply = select('Datepicker has Apply', { true: true, false: false }, true);
```
atomworks, line 124: Could be a boolean similar to `singleFilter` and `hasShowMore` do above, for consistency.

packages/ui-lib/src/Filters/molecules/DropdownDatePicker.tsx
```diff
@@ -65,39 +73,61 @@ const DropdownDatePicker: React.FC<IDropdownDatePicker> = ({
...
+    if (isOpen && (!mountedPicker.isMount)) {
+      setMountedPicker((prev) => {
+        return { ...prev, isMount: true };
+      });
+    }
+  }, [hasApply, mountedPicker.isMount, onToggleCallback, selected]);
+
+  const handleOnCancel = useCallback(() => {
+    if (pickerValue.current && (pickerValue.current !== selected)) {
+      pickerValue.current = selected === undefined ? null : selected;
```
atomworks, line 90: There are a few places you have used this pattern to override `pickerValue.current`, which is directly updating that ref. My reaction is that while it works, it seems like it's not best practice... however I think this is likely an issue that's not relevant in this particular usage.

So my question here is, is there any chance of problems directly manipulating this `.current` state? Is there a better alternative pattern to this, or is this best practice?

Thanks! Happy to read any links if you've got any examples or if you want to jump on a call let me know. Aside from this sanity check and the other comment though, I think we're good!

atomworks (commented):
Thanks for all your effort on this, I can see you've put a lot of care and attention in to both the code and the user experience considerations too.

I have a question, though perhaps unlikely to require any changes, probably more for my understanding. And then one very small request for the variable type in the story. I've made comments in the code for each.

Edit: One more small request actually, that's minor but a bit more work than the above. Can we disable the apply button when there is no change to the value? I think this will help a little when making the remaining pending state obvious if you toggle the filter closed and open from the button.

packages/ui-lib/src/Filters/molecules/DropdownDatePicker.tsx
isacoder: I think I would understand better your concerns in a call.
Will follow up on slack

packages/storybook/src/stories/Filters/organisms/FilterBar.stories.tsx
isacoder: Updated

packages/ui-lib/src/Filters/molecules/DropdownDatePicker.tsx
isacoder: Since we have review together on chat the usage of this reference is to keep value of the selection between re-renders, I will close this comment.

The conclusion:

The reassignment of values is necessary because the `initialValue` prop in DatePicker can be undefined, but null values are not accepted. Meanwhile, pickerValue.current can be null, but it cannot be undefined.

`pickerValue` is the middle reference between the displayed value and the selected value and it helps with validation and becomes the source of truth of what to send as proper callback.

isacoder:
I have updated the logic to disable Apply button when the selected value is the same to displayed value as requested

Screenshot 2025-02-14 at 15 01 51

atomworks (changes requested):
Changes look good but just spotted on more small thing we should add.

Can you add the buttons to the basic dropdown story too under `Filters > Molecules > Datepicker`.

Sorry I missed this on the first pass

PR #542, opened 2025-01-14
Added rules for spacing require dot (Fixes #531)

### LINKED ISSUE #531: Required Blue Dot has lost spacing
I think this must have occured while I was refactoring the labelled and I missed it.

The required dot when it is used is sitting against the label rather than with the correct gap.
image

atomworks:
After the label refactor, the new required dot lost its spacing as reported here #531. It's been restored here:

Before:
image

After:
image

I checked various implementations of Label after doing this and noticed no regression or errors in things like Pagination. However, if these can be double checked it would be highly appreciated.


isacoder (approved):
I did review inputs, normal behavior keeps the gap if the content is squished with less than 50px the dot gets smashed. However I don't think that would be normal behavior.  I will approve.

Issue #538, opened 2024-12-24
Enchance SplitLayout for better code integration related to open state


atomworks:
The SplitLayout component needs a way to track the current state easily, and / or also a callback to run when the side area is shown or hidden.

atomworks:
This has been fixed and released

PR #543, opened 2025-01-21
Add SideArea initial State and callbacks

isacoder:
### Description

This PR is related to the request of [Enhance SplitLayout for better code integration related to open state](#538)
 

 New `onSideAreaStateChange` for sideAreaState Changes:
A callback is now triggered every time the sideAreaState changes, allowing developers to monitor and respond to state transitions. 

New `defaultCollapsed` Prop added to sideArea prop:
A defaultCollapsed prop has been added to allow the side area to start in a collapsed state. This is particularly useful for scenarios where no stream data is available.

Behavior:
The defaultCollapse prop serves as an initial state only. If a localStorage variable for the side area exists (e.g., from previous interactions), its value takes precedence over defaultCollapse. As such, if the side area was last left open by the user, it will open again by default, ensuring the previous behavior remains unchanged.

 ### Reviewing/Testing steps
 
 I have added the callback and defaultCollapse prop to Examples and to Storybook.
 However currently the Storybook knobs crashes on selection :( ... this might be fixed with the update that is about to be merged. However I didn't tested this yet due time.
 
Due this the way to  review this is in Storybook (Global UI -> Organisms -> SplitLayout -> Nested) the callback functionality and review the defaultCollapse functionality in examples
 
 
Screenshot 2025-01-20 at 21 59 52

[Video Examples](https://drive.google.com/file/d/1K0BxetzlqZHK4oumgTu9Mj9uEwWNwmRa/view?usp=drive_link)
[Video Storybook](https://drive.google.com/file/d/1ST0V-mK1g113WNr4DY9yJ38RrfkUvuyh/view?usp=drive_link)

### DIFF
--- a/packages/example/src/pages/SplitLayout.tsx
+++ b/packages/example/src/pages/SplitLayout.tsx
@@ -12,11 +12,11 @@ const SplitLayouts: FC = () => {
 
 
   const sideStateCallback = useCallback((currentState) => {
-    console.log(`Side Area Current State ${currentState}`);
+    console.log(`Side Area Current State: ${currentState}`);
   },[]);
 
-  const areaAContent = <FlexContentPlaceholder title='Area A' />;
-  const areaBContent = <FlexContentPlaceholder title='Area B' />;
+  const areaAContent = <FlexContentPlaceholder title='Main Area' />;
+  const areaBContent = <FlexContentPlaceholder title='Side Area' />;
 
   return (
       <ThemeProvider theme={defaultTheme}>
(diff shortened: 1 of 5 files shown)


packages/example/src/pages/SplitLayout.tsx
```diff
@@ -1,15 +1,20 @@
...
   const {onThemeToggle, isLightMode} = useThemeToggle();
-  
+
   const [layout] = useState<'horizontal'|'vertical'>('horizontal');
   const [reverse] = useState<boolean>(false);
-  
+
+
+  const sideStateCallback = useCallback((currentState) => {
+    console.log(`Side Area Current State ${currentState}`);
```
atomworks, line 15: Just as a small readability tweak, it would be good to have a `:` in there, like this...

console.log(`Side Area Current State: ${currentState}`);

atomworks (approved):
Changes look good. One small request in the comments but don't need to re-review or hold things up for that. 

And there's one consideration we should follow up on that I've messaged you about. This is just making sure the nested story shows a distinct message for each instance, so we can see the callbacks are different.

Issue #536, opened 2024-12-23
$IconInGutter prop in UtilityHeader needs responsive integration in main layout component


atomworks:
This feature was not implemented in `Layout` although all the parts are there!

PR #548, opened 2025-02-04
Responsive alignment of UtilityHeader

atomworks:
Fixes issue #536 .

Previously the `$iconInGutter` was hardcoded so the `UtilityHeader` alignment didn't change on small breakpoints. This has been updated to be responsive but can be overridden by the props.

Before:
Screenshot 2025-02-04 at 20 26 12

After:
Screenshot 2025-02-04 at 20 24 55


isacoder (approved):
Thanks for the update.

Issue #556, opened 2025-02-26
[TypeTable] - Width prop is not functional.


dev-shivang:
The `Width` prop is present but does not affect the table's column width and functional. 

### Current Functionality
The attached image showcases a table created using the `TypeTable` component. In this use case, the `S. No.` column has `min-width` prop.

### Expected Functionality
The `S. No.` column should have a fixed width since it does not need to scale with the table width.

![Image](https://github.com/user-attachments/assets/347d0d1c-c2ea-4aff-a680-489212379351)

PR #559, opened 2025-03-03
Fixes #554 - Switch toggle functionality Issue, Fixes #555 - Notification working after quick async operation, Enhancement #556 - Added width prop for Table, Fixes #557 - Fix alignment between text and icon of Top bar

### LINKED ISSUE #554: [Switch] - Toggle functionality issue.
The <input type="checkbox"> element has a native `checked` property that indicates whether it is selected (true) or not (false). However, the current implementation uses a custom property called `defaultChecked` to manage the switch's state.

### Current behaviour
When the switch is toggled, the logic checks the `checked` property instead of `defaultChecked`.
`inputRef.current?.checked`

### Expected  behaviour
The switch position should remain unchanged until the process successfully completes, as it depends on `defaultChecked` prop.
`inputRef.current?.defaultChecked`

### LINKED ISSUE #555: [Notification] - Stops showing after quick async operation.
When an action triggers a notification on the UI and, before it disappears, another action (typically an asynchronous action) taken that also triggers a notification, no further notifications appear from that point onward.
`activeNotification` state is not updating properly.

### Impact:
No further notifications appear after this point.

### LINKED ISSUE #557: [Top Bar] - Misalignment between text and icon of button.
The SVG icon has extra `top margin` of `7px`, causing misalignment between the icons and text.
We can removing the extra margin, ensuring proper vertical alignment of the icons and text.

Image

dev-shivang:
## Description

The PR addresses multiple bug fixes and enhancements related to UI functionality and alignment issues in the project. Here's the context based on the fixes mentioned:

1. Fixes #554 - Switch Toggle Functionality Issue:

- This likely resolves an issue where the toggle `switch` was not behaving as expected. Instead of updating its position based on the `defaultChecked` prop, it was relying on `checked`, allowing it to handle process failures more effectively.

### Test scope: 
**Case:**
- Ensure the switch does not toggle until the network request successfully completes.

<br>
<br>

2. Fixes #555 - Notification Working After Quick Async Operation:
- Clear the currently displayed notification when a new notification is triggered.
- Ensures that notifications are displayed correctly even if an asynchronous operation takes quickly before the current notification is dismissed from the UI.

<br>
<br>

3. Enhancement #556 - Added Width Prop for Table:

- Added a `width` prop to `TypeTable` to set a fixed width for table columns.

image

<br>
<br>

4. Fixes #557 - Fix Alignment Between Text and Icon of Top Bar:

- Fix UI alignment issue where the text and icon in the top bar were not properly aligned, Removed `margin-top: 7px` from CSS.

image

### DIFF
--- a/packages/ui-lib/src/Alerts/atom/Notification.tsx
+++ b/packages/ui-lib/src/Alerts/atom/Notification.tsx
@@ -143,6 +143,8 @@ export type INotificationProps = {
   closeCallback?: () => void
 }
 
+export let handleDismiss = () => { };
+
 const Notification : React.FC<INotificationProps> = ({id, type ='info', message, icon = '', isPinned = false, actionTextButton, closeNow = false, closeCallback, onTextButtonClick}) => {
   const [dismiss, setDismiss] = useState(false);
   const [slideUp, setSlideUp] = useState(false);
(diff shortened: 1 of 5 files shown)


atomworks (changes requested):
Thank you for this but please could you break this down in to multiple PRs, one per issue so they can be reviewed individually. I've put a more detailed follow up note on Slack.

PR #562, opened 2025-03-06
Fixes #554 - Switch toggle functionality Issue

### LINKED ISSUE #554: [Switch] - Toggle functionality issue.

dev-shivang:
## Description

Fixes https://github.com/future-standard/scorer-ui-kit/issues/554 - Switch Toggle Functionality Issue:
This likely resolves an issue where the toggle switch was not behaving as expected. Instead of updating its position based on the `checked` prop passsed, it was relying on input element's native `checked` property, allowing it to handle process failures more effectively.

### Test scope:
**Case**:

Ensure the switch does not toggle until the network request successfully completes.

### Fix:

- Introduce a new prop called `defaultChecked` that defines the initial value and is primarily used in uncontrolled components.

- If the `checked` prop is used along with an `onChange` function, the component should handle asynchronous calls, including both success and failure scenarios and work as controlled component.


packages/example/src/pages/SwitchTest.tsx
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  return (
+    <div style={{ padding: '50px' }}>
+      <Switch
+        labelText="Fetch Data"
+        defaultChecked={isChecked} // To make it controlled
+        onChangeCallback={handleSwitchChange} // To make it controlled
+        // checked={isChecked}
+        // onChangeCallback={() => {setIsChecked(!isChecked)}}
```
isacoder, line 101: I think there might be a misunderstanding of the expected behaivour in a uncontrolled/controlled scenario. 
Just reading the comments this might be working backwards.

I have looked for a related example. This is a very general example please don't hesitate if you have specific questions about it. We can also follow up in slack.

https://www.geeksforgeeks.org/react-js-uncontrolled-vs-controlled-inputs/

packages/ui-lib/src/Form/atoms/Switch.tsx
```diff
@@ -281,7 +294,7 @@ const Switch : React.FC<IProps> = ({ state = 'default', leftTheme = 'off', right
         </SwitchInner>
       </SwitchOuter>
       {labelText ? <LabelText>{labelText}</LabelText> : null}
-      <RealInput ref={inputRef} type='checkbox' disabled={state !== 'default' && state !== 'failure'} defaultChecked={checked} />
+      <RealInput ref={inputRef} type='checkbox' disabled={state !== 'default' && state !== 'failure'} {...(isDefaultPropPassed ? { defaultChecked } : { defaultChecked: checked })} />
```
isacoder, line 297: The condition here although is syntactically valid it's a bit complicated to read 
{...(isDefaultPropPassed ? { defaultChecked } : { defaultChecked: checked })}

I would recommend to use the de verbose version for such occasion :

{
  isDefaultPropPassed 
    ? defaultChecked={defaultChecked} 
    : defaultChecked={checked}
}

However I think a review of how to handle the values is required since input type checkbox does have a defaultCheck property as well as a checked property.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/checkbox

packages/ui-lib/src/Form/atoms/Switch.tsx
```diff
@@ -191,35 +191,48 @@ interface IProps {
...
 
-const Switch : React.FC<IProps> = ({ state = 'default', leftTheme = 'off', rightTheme = 'on', labelText, onChangeCallback, checked = false}) => {
+const Switch : React.FC<IProps> = ({ state = 'default', leftTheme = 'off', rightTheme = 'on', labelText, onChangeCallback, checked = false, defaultChecked }) => {
 
   const inputRef = useRef<HTMLInputElement>(null);
   const innerRef = useRef<HTMLDivElement>(null);
-  const [ position, setPosition ] = useState<SwitchPosition>(checked ? SwitchPosition.On : SwitchPosition.Off);
+
+  const isDefaultPropPassed = defaultChecked !== undefined;
+  const initialChecked = isDefaultPropPassed ? defaultChecked : checked ?? false;
```
isacoder, line 204: When using evaluated constants is better to use useMemo for efficiency.

isacoder (changes requested):
Thanks for helping  with fix the toggle functionality of the Switch, @dev-shivang.

I’ve integrated the test file you provided into the examples project for easier review and updated the API URL for the failure scenario.

I’ve also added some general comments. I believe the current approach still doesn’t fully address the main issue. The input element is not being updated correctly in the controlled scenario.

Screenshot 2025-04-22 at 12 08 20

I have added a few references to explain what I believe might help us to reach a better solution but please don't hesitate to ask directly on slack if you have any questions.

packages/ui-lib/src/Form/atoms/Switch.tsx
```diff
@@ -191,53 +191,64 @@ interface IProps {
...
+      onChangeCallback?.(!checked);
+    } else {
+      const newChecked = !internalChecked;
+      setInternalChecked(newChecked);
+      onChangeCallback?.(newChecked);
+    }
     positionSwitch();
-    setJustUpdated(true);
-    if(onChangeCallback){ onChangeCallback(inputRef.current?.checked || false); }
   };
```
isacoder, line 252: I think this function needs to be wrapped in a "useCallback" to be aware of the internalCheck state change.

packages/ui-lib/src/Form/atoms/Switch.tsx
```diff
@@ -294,7 +305,7 @@ const Switch : React.FC<IProps> = ({ state = 'default', leftTheme = 'off', right
         </SwitchInner>
       </SwitchOuter>
       {labelText ? <LabelText>{labelText}</LabelText> : null}
-      <RealInput ref={inputRef} type='checkbox' disabled={state !== 'default' && state !== 'failure'} {...(isDefaultPropPassed ? { defaultChecked } : { defaultChecked: checked })} /> 
+      <RealInput ref={inputRef} type='checkbox' disabled={state !== 'default' && state !== 'failure'} defaultChecked={checked} />
```
isacoder, line 308: Although this preserves the usage of the previous implementation. I think this line needs to be updated, now that we are adding a defaultCheck value to the component. If the user sends a defaultCheck value this should be assigned instead of the checked value.

I'm suggesting

defaultChecked={defaultChecked?  defaultChecked : checked}

I think this way it won't become a breaking change but will improve the implementation. Please let me know your thoughts in this part of the implementation.

isacoder (changes requested):
Thank you for updating the component based on the suggested changes from the review.

I’ve tested the functionality, and it’s working well.
I think this is a great solution. It allows the user to control the checked state, and when the user chooses not to, the internal state in the component keeps it updated and ensures the UI stays in sync through re-renders.

I’ve added a couple of comments, but I believe we’re very close to wrapping this up.

Thank you for troubleshooting this complex issue and this effective solution.

isacoder (approved):
Thank you for wrapping the callback in the `customOnChange` function and for addressing the `defaultChecked` and `checked` values in the checked prop of the input.
I tested the changes locally, and everything is working as expected.
I’ll approve this PR and proceed to merge it after some cleanup in the SwitchTestCase.

isacoder:
I have updated the SwitchTest to showcase the scenarios by this fix, removing the axio dependency.

## Fail Scenario
Screenshot 2025-05-08 at 13 39 28

## Common Scenario
Screenshot 2025-05-08 at 14 08 58

atomworks (changes requested):
There's a few comments of things to change. Mostly code readability and best practice but not show stoppers. On the whole, it looks great.

atomworks (approved):
Changes look good, thank you. The mocked API call is nice and tidy.

We're missing the update to the Storybook story but let's follow up on that once this is merged.

There seems to be a merge conflict that needs fixing if you can take a look at that but I'll approve on my side and don't need to re-review after that fix.

PR #566, opened 2025-03-19
Remove italics from japanese

isacoder:
### Description
This PR is to update all the code to check for Japanese and prevent it to be italics

Current  Stories are based on newest [PR lang implementation](https://github.com/future-standard/scorer-ui-kit/pull/565), but can be updated if rejected.

 ### Reviewing/Testing steps
Storybook -> Form -> Input - Text Input
Screenshot 2025-03-19 at 12 00 08

[Input placeholder video](https://drive.google.com/file/d/1N8lsbCu55CIdsVty_xr7hDo0XSl8pPDe/view?usp=drive_link)

### DIFF
--- a/packages/example/src/pages/CustomUserDrawerPage.tsx
+++ b/packages/example/src/pages/CustomUserDrawerPage.tsx
@@ -48,7 +48,7 @@ const searchPlaceholder = "Search area names, etc.";
 
 
 const CustomUserDrawerPage : React.FC = () => {
-  const  [attributeLanguage, setAttributeLanguage] = useState('us');
+  const  [attributeLanguage, setAttributeLanguage] = useState('ja');
 
   const onLanguageToggle = useCallback(() => {
     setAttributeLanguage((prev:  string) => {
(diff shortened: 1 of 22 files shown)


atomworks (commented):
Everything is looking good here. Are you able to update this based on #565 feedback? Just want to do a sanity check with those things in place before I approve here.

isacoder:
I have updated 'us' value to 'en' value as requested and  added the missing text in the BottomDrawer example that updated to show the italics in English and the non-italics in Japanese.

Screenshot 2025-03-19 at 14 13 51


Screenshot 2025-03-19 at 14 13 45

packages/example/src/pages/CustomUserDrawerPage.tsx
```diff
@@ -45,20 +48,49 @@ const searchPlaceholder = "Search area names, etc.";
...
+      return newLang;
+    })
+  },[])
 
 
   const userDrawerBespoke = <CustomContentContainer>
     <ExampleText>
-      <p>Custom content can be injected under the user menu like this.</p>
-      <p>The styling for this is left to you to include, keeping it flexible.</p>
+      {attributeLanguage === 'en' ?
```
atomworks, line 63: As we tend to do everything English first, and English becomes our fallback language by default, it would be good to swap the order here, so it's conditional on `ja` else it goes to the English.

atomworks (approved):
I've added a small comment on something that I was unsure of yesterday but even if it's not updated, this is good. Thank you.

packages/example/src/pages/CustomUserDrawerPage.tsx
isacoder: Thank you for the suggestion, I will update.

PR #570, opened 2025-04-15
Button text-only outline

isacoder:
### Description

Update request for buttons:

Within the designs for improvements to filters, new button variations are now being used and need to be added to the UI Kit.

**The new theme options** are `Text Only` and `Outline`. These are mostly changes to the colours of the buttons and are all linked to the theme. `Text Only` implies there is no button although the clickable area is still the same with the optional icon still there although not it is expected to rarely be used.

The `Outline` variant also has a thicker outline than other buttons.

**For the xsmall button variant** change, the size of the icon has gone from `10` → `12` and the horizontal padding in the icon area has gone from `8` → `4`

No updates were made for any design differences for primary, secondary and warning buttons

 ### Reviewing/Testing steps

The storybook stories for Standard Button, WithLoading and WithIcon were added and can be review under Form, Buttons.

[Button With Icon video](https://drive.google.com/file/d/1KQnEpzbyvebNX0RjpMdq4i2RP-3UZBgd/view?usp=sharing)

Screenshot 2025-04-15 at 15 08 55

### DIFF
--- a/packages/storybook/src/stories/Form/Buttons/Button.stories.tsx
+++ b/packages/storybook/src/stories/Form/Buttons/Button.stories.tsx
@@ -12,7 +12,7 @@ export default {
 
 export const StandardButton = () => {
   const buttonText = text("Button Text", "Example Title");
-  const buttonDesign = select("Design", { Primary: "primary", Secondary: "secondary", Danger: "danger", TextOnly: "text-only" }, "primary");
+  const buttonDesign = select("Design", { Primary: "primary", Secondary: "secondary", Danger: "danger", TextOnly: "text-only", Outline: 'outline' }, "primary");
   const buttonSize = select("Size", { Xsmall: 'xsmall', Small: "small", Normal: "normal", Large: "large" }, "normal");
   const buttonDisabled = boolean("Disabled", false);
   const buttonShadow = boolean("Shadow", false);
--- a/packages/storybook/src/stories/Form/Buttons/Button.stories.tsx
+++ b/packages/storybook/src/stories/Form/Buttons/Button.stories.tsx
@@ -12,7 +12,7 @@ export default {
 
 export const StandardButton = () => {
   const buttonText = text("Button Text", "Example Title");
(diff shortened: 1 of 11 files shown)


atomworks (changes requested):
Look good. Just one small detail to change. 

~~The text only button should be the same size as the other buttons, so changing style wont change how much space its taking or the alignment of the text. In code I would just keep the border but make it transparent in this case however any way you feel is best is good.~~

Update: After speaking, I realised that I made a mistake and it's the outline buttons 2px border that is making that one bigger, so we should reduce the padding by 1px for these to compensate. 

Also, you mentioned the issue with button animation inconsistency so I took a look. It seems that the main thing missing is the transition for the text color itself. In `Button.tsx` around line `:72` we need to add one more rule to ensure the text transitions color... that should look like `color var(--speed-normal) var(--easing-primary-in-out),`

Let me know if you have any other questions on that.

isacoder:
> Look good. Just one small detail to change.
> 
> ~The text only button should be the same size as the other buttons, so changing style wont change how much space its taking or the alignment of the text. In code I would just keep the border but make it transparent in this case however any way you feel is best is good.~
> 
> Update: After speaking, I realised that I made a mistake and it's the outline buttons 2px border that is making that one bigger, so we should reduce the padding by 1px for these to compensate.
> 
> Also, you mentioned the issue with button animation inconsistency so I took a look. It seems that the main thing missing is the transition for the text color itself. In `Button.tsx` around line `:72` we need to add one more rule to ensure the text transitions color... that should look like `color var(--speed-normal) var(--easing-primary-in-out),`
> 
> Let me know if you have any other questions on that.

Thank you for the review and the update in design. I will update accordingly 🙏

isacoder:
Animation for color text and  remove of 1px Padding for outline has been updated.
Screenshot 2025-04-21 at 14 04 27

Screenshot 2025-04-21 at 14 04 37

atomworks (approved):

PR #571, opened 2025-04-16
Circular Error fixes

isacoder:
### Description

Recently there were some circular dependency errors in the console.
I have found at least 3 circular dependencies to fix during this troubleshooting.

I decided to fix them using the Roo Code extension as an experiment/practice of the tool.

I found it useful to have a tool that can analyze the folders quickly and promote a change.
The first propose change was to create new helper files almost for each function, but I requested moving the functions inside the only file that require them.

```
Let's summarize what we did to fix the circular dependencies.

Great! We've successfully fixed the circular dependencies in the project. The build is now working without any circular dependency warnings.

Here's what we did to fix the issues:

First circular dependency: src/index.tsx -> src/Form/index.ts -> src/Form/atoms/SliderInput.tsx -> src/helpers/index.tsx -> src/index.tsx

We moved the isInsideRange function from helpers/index.tsx directly into the SliderInput.tsx file
This broke the dependency on the helpers file from SliderInput

Second circular dependency: src/index.tsx -> src/Form/index.ts -> src/Form/molecules/DurationSlider.tsx -> src/helpers/index.tsx -> src/index.tsx

We identified that the areDatesEqual function was only used in DropdownDatePicker.tsx
We moved the areDatesEqual function from helpers/index.tsx to DropdownDatePicker.tsx
We removed the function from the exports in helpers/index.tsx
We also cleaned up unused imports in helpers/index.tsx
```

While writing code, I often see the potential for certain functions to be reused as helpers across components. However, I’ve realized that this pattern has been causing circular dependencies. Going forward, I’ll avoid adding such functions to the helpers module.

 ### Reviewing/Testing steps
Review this components in Storybook

Forms-> Input -> Slider Input
- Review that the value in actions is valid and is not an overflow value
Forms -> Input Duration
- Review that the text-value in duration buttons show Hours- Minutes - Seconds text correctlu
Filters -> Organisms -> Filter Bar 
- Review that the same Date input will disable the apply button

### Screenshots
I did a manual review of the components, however I also requested Roo Code to do a review of the changes and it navigated to the components. It was unable to reach DropdownDatepicker since this one is reviewed in the FilterBar function.

It's far from a good test, but it shows some good testing capabilities to explore.

Screenshot 2025-04-16 at 15 04 57

### DIFF
--- a/packages/ui-lib/src/Filters/molecules/DropdownDatePicker.tsx
+++ b/packages/ui-lib/src/Filters/molecules/DropdownDatePicker.tsx
@@ -1,14 +1,34 @@
 import React, { useRef, useCallback, useEffect, useState } from 'react';
 import styled from 'styled-components';
 import FilterDropdownContainer from '../atoms/FilterDropdownContainer';
-import DatePicker, { DateInterval, IDatePicker } from './DatePicker';
+import DatePicker, { DateInterval, IDatePicker, isDateInterval } from './DatePicker';
 import FilterDropHandler, { FilterDropHandlerRef } from '../atoms/FilterDropHandler';
-import { areDatesEqual } from '../../helpers';
+import { isEqual } from 'date-fns';
 
 const MIN_WIDTH = 470;
 const MIN_HEIGHT = 360;
 
 const Container = styled.div``;
+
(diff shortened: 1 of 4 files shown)


isacoder:
Closing this one as the alternative https://github.com/future-standard/scorer-ui-kit/pull/572  has been merged

PR #572, opened 2025-05-12
Circular Dependency Fix (Manual Alternative)

atomworks:
This PR contains two commits, which when reviewing I recommend you switch to review individually. This is because the first is moving the parts that caused the circular dependency issue, then the other renames the interfaces to match our naming convention in such cases. Viewing together just has a lot of noise as a result.

In response to #571, I was bothered by the fact it moved functions out of centralised places which goes against how we try to keep things scalable and reusable. Even though it was okay, it felt like bad practice that left some rough edges.

This fix moves `isDateInterval` out of `DatePicker.tsx` and in to `helpers`. It also moves the two interfaces it uses out of the same file to the `Filters/index.ts` file which is should be the central point for all the types and interfaces being past around in the folder.

I feel this approach keeps within our best practices and that move the code outward to fix the problem rather than inward (to individual components).

# Notes For Documentation
While fixing the circular dependency issue, the two interfaces `DateInterval` and `DateRange` were moved to a higher level and renamed `IDateInterval` and `IDateRange` to become consistent with how we name exported interfaces.

If you are using either of these in projects then you will need to update them in your code base. While you can do this any way you prefer, it can be done easily with find and replace in VSCode (or any IDE with a Regex based find and replace feature) by using the following...
Find: `(?<![A-Za-z])DateInterval` -> Replace `IDateInterval`
Find: `(?<![A-Za-z])DateRange` -> Replace `IDateRange`

_Note: This regex should fixed references that aren't already prefixed and also exclude similar strings such as the function `isDateInterval`._

_Please however take a moment to check that any changes are valid and compile without error._

### DIFF
--- a/packages/storybook/src/stories/Filters/molecules/DatePicker.stories.tsx
+++ b/packages/storybook/src/stories/Filters/molecules/DatePicker.stories.tsx
@@ -21,7 +21,7 @@ export default {
 const exampleCallback = <T extends Function>(fn: T): T => {
   /** A toString to render the function in storybook */
   // eslint-disable-next-line no-param-reassign
-  fn.toString = () => '   updateCallback: (data: DateInterval | Date) => {}';
+  fn.toString = () => '   updateCallback: (data: IDateInterval | Date) => {}';
   return fn;
 };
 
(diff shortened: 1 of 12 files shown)


isacoder (approved):
Thank you for the alternative update to remove the circular error. 
The error was removed and this keeps consistency of our code and reusability of the code. 

I will add the update requirements of the interfaces in the release notes.

PR #573, opened 2025-05-15
Filter Dropdown Interaction Improvements

isacoder:
### Description

Filter Dropdown Interaction Improvements. The update will include:

- Adding optional Submit, Cancel and Reset buttons to each dropdown as a standard.
- Considerations for custom components to easily use the same dropdown UI pattern. 
*Possibly this goes in to the DropdownContainer component.*

~Currently evaluating if onSelect property should become optional and callback be disabled if hasApply prop is provided.~
 onSelect will act as onApply 
 
Will Follow up in a different PR with 
 
 -> FilterBar correct integration and Storybook example
 -> Integration of Footer controls with Datepicker

New `IWeight` interface for Icons weight. It was added for the ButtonWithIcon to have font-weight for text and icon. It does not cause a breaking change,  but maybe there is a suggestion for a more polish version of this update. This was required for the `SortingButtonWrapper`  in the  `FilterResultsToolbar` of the `FilterDropdown`.

 ### Reviewing/Testing steps
 
 Run Storybook
 
 Filters -> Molecules -> FilterDropdown -> MultipleExamples

 -> Verify dropdown callbacks work when selection as previously to this implementation

 
 Filters -> Molecules -> FilterDropdown -> Dropdown With Apply and Reset
 
 
 - Review Dropdown interactions following the Design rules in Figma
 
 ```
- These new buttons will be shown if the filter pattern requires manual applying of changes.
- If we use the older pattern, these will not be shown.
- Apply will submit the values selected and close the dropdown.
- Cancel will revert any changes made and return to the currently active state for this filter.
- Reset will go the default state of the filter, for instance, no values selected, search empty and sorting reset. It will not close the dropdown.
- Buttons are disabled if there actions would not change anything, for example:
- Reset is disabled in the default state.
- Cancel is changed to Close when the filter set is the same as the active one.
- Apply is only active when there is different value to the current state.
- After a Reset, it is assumed that Apply will be disabled as it should meet those requirements described.
```

I’ve put together a [document]( https://docs.google.com/spreadsheets/d/1XG2n_Gh-lawRhCg6r3bmMtO1V6_v-m-3sM9OXuyb1s0/edit?usp=sharing 
) to demonstrate some button interactions, such as selection and reset.
Feel free to use it for reference or request changes if there are any mistakes.
 
 
### Screenshot and Video
   [Video Demo](https://drive.google.com/file/d/1ph5imm99nJU_ZYiLqZH1LR-ZYV2qynoW/view?usp=sharing)
   
Screenshot 2025-06-23 at 16 22 56

### DIFF
--- a/packages/storybook/src/stories/Filters/molecules/FilterDropdownExamples.stories.tsx
+++ b/packages/storybook/src/stories/Filters/molecules/FilterDropdownExamples.stories.tsx
@@ -18,7 +18,7 @@ import {
 import { generateIconList } from '../../helpers';
 
 export default {
-  title: 'Filters/molecules',
+  title: 'Filters/molecules/FilterDropdown',
   component: FilterDropdown,
   decorators: []
 };
@@ -90,7 +90,7 @@ const baseExample = [
   { text: "Fugiat Nulla", value: 8 },
 ];
 
-export const _FilterDropdown = () => {
+export const MultipleDropdownExamples = () => {
   const iconList = generateIconList();
 
   const language = select("Language", { English: 'english', Japanese: "japanese" }, "japanese");
--- a/packages/storybook/src/stories/Filters/molecules/FilterDropdownWithFooter.stories.tsx
+++ b/packages/storybook/src/stories/Filters/molecules/FilterDropdownWithFooter.stories.tsx
@@ -1,0 +1,108 @@
+import React, { useState, useCallback } from 'react';
+import styled from 'styled-components';
+import { boolean, text, select } from "@storybook/addon-knobs";
+import { action } from '@storybook/addon-actions';
+import {
+  FilterDropdown,
+  IFilterValue,
+  PageHeader,
+} from 'scorer-ui-kit';
+import { generateIconList } from '../../helpers';
+
+export default {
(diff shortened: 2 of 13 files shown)


atomworks (approved):
All looks good though there is one layout thing I'd like to fix, it can be in a follow up PR.

The sort order seems to be centre aligned to the area to the design. I can see how it's ended up there, though there are two small changes needed. First is to left align the content, at the moment it's centred which for this kind of button isn't intended. Second, we take shrink the width a bit as with the longest word it should be the width of the content + padding. 

For reference, this is it in the implementation:
image

And this is from the design...
image

Again, you are more than welcome to follow up on this is another PR and release this current PR code without that fix though.

isacoder:
Thank you for approving the PR. 
I felt it was easier to make this quick change and request the review of the update in this PR to keep the context.

I want to confirm this is the expected result of these changes. I'm mostly interested that the Japanese version is working as expected.

Japanese Version
Screenshot 2025-07-01 at 17 28 18

English Version

Screenshot 2025-07-01 at 17 28 47

atomworks (approved):

PR #574, opened 2025-05-19
Fixed eslint error with default exports

atomworks:
This PR fixes ESLint warnings related to the direct export of story objects in Storybook files. 

The issue was that story configuration objects were being directly exported using export default {...} syntax, which violates the ESLint rule requiring objects to be assigned to variables before export. The fix changes the current export in to a const and re-adds the `export default` to the end of the file.

It also removes one unused import that was also showing as a warning in ESLint.


isacoder (approved):
All looks right and the project is running, this PR is approved.

Thanks for this update, removing the warnings will help to review other warnings and errors output on the console.

PR #575, opened 2025-06-02
Updating switch case knobs for storybook

dev-shivang:
### Description

This PR introduces improvements to the Switch Story in Storybook to allow better testing and visualization of both checked and defaultChecked props using knobs.

Previously, the checked prop was not clearly controlled via knobs, which made it difficult to simulate and test different states of the Switch component (checked, unchecked, and undefined). This change addresses that by adding a select knob for checked with three options: undefined, true (checked), and false (unchecked). The defaultChecked prop remains separately configurable via a boolean knob.

```
Introduced a new select knob for the checked prop with tri-state options:

"undefined" – to allow default behavior via defaultChecked

"true" – to force the switch to be in the checked state

"false" – to force the switch to be in the unchecked state

Ensured that defaultChecked is only passed when checked is undefined to avoid React warnings.

Cleaned up the checkedOption parsing for clarity and predictability.

Retained backward compatibility for consumers of the component in Storybook.
```

This improves developer experience when using Storybook to simulate real component usage, including controlled and uncontrolled scenarios.

 ### Reviewing/Testing steps

Please review the following in Storybook under:

 Form → atoms → Switch

- Verify that the Checked knob appears as a dropdown with options: undefined, checked, unchecked.

- Verify that the defaultChecked knob appears as a boolean toggle.

- Test different combinations:

  - Checked = undefined, defaultChecked = true/false
  
  - Checked = true or false, defaultChecked should have no effect
  
  - Observe that the onChangeCallback fires correctly in all cases.
  
  ### Screenshots
  
  Here is a screenshot of the expected result.
  
Screenshot from 2025-06-02 19-08-22

### DIFF
--- a/packages/storybook/src/stories/Form/Switch.stories.tsx
+++ b/packages/storybook/src/stories/Form/Switch.stories.tsx
@@ -1,3 +1,4 @@
+
 import React from 'react';
 import { action } from '@storybook/addon-actions';
 import { text, select, boolean } from '@storybook/addon-knobs';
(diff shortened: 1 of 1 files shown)


isacoder (changes requested):
Thanks for updating the story.

I have added just a couple of recommendations to improve the clarity of the implementation. Please verify and let me know if you have any questions.

isacoder (approved):
Thank you for addressing my recommendations. 
I approved this PR.

PR #585, opened 2025-11-08
Add HLS support to LineUIVideo component

kndt84:
### Description

This PR adds HLS (HTTP Live Streaming) support to the LineUIVideo component.

**What does this PR do?**
- Adds HLS streaming capability to the LineUIVideo component
- This is a new feature implementation
- Enables playback of `.m3u8` HLS streams in addition to regular video files

**Background Context:**
The LineUIVideo component previously only supported standard video file formats. This enhancement adds support for HLS streaming, which is widely used for adaptive bitrate streaming and live video delivery.

**Changes:**
- Added `hls.js` dependency (v1.4.0) for HLS playback support
- Added `@types/hls.js` for TypeScript type definitions
- Implemented HLS detection and initialization logic in LineUIVideo component
- Added error recovery mechanisms for network and media errors
- Supports both HLS.js (for browsers that don't natively support HLS) and native HLS support (Safari)

- The implementation automatically detects HLS streams by checking for `.m3u8` file extensions in the src URL
- For browsers supporting HLS.js, the component uses HLS.js with low latency mode enabled
- For Safari and other browsers with native HLS support, it falls back to native playback
- Error handling includes automatic recovery attempts for network and media errors
- The HLS instance is properly cleaned up on component unmount to prevent memory leaks
- The component maintains backward compatibility with regular video file playback

### Reviewing/Testing Steps

**Testing Instructions:**
1. Test with an HLS stream URL (`.m3u8` file)
2. Verify playback works in different browsers (Chrome, Firefox, Safari)
3. Test with regular video files to ensure backward compatibility
4. Test error scenarios (invalid URL, network interruption)
5. Verify that autoPlay, controls, muted, and loop props work correctly with HLS streams

**Example Usage:**
```typescript
<LineUIVideo
  src="https://example.com/stream.m3u8"
  autoPlay={true}
  controls={true}
  muted={false}
/>
```

**Key Test Cases:**
- HLS stream playback in Chrome/Firefox (using HLS.js)
- HLS stream playback in Safari (using native support)
- Regular video file playback (backward compatibility)
- Error recovery when network issues occur
- Component cleanup on unmount

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
   },
   "devDependencies": {
     "@svgr/cli": "^5.4.0",
+    "@types/hls.js": "^1.0.0",
     "@types/jest": "^25.1.4",
     "@types/lodash.debounce": "^4.0.6",
     "@types/lodash.defaultsdeep": "^4.6.6",
@@ -81,6 +82,7 @@
     "@future-standard/icons": "^2.0.0",
     "@types/react-dom": "^17.0.0",
     "date-fns": "^2.16.1",
+    "hls.js": "^1.4.0",
     "immutability-helper": "^3.1.1",
     "lodash.debounce": "^4.0.6",
     "lodash.defaultsdeep": "^4.6.1",
--- a/src/LineUI/LineUIVideo.tsx
+++ b/src/LineUI/LineUIVideo.tsx
@@ -1,5 +1,6 @@
 import React, { useState, useEffect, useCallback, useRef, useContext } from 'react';
 import styled, { css } from 'styled-components';
+import Hls from 'hls.js';
 
 import LineSet from './LineSet';
 import { LineSetContext } from './Contexts';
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -81,6 +82,7 @@
     "@future-standard/icons": "^2.0.0",
     "@types/react-dom": "^17.0.0",
     "date-fns": "^2.16.1",
+    "hls.js": "^1.4.0",
```
JoshLipps: I wonder if this will cause a failure if not used as a peer dep?

JoshLipps:
Superseded by #640, which rebuilds this feature on current `main` (the diff here is 160 commits behind and conflicts with the monorepo + React 19 restructure).

Key difference: #640 makes `hls.js` an **optional peer dependency** loaded via dynamic `import()`, so consumers who don't stream HLS don't pay the ~230 KB bundle cost. Also fixes a few details in the effect (dep list, detection regex, StrictMode cleanup, Safari native path).

Suggest closing this PR in favor of #640.

PR #587, opened 2025-11-11
Fix placement when scrolled

atomworks:
After scrolling, trying to drag the anything on the line UI would be offset from the mouse by the distance of the drag. While I'm sure we fixed this at least twice in the past, it seems like it snuck back in!

I've (re?)added an offset to the three instance the position is calculated to accommodate the offset. You can test this on the examples Line UI page by scrolling.


isacoder (approved):
I tested this update in LineUI by adding padding in the examples to generate scrolls. In the previous implementation, the line would jump the scroll distance, but in this one it works correctly.

Thank you for this update, I will go ahead and merge.

PR #594, opened 2026-02-10
Migrate ui-lib build from microbundle-crl/react-scripts to Vite/Vitest

JoshLipps:
Replace outdated CRA-era tooling with modern equivalents:
- microbundle-crl → Vite library mode (ESM + CJS output)
- react-scripts test → Vitest with jsdom
- Convert CRA-style SVG ReactComponent imports to TSX components
- Add vite-plugin-dts for TypeScript declaration generation
- Update tsconfig to es2018/react-jsx
- Fix broken ExampleComponent test with valid smoke tests

PR #595, opened 2026-02-16
React 19 Migration 

isacoder:
## Description

This PR migrates scorer-ui-kit from **React 17 → React 19** with full TypeScript support across all three packages (`ui-lib`, `example`, `storybook`).

The Beta release is now available: **3.0.0-beta.1**

This migration was made possible in large part thanks to the work done by @JoshLipps in [Migrate ui-lib build from microbundle-crl/react-scripts to Vite/Vitest](https://github.com/future-standard/scorer-ui-kit/pull/594).

The migration was carried out in structured phases to manage risk:

1. **Build system modernisation** — replaced the outdated `microbundle-crl` / `react-scripts` / CRA toolchain with **Vite** (library mode for `ui-lib`, app mode for `example`) and **Vitest**. Storybook was migrated from `@storybook/react-webpack5` to `@storybook/react-vite`.

2. **Dependency upgrades** — all packages with transitive `@types/react@17` conflicts were upgraded before touching the types:
   - `styled-components` v5 → v6
   - `react-router-dom` v5 → v6
   - `@future-standard/icons` v2 → v3
   - `react-i18next` v11 → v14 + `i18next` v21 → v23

3. **Type system upgrade** — `@types/react@18` was used as a bridge during initial runtime work, then all three packages were upgraded atomically to `@types/react@19` with npm root-level overrides to prevent duplicate versions in the hoisted workspace.

4. **React 19 compatibility** — resolved 33 TypeScript errors surfaced by the new types (theme property renames, `useRef` without initial values, `formAction` type widening, legacy JSX namespace) and applied systematic workarounds for the outstanding `styled-components` v6 / React 19 type incompatibility across 38 files.

5. **Runtime safety audit & fixes** — a full `useEffect` sweep (45 blocks, 27 files) identified and fixed two P1 runtime issues introduced by React 16→17 async cleanup timing changes:
   - `usePoll.ts`: `canceled` ref was not reset on StrictMode remount, silently killing the poll chain
   - `useTitle.tsx`: async title-restore cleanup raced the incoming page's title effect; fixed with `useLayoutEffect`

---

**1. styled-components v6 — transient props added when required but all changes internal impact**
All props passed exclusively to styled components that should not reach the DOM were renamed with a `$` prefix (e.g. `isOutline` → `$isOutline`, `loading` → `$loading`). Any consuming application that was passing these props by name will need to update to the new names. 

**2. styled-components + React 19 `formAction` workaround**
React 19 widened the `formAction` type to `string | ((formData: FormData) => void)`. styled-components v6 doesn't handle the function variant, so 38 files apply the guard `formAction={typeof formAction === 'string' ? formAction : undefined}`. This is safe — the function form is a React 19 server action feature not used in this library. The workaround should be removed once styled-components ships updated types.

**3. React Router v6 — breaking API changes**
`<Switch>` → `<Routes>`, `component=` → `element=`, `useHistory` → `useNavigate`, `RouteComponentProps` → `useParams`. Consumers of the `example` app routing pattern will need to follow the same migration.

**4. `forwardRef` deprecated (not yet removed)**
Two public API components still use `forwardRef`: `FilterDropHandler` (exposes `FilterDropHandlerRef`) and `SplitLayout` (exposes `ISplitLayoutHandles`). Both are functional in React 19 with a console deprecation warning in development. Migration to the React 19 `ref`-as-prop pattern is planned before React 20. 

**5. SVG imports**
CRA's SVGR `{ ReactComponent as X }` pattern is not supported by Vite out of the box. Affected SVGs were converted to hand-authored TSX components. The `vite-env.d.ts` in `example` covers `*.svg`, `*.svg?url`, and `*.svg?react` module declarations for future use.

**6. Deferred P4 fixes (non-blocking, documented)**
Two timer-related issues were found during audit but are deferred — they do not affect production correctness:
- `Notification.tsx`: `setTimeout` not cleared in cleanup (wastes a background timer per StrictMode cycle)
- `WebRTCClient.tsx`: reconnect timer not cancellable on cleanup (production risk on unmount mid-reconnect)
- Some other concerns were raised during audit, this has been documented and will be followed up in [Issue 597](https://github.com/future-standard/scorer-ui-kit/issues/597)

## Known Issues

- A recent theme update introduced animation issues that require follow-up.
- Notifications open and close too quickly.
- The following animations need review and likely adjustments:

  - `packages/ui-lib/src/Tables/atoms/TableRowThumbnail.tsx`
  - `packages/ui-lib/src/Tables/atoms/TypeTableCell.tsx`
  - `packages/ui-lib/src/Tabs/atoms/MobileTab.tsx`

I'm attaching the React Migration plan that was executed in case you need more information for this review

[REACT19_MIGRATION_PLAN-2.md](https://github.com/user-attachments/files/25440302/REACT19_MIGRATION_PLAN-2.md)

---

#### ui-lib
- `packages/ui-lib/package.json`
- `packages/ui-lib/tsconfig.build.json`
- `packages/ui-lib/tsconfig.json`
- `packages/ui-lib/vite.config.ts`
- `packages/ui-lib/vitest.config.ts`

#### example
- `packages/example/package.json`
- `packages/example/tsconfig.json`
- `packages/example/vite.config.ts`

#### storybook
- `packages/storybook/package.json`
- `packages/storybook/tsconfig.json`

### Files Requiring Additional Verification in Real Projects

The following files should receive extra validation in real project environments to ensure no unintended side effects were introduced:

- `packages/ui-lib/src/hooks/useInterval.ts`
- `packages/ui-lib/src/hooks/useTo.tsx`
- `packages/ui-lib/src/Layouts/molecules/SplitLayout.tsx`
- `packages/ui-lib/src/LineUI/LineUIRTC.tsx`

## Testing Steps

**Build verification**
```bash
npm install
npm run build -w packages/ui-lib       # expect: clean build ~8s
npm run build -w packages/example      # expect: clean build ~2s
npm run build-storybook -w packages/storybook  # expect: clean build ~16s
```

**Type check**
```bash
cd packages/ui-lib && npx tsc --noEmit      # expect: 0 errors
cd ../example && npx tsc --noEmit           # pre-existing errors unrelated to this PR
cd ../storybook && npx tsc --noEmit         # pre-existing errors unrelated to this PR
```

**Browser verification (P2 — manual)**

| Scenario | Component | What to check |
|----------|-----------|---------------|
| Click outside to close | Filter dropdown (`FilterDropHandler`) | Closes correctly; no double-close or missed-close |
| Focus/blur commit | `Pagination` page-number input | Value commits on tab-out |
| Drag-resize then navigate | `SplitLayout` | No React "setState on unmounted component" warnings |
| WebRTC in dev mode | `WebRTCClient` | No stale connections or spurious errors from StrictMode double-mount |
| Send a notification | `NotificationContext` | Notification appears exactly once (not twice) in dev mode |
| Styled-components form props | `Button`, `Input`, `Form`, `Switch` | No DOM prop warning for `$isOutline`, `$loading`, etc. |
| Document title on navigation | Any page using `useTitle` | Title updates correctly; no title from previous page persisting |
| Polling hook | Any component using `usePoll` | Polling continues running after initial mount (no silent stop in dev mode) |

**Storybook**
```bash
npm run storybook -w packages/storybook
```
Browse all component stories and verify no visual regressions.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -37,3 +37,15 @@ jobs:
 
     - name: eslint
       run: npm run test:lint -w packages/ui-lib
+
+  lint-status:
+    name: lint
+    needs: lint
+    runs-on: ubuntu-latest
+    if: always()
+    steps:
+      - name: Check lint results
+        run: |
+          if [ "${{ needs.lint.result }}" != "success" ]; then
+            exit 1
+          fi


packages/ui-lib/src/Filters/molecules/FilterLayout.tsx
atomworks: Same as previous comment.

packages/ui-lib/src/Tabs/atoms/MobileTab.tsx
atomworks: Same as previous message regarding this

isacoder:
> Looks good on the whole with a few small things which I'll summarise below!
> 
> 1. There was easing in the legacy theme that got renamed but didn't need to be. I can't find in the changed files that variable being renamed, so I guess these were broken and this is why your PR message said animations might need review. Either way, if we're going to fix these, it's better to just changed them for the CSS variable that replaces it. If the theme was renamed and I missed it, it's better to revert to avoid a breaking change that really isn't needed. _I commented all the ones I saw regarding this by the way, sorry for the noise._
> 2. I'm seeing in a couple of places `input` fields that have `formAction` but these aren't valid for those input types.
> 3. There is a change in `usePoll` which seems like a bug fix but not part of the planned work, so I'm hesitant about it unless it's been fully tested.
> 
> Other than that, all good! Thanks for your diligent work on this.

Hi @atomworks,

Thanks for the quick review. I’ve updated the PR and responded to your comments below.

1. Theme easing variables
Your analysis is correct. The variable replacements were made because the original values did not exist, and TypeScript was flagging them as errors. 
As mentioned in the PR description, I’ve opened a follow-up issue to properly address the theme inconsistencies in a separate PR. I agree that updating to the correct CSS variables is the better long-term approach rather than just patching the values. I’ve now committed the changes as requested.
2. formAction on inputs
The prop was destructured because Styled Components does not handle it automatically, and I initially added a workaround to allow it to be passed through if needed. My understanding is that formAction is primarily used for form submissions and can be relevant in server-rendered environments (e.g., Next.js), which is why I considered preserving it.
That said, I don’t have a strong preference for keeping it, so I’ve removed the formAction passing from the affected components.
3. usePoll change
Fully agree. I’ve reverted the change and will follow up separately with proper testing in a dedicated PR.

Thanks again for the careful review.

If you have any other change requests, please let me know. Otherwise, I’m happy to proceed with the new beta release process.

packages/ui-lib/src/hooks/usePoll.ts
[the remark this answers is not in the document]
JoshLipps: i tested this is seems okay to me

PR #599, opened 2026-02-24
Fix React 19 compatibility bugs in WebRTC and LineUI components

JoshLipps:
## Summary

Post-React 19 migration bugfixes for WebRTCClient, HandleUnit, LineUnit, and LineSet components. These address issues caused by React 19's StrictMode double-mount behavior, stricter lifecycle semantics, and PR review feedback.

### WebRTCClient — React 19 compat and timer leaks (`0ac978b`)
- Added `mountedRef` guard for StrictMode double-mount safety
- Track reconnect timeout and hello interval in refs, clear on cleanup
- Changed `connectionAttempts` from `useState` to `useRef` (non-render control flow)
- Removed misleading `await` on synchronous WebSocket/RTCPeerConnection methods
- Made `closeWebSocket` and `closePeerConnection` synchronous

### LineUI — Leaked window listeners and HandleUnit createRef (`8a70981`)
- **HandleUnit**: Replaced `React.createRef()` with `useRef()` — `createRef` creates a new ref every render in function components
- **HandleUnit**: Fixed touch target comparison to use `handleInstance.current`
- **HandleUnit + LineUnit**: Track `window.addEventListener('mousemove'/'mouseup')` listeners in refs with `useEffect` cleanup to prevent leaks when component unmounts mid-drag (StrictMode remount)

### WebRTCClient — Stale WebSocket event race condition (`50c61af`)
- Ignore `close`/`error` events from previous WebSocket connections by checking `event.target !== webSocket.current`
- Prevents StrictMode cleanup of ws1 from disrupting ws2's live connection
- Added proper TypeScript types for event handler params (`CloseEvent`, `Event`)

#### reconnectTimeoutRef race condition fix (`8f99b9d`)
- Clear existing reconnect timeout before reassigning in `onServerClose` to prevent duplicate concurrent `connectToPeer` calls during rapid reconnect cycles

#### LineUnit useCallback + proper types (`2b6dac2`)
- Wrapped `handleMouseMove`, `handleMouseUp`, `handleMouseDown` in `useCallback` with appropriate deps, matching HandleUnit pattern
- Replaced `any` types with `MouseEvent` / `React.MouseEvent` for consistency

#### Stale `enabled` closure fix (`66a1e8e`)
- Added `enabledRef` that syncs with the `enabled` prop so `onServerClose` always reads the current value instead of a stale closure capture

#### Line boundary clamping on load (`d04259b`)
- Clamp out-of-bounds line points to video dimensions when boundaries first become valid (video metadata loaded)
- Uses a ref to ensure clamping only runs once, preventing data corruption on window resize

#### Scroll spacer test tool (`715a52f`)
- Added toggle button and height slider to LineRTC example page for testing drag accuracy at different scroll positions

## Test plan
- [x] Build ui-lib: `cd packages/ui-lib && npm run build`
- [x] Start example app: `cd packages/example && npm start`
- [x] Verify `/line` page — lines render, drag handles work, no functional console errors
- [x] Verify `/linertc` page with live WebRTC connection — video streams, line overlay renders correctly
- [x] Line points clamped to video boundaries on initial load (e.g. x:2293 → x:1920)
- [x] Points remain stable on window resize (no data corruption)
- [x] Drag handles respect boundaries during interaction
- [x] Scroll position does not affect drag accuracy
- [x] StrictMode double-mount: mount → cleanup → remount, stale ws1 close event ignored
- [x] No errors after navigating away (no leaked listeners/timers)

### Known pre-existing issues (not in scope)
- styled-components transient prop warnings throughout — need `$` prefix
- "Open Video Modal" on `/linertc` is blank — passes `ws://` URL to `<video src>` instead of reusing the MediaStream via `srcObject`

### DIFF
--- a/packages/example/src/pages/LineRTCPage.tsx
+++ b/packages/example/src/pages/LineRTCPage.tsx
@@ -35,6 +35,8 @@ const Line: React.FC<{}> = () => {
   const [ws, setWS] = useState('localhost/wsapp');
   const [wsURL, setWsURL] = useState('');
   const [isShowDirection, setShowDirection] = useState<boolean>(false);
+  const [showSpacer, setShowSpacer] = useState<boolean>(false);
+  const [spacerHeight, setSpacerHeight] = useState<number>(400);
   const [isLabelShadow, setLabelShadow] = useState<boolean>(false);
   const {createMediaModal} = useMediaModal();
 
@@ -216,6 +218,14 @@ const Line: React.FC<{}> = () => {
       </Sidebar>
       <Content>
         {error && <AlertBar message={error} type='error' /> }
+        <div style={{ display: 'flex', alignItems: 'center', gap: '10px', marginBottom: '10px' }}>
+          <Button design="secondary" onClick={() => setShowSpacer(s => !s)}>{showSpacer ? 'Hide' : 'Show'} Scroll Spacer</Button>
+          {showSpacer && <input type="range" min={100} max={2000} step={50} value={spacerHeight} onChange={e => setSpacerHeight(Number(e.target.value))} style={{ width: '200px' }} />}
+          {showSpacer && <span style={{ color: '#888' }}>{spacerHeight}px</span>}
+        </div>
+        {showSpacer && <div style={{ height: `${spacerHeight}px`, background: 'repeating-linear-gradient(45deg, transparent, transparent 10px, rgba(255,255,255,0.03) 10px, rgba(255,255,255,0.03) 20px)' }}>
+          <p style={{ padding: '20px', color: '#888' }}>TEMP: Scroll test spacer</p>
+        </div>}
         <TextField label='Host' name='host' fieldState='default' value={ws} onChange={({target:{value}})=> setWS(value)} ></TextField>
         <Button onClick={connect}>Connect</Button>
         {


isacoder (approved):
Thanks for this update.

I was able to use the LineUI and the LineRTC page without issues, I consider this PR not only a compatibility fix for React19 but a necessary fix since current implementation was not functional. I consider current state approved.

I co-reviewed this PR with AI using this [code-review skill ](https://skills.sh/supercent-io/skills-template/code-review) and it made some recommendations, fell free to add them or we can defer this to a different PR:

-  LineUnit to be updated like it was done to the HandleUnit.
- reconnectTimeoutRef issue that is a low probability edge case but if there is a fix to it might be worth to add it.
- A stale value fix of a pre-existing issue.

 🟡 Important
  I1 — reconnectTimeoutRef overwritten without clearing old timeout

  - File: packages/ui-lib/src/WebRTCClient.tsx — onServerClose
  - Problem:
  reconnectTimeoutRef.current = setTimeout(connectToPeer, 3000);
  - If onServerClose fires a second time before the 3-second timeout elapses (e.g., rapid reconnect cycle), the old setTimeout handle is overwritten without being cleared. The original timer still fires.
  - Why it matters: Could result in two concurrent connectToPeer calls. The connectionAttemptsRef guard limits total attempts, but it could create two simultaneous WebSocket connections.
  - Fix:
  if (reconnectTimeoutRef.current !== null) {
    clearTimeout(reconnectTimeoutRef.current);
  }
  reconnectTimeoutRef.current = setTimeout(connectToPeer, 3000);

  ---
  🟢 Nice-to-have

  N1 — Inconsistent useCallback usage between HandleUnit and LineUnit

  - File: packages/ui-lib/src/LineUI/LineUnit.tsx
  - Problem: handleMouseMove, handleMouseUp, handleMouseDown in LineUnit are plain functions (no useCallback), while the equivalent handlers in HandleUnit are all useCallback-wrapped. The ref-based cleanup pattern works correctly in both cases, but the inconsistency makes LineUnit
   slightly harder to reason about.
  - Fix: Wrap all three in useCallback with appropriate deps, matching the HandleUnit pattern.

  ---
  N2 — any types remain in LineUnit mouse handlers

  - File: packages/ui-lib/src/LineUI/LineUnit.tsx
  - Problem: handleMouseMove(e: any), handleMouseUp(e: any), handleMouseDown(e: any). HandleUnit correctly types its handlers as MouseEvent / React.MouseEvent<SVGSVGElement, MouseEvent>.
  - Fix: Type as MouseEvent for the window-level handlers and React.MouseEvent<SVGGElement> (or the appropriate SVG element type) for handleMouseDown.

  ---
  N3 — enabled stale closure in reconnect path

  - File: packages/ui-lib/src/WebRTCClient.tsx — onServerClose, connectToPeer
  - Problem: enabled is captured by closure when onServerClose and connectToPeer are defined. If the enabled prop changes while a socket is open, these handlers use the stale value.
  - Fix: const enabledRef = useRef(enabled); useEffect(() => { enabledRef.current = enabled; }, [enabled]); then reference enabledRef.current in the handlers. Acceptable to defer — this is a pre-existing pattern in the component.

PR #611, opened 2026-03-24
Fix styled-components transient props and reduce console noise for React 19

JoshLipps:
## Summary
- Adds `$` prefix to all custom styled-component props across `packages/ui-lib/src/` to prevent them from leaking to the DOM in React 19
- Fixes ~173 styled component definitions across 56+ files in 12 component groups
- Preserves valid DOM attributes (`disabled` on `<input>`, `readOnly` on `<input>`, `r` on `<circle>`) without prefixing
- No public API changes — the `$` prefix is applied internally when components pass props to their styled components
- Resolves additional console warnings/errors across the example app and storybook

## Console noise eliminated
- **Transient prop warnings**: All `React does not recognize the X prop on a DOM element` warnings from styled-components
- **Deprecation warnings**: Replaced `design='danger'` with `design='warning'` in example app
- **SVG import error**: Fixed login page SVG `createElement` error by using `?react` import suffix
- **Empty src warning**: Fixed `TableRowThumbnail` rendering `` and PTZ page ``
- **Controlled input warnings**: Added `onChange` handlers to PTZ page form fields
- **Font 404s**: Fixed login SVG font paths for Vite base route
- **Prop leaks**: Fixed `customComponent` leak in MobileTab, `design`/`size` in Button, `position` in Switch
- **console.log cleanup**: Downgraded noisy `console.log` calls to `console.debug` in ui-lib and storybook; upgraded `useLocalStorage` catch handlers to `console.warn`/`console.error` with context messages
- **FilterInputs prop leaks**: Stripped `canHide`, `id`, `showFieldText`, `selected` from spreads onto BasicSearchInput, FilterDropdown, and DropdownDatePicker
- **Storybook fixes**: Fixed transient props in story-local styled-components (CameraPanels, ToggleButton, Spinner, Fonts, SelectField, LoginModalExample), fixed broken placeholder image path, switched http:// URLs to https://

## Component groups fixed
PTZControl, Alerts, Icons, Indicators, Tabs, Pages, Modals, CameraPanels, Misc, Global, Tables, Filters, Form, LineUI

## Test plan
- [x] `npm run build` passes with no errors
- [x] Lint passes
- [x] All 15 example app pages verified clean in browser console
- [x] All 61 storybook stories scanned for transient prop warnings — zero warnings
- [x] Interactive testing of FilterInputs (Show More button, dropdowns) — no prop leaks


packages/ui-lib/src/hooks/useLocalStorage.tsx
```diff
@@ -36,7 +36,7 @@ export function useLocalStorage<T>(key: string, initialValue: T) {
      
     } catch (error) {
       // A more advanced implementation would handle the error case
-      console.log(error);
+      console.debug(error);
     }
   },[key]);
```
isacoder: Maybe keeping this one is not so bad, we use this one a lot and errors on this I think are important to catch/see.

isacoder (approved):
I think there might still be some outputs with warnings that we might not caught in this moment, but I think is valid to caught them and fix them on a different PR.

SelectField for example is showing some warnings.

Feel free to merge when you are ready to make a new Beta.

packages/ui-lib/src/hooks/useLocalStorage.tsx
JoshLipps: It's still there just need to enable debug on chrome dev tools

JoshLipps: maybe it should be an error tho

JoshLipps: fixed

packages/ui-lib/src/Form/atoms/SliderInput.tsx
```diff
@@ -248,6 +248,7 @@ const SliderInput : React.FC<ISlider> = ({
   inputCallback = () => {},
   onChangeCallback = () => {},
   allMarkCentered = false,
+  showValue: _showValue,
```
isacoder, line 251: I realized is not used, I don't remember why it was created, but since it's a custom prop, maybe removing it from the interface is the best call (just check the other components that depend on this Component)

isacoder (approved):
I have read the new code and found nothing breaking. It's approved

packages/ui-lib/src/Form/atoms/SliderInput.tsx
JoshLipps: just to keep the ts interfaces same as before i will leave it for now

PR #612, opened 2026-03-24
Export ThemeType from scorer-ui-kit

JoshLipps:
## Summary
- Exports `ThemeType` (derived from `typeof defaultTheme`) from the package
- Consumers can now import `ThemeType` directly instead of using `typeof defaultTheme`
- No runtime changes — type-only export

## Usage
```typescript
import { ThemeType } from 'scorer-ui-kit';
```

## Test plan
- [x] `npm run build` passes cleanly
- [x] Verify `ThemeType` is available in consuming projects via autocomplete


isacoder (approved):
Works on local, thanks for the fix.

PR #614, opened 2026-03-31
Fix Notification after react19 upgrade - Blocker medium

isacoder:
## Description

This PR fixes a set of React 19 / React StrictMode compatibility bugs in the notification system (`Notification.tsx` and `NotificationContext.tsx`).

**What does this PR do?**

This is a bug fix. React 18 extended StrictMode to double-invoke effects (mount → cleanup → remount) to surface latent bugs. This change was not present in React 16/17, so migrating to React 19 exposed two previously hidden issues:

1. **Module-level notification queue (`NotificationContext.tsx`)** — `notificationList` was declared as a module-level array, shared across all instances and surviving StrictMode's simulated unmount/remount cycle. This caused `closeCallback` to fire twice per dismissed notification. Fixed by moving the array inside the provider component as a `useRef`.

2. **Slide-in double-flash (`Notification.tsx`)** — The slide-in used a CSS `@keyframes` animation (`initAnimation`). CSS keyframe animations restart every time a DOM element is created. StrictMode creates the element twice (mount → cleanup → remount), causing the notification to flash twice on entry. Fixed by replacing the keyframe with a CSS transition gated behind a `requestAnimationFrame`. StrictMode's cleanup is synchronous, so the rAF is always cancelled during the simulated unmount and only fires on the stable remount — a single, clean slide-in.

3. **Slide-out animation inconsistency** — With the slide-in now using a CSS transition, the close path retained a CSS keyframe animation which caused a one-frame position jump when switching CSS blocks. Both directions are now pure CSS transitions with the same easing and duration, ensuring the `transition` property is always present when visible and never changes at the moment the `transform` does.

**Additional improvements:**
- `isClosing` / `isVisible` renamed to `$isClosing` / `$isVisible` (styled-components transient props) to prevent non-HTML attributes from being forwarded to the DOM.
- `ReactDom` → `ReactDOM` import casing corrected.
- `Boolean` (boxed) → `boolean` (primitive) in styled-component prop types.
- Auto-dismiss timer pattern updated from a `mounted` flag to `clearTimeout` — the correct React cleanup pattern.
- Unmount cleanup effect added: if the component is force-removed while the close transition is in-flight, `closeCallback` and `onTextButtonClick` are still invoked.
- Storybook `preview.tsx` updated to wrap stories in `<React.StrictMode>` (placed after `ThemeVariables`/`Style` to avoid CSS variable teardown during cleanup).
- Storybook Notification story updated to cycle through all 5 alert types on each send click, making queue behaviour and animation quality easy to verify.

Animation in close button has areas of improvement. I think is good enough for this PR but if you have any suggestions I'm very open to update.

[  Notifications v1](https://drive.google.com/file/d/10q7A2wNhpcYVIut_cN1CO7-1nwuwJ3a9/view?usp=drive_link)
[Notifications Broken](https://drive.google.com/file/d/1u0zTfyOjaddcO3jNuAOfYcQ12kwCzlN4/view?usp=drive_link)
Notifications Broken currently on storybook in https://future-standard.github.io/scorer-ui-kit/storybook/?path=/story/alerts-atoms--notification
[Notifications Fixed in this PR](https://drive.google.com/file/d/1RUN_W9XTPmeNtS4BRuy7bSQBQvrRFBhl/view?usp=drive_link)

- **`requestAnimationFrame` + `useRef` StrictMode gate:** The rAF is cancelled synchronously by StrictMode's cleanup before it fires. On the stable remount the rAF fires and sets `isVisible = true`, triggering the CSS transition. `animatedIdRef` distinguishes a genuine new notification (different `id` → reset `isVisible`) from a StrictMode remount (same `id` → skip reset, let rAF run).

- **Transition always present when visible:** The `transition` CSS property is applied in a block gated only on `$isVisible`. When `$isClosing` becomes true, `transition` stays unchanged and only the `transform` override is removed — the browser sees a clean property change on an already-transitioned property and animates it smoothly.

- **`closeStateRef` unmount cleanup:** Uses the "latest ref" pattern (no-deps `useEffect` keeps ref in sync after every committed render). The mount-only cleanup reads the ref so it always has the latest state values at unmount time, regardless of when the parent removes the component.

## Reviewing / Testing Steps

**Setup:** Run Storybook (`packages/storybook`) and navigate to **Alerts → atoms → Notification**.

> All visual tests below should be performed with Storybook running — `<React.StrictMode>` is now active in the preview, so StrictMode bugs are directly visible.

1. **Slide-in (no double flash)**
   Click **Send notification** once. The bar should slide down smoothly from the top in a single motion. No flicker or double animation.

2. **Close button (smooth slide-out)**
   Click the × button. The bar should slide back up smoothly over ~0.7 s with the same easing as the slide-in. No jump or snap at the start of the exit.

3. **Auto-dismiss**
   Send a notification without pinning. It should auto-dismiss after 7 seconds with the same smooth slide-out.

4. **Queue behaviour**
   Click **Send notification** 4–5 times rapidly. Each variant (info → success → warning → error → neutral) should appear and dismiss in order. No duplicates, no skipped entries.

5. **`closeCallback` fires exactly once**
   Check the **Actions** panel. Each dismissed notification should log exactly **one** `closeCallback` call. Two calls = the module-level queue bug has regressed.

6. **Clear Notifications**
   While a notification is visible (and more are queued), click **Clear Notifications**. The current notification should slide out and no further ones should appear.

7. **Pinned notification**
   Enable the **Is Pinned** knob, send a notification. It should not auto-dismiss. Close manually — one `closeCallback` action should be logged.


JoshLipps (approved):
looks good and works

PR #616, opened 2026-04-01
Filterbar effect dep fix - 3.0 non blocking

isacoder:
### Description

**What does this PR do?**

This is a bug fix. It resolves a stale dependency in `FilterBar`'s translation-sync `useEffect` that caused selected filter item labels to not update when `dropdownsConfig` changed after mount (e.g. on language switch).

**The bug**

`FilterBar` maintains an internal `filtersValues` state that tracks the text label of each selected filter item. A `useEffect` is responsible for syncing those labels whenever the dropdown config changes — for example, when the parent app switches locale and provides a new `dropdownsConfig` with translated strings.

The effect's dependency array was `[dropdownsConfigRef.current]`. React does not track mutations to `ref.current`, so in practice the effect relied on a render-time ref mutation (`isequal` guard on lines 317–319) to indirectly update the dep value. This is an anti-pattern — the `eslint-disable-next-line react-hooks/exhaustive-deps` suppression on the dep array confirmed the linter flagged it. Additionally, the effect body contained a dead `mountConfig` boolean flag that was always `true` when checked (the cleanup set it to `false`, but only after the body had already executed).

**Background**

This bug was not introduced by the React 19 migration -  it pre-dates it. It was surfaced during the systematic effect-pattern audit carried out as part of the v3.0 / React 19 upgrade (`REACT_17_CODE_AUDIT_AND_FIX_PLAN.md §3.4`, flagged P3 — "human verification required"). The audit reviewed all `useEffect` dependencies across the library for patterns that are unsafe under React 18+ strict rules: stale closures, ref.current deps, missing cleanups, and dead guard flags. This item was tracked in `blockers_uikit.3.0.md` as Blocker 4 and carried into `blockers_3.0_fix_plan.md` as PR C, required before v3.0.0 stable.

**Demo page**

A `FilterBar` with a language-switchable Status dropdown was added to the existing `TablePage` example (`/table`). It uses `useTranslation('Common')` with new `filterBar.*` keys in both `en/Common.json` and `ja/Common.json`. The dropdown config is built via `useMemo([t])` - a stable reference that only rebuilds on language change mirroring the real-world consumer pattern. Selecting a filter value then switching language demonstrates the fix: the active filter tag label updates to the new locale's text.

---

Two changes to `FilterBar.tsx`:
- Replace `[dropdownsConfigRef.current]` with `[dropdownsConfig]` as the effect dependency — the actual prop value React can track
- Remove the dead `mountConfig` flag and its cleanup, and the `eslint-disable` suppression
- Add an `if (dropdownsConfig.length === 0) return` early exit to skip unnecessary state updates when no dropdowns are configured

- **`[dropdownsConfig]` dep and consumer memoization** - With the new dep, the effect fires whenever `dropdownsConfig` is a new reference. Consumers who construct their config inline without `useMemo` will trigger the effect on every render. This is not a correctness issue (no infinite loop, no wrong state), but consumers who care about performance should memoize their config. The demo page uses `useMemo([t])` as a reference example of the correct pattern.

- **The `isequal` ref guard (lines 317–319) remains** - This render-time ref update was the mechanism that previously made the broken dep array accidentally work. It is not removed in this PR - it may be in use by other effects or future code. Removing it is out of scope.

---

### Reviewing / Testing Steps

**To verify the core fix (`FilterBar.tsx`)**

Start the example app in development mode:

```bash
npm start -w packages/ui-lib   # terminal 1
npm start -w packages/example  # terminal 2
```

Navigate to `/table`.

1. Open the **Status** dropdown → select **Error** → click **Apply**
2. Confirm the filter tag reads **"Error"** and the table shows 1 row
3. Click **"Switch to Japanese"**
4. Confirm the filter tag updates to **"エラー"** — all FilterBar UI strings translate simultaneously
5. Click **"英語に切替"** — confirm tag reverts to **"Error"**
6. Click **Clear All** — confirm all 5 rows return

### Screenshot
Screenshot 2026-04-01 at 18 34 51
Screenshot 2026-04-01 at 18 34 51


JoshLipps:
Merged in main to deploy for PR review

JoshLipps (approved):
Lgtm works well thanks

PR #632, opened 2026-04-13
Fix styled buttons class issues

### LINKED ISSUE #626: styled(Button) overwrites internal className in v3 beta 6
## Description

When wrapping scorer-ui-kit's `Button` (or `ButtonWithIcon`) with styled-components' `styled()`, the internal `className` containing `button-design-${design}` and `button-size-${size}` is overwritten by the className injected by styled-components.

This causes buttons to lose their design variant styling (e.g. `secondary` renders as `primary`).

## Root Cause

In the Button component, `className` is set before `{...props}`:

```jsx
<StyledButton className={`button-design-${design} button-size-${size}`} {...props}>
```

When `styled(Button)` wraps it, styled-components passes its own `className` via props. Since `{...props}` comes after the hardcoded `className`, it overwrites it.

## Reproduction

```tsx
import { Button } from 'scorer-ui-kit';
import styled from 'styled-components';

const CancelButton = styled(Button)`
  margin-right: 10px;
`;

// This loses button-design-secondary class:
<CancelButton design='secondary'>Cancel</CancelButton>

// This works fine:
<Button design='secondary'>Cancel</Button>
```

## Suggested Fix

Merge the incoming className with the internal one:

```jsx
<StyledButton className={`button-design-${design} button-size-${size} ${props.className || ''}`} {...props}>
```

Or move `{...props}` before `className`.

## Environment

- scorer-ui-kit: 3.0.0-beta.6
- styled-components: 6.3.12
- React: 19.2.4

isacoder:
## Description

**Bug fix** — resolves [#626](https://github.com/future-standard/scorer-ui-kit/issues/626): `styled(Button)` wrappers lose the button's design variant styling.

When wrapping `Button` with styled-components' `styled()`, the internal `button-design-{design}` and `button-size-{size}` class names were being silently overwritten by the `className` that styled-components injects via props. Because `{...props}` appeared after the explicit `className` in the JSX, the spread won and the design classes were discarded — causing a `secondary` button to render identically to a `primary` button.

**Before (broken):**
```tsx
const CancelButton = styled(Button)`margin-right: 10px;`;
<CancelButton design="secondary">Cancel</CancelButton>
// Rendered class: "sc-xxx sc-generated" — missing button-design-secondary
```

**After (fixed):**
```tsx
<CancelButton design="secondary">Cancel</CancelButton>
// Rendered class: "sc-xxx button-design-secondary button-size-normal sc-generated"
```

**Root cause:** In `Button.tsx`, `className` was set before `{...props}`, so `props.className` (injected by `styled()`) overwrote it.

**Fix:** `className` is now destructured out of `...props` and merged with the internal classes using `filter(Boolean).join(' ')`, so both the design/size classes and any externally passed class are preserved.

A new Storybook story has been added to visually reproduce this fix. It may also be useful as a reference for catching similar issues when using styled(Component) wrappers, though it's can be deleted if you prefer.

- **Why not move `{...props}` before `className`?** That approach restores the internal classes but silently discards the styled-components wrapper's own generated class — making `styled(Button)` functionally useless as an extension point. The merge approach is the only one that preserves both.
- **`filter(Boolean)`** safely drops `undefined` when no external `className` is passed, avoiding trailing spaces or the string `"undefined"` in the class attribute.
- `ButtonWithIcon` passes `{...props}` through to `Button`, so it inherits this fix with no additional changes needed.
- No breaking changes to the public API — the `className` prop was already accepted via `ButtonHTMLAttributes<HTMLButtonElement>`.

## Reviewing / Testing steps

1. **Visual regression in Storybook** — a dedicated story has been added under `Form/Buttons → Styled Wrapper Preserves Design Variant`. It renders three buttons side by side:
   - Plain `Button design="secondary"` (reference)
   - `styled(Button)` wrapper with `design="secondary"` — should visually match the plain secondary
   - Plain `Button design="primary"` (reference)

   Before the fix the middle button renders identically to the primary. After the fix it matches the secondary.

2. **DOM inspection** — open DevTools on the story and confirm the middle button's `class` attribute includes both `button-design-secondary` and the styled-components generated hash class.

3. **Manual smoke test** — wrap `Button` in any `styled()` call and pass any `design` prop. Confirm the variant styling is preserved.

### Before

Screenshot 2026-04-13 at 14 38 20

### After Fix
Screenshot 2026-04-13 at 14 48 29


packages/ui-lib/src/Form/atoms/Button.tsx
```diff
@@ -140,9 +140,10 @@ interface OwnProps {
 
 type Props = OwnProps & ButtonHTMLAttributes<HTMLButtonElement>
 
-const Button : React.FC<Props> = ({ design='primary', size='normal', shadow = false, noPadding = false, loading=false, children, formAction, ...props }) => {
+const Button : React.FC<Props> = ({ design='primary', size='normal', shadow = false, noPadding = false, loading=false, children, formAction, className, ...props }) => {
   design === 'danger' ? console.warn('Button.tsx - Warning, the design prop value danger is being deprecated. Use warning instead.') : null;
-  return <StyledButton type='button' $isOutline={design === 'outline'} className={`button-design-${design} button-size-${size}`} $design={design} $size={size} $noPadding={noPadding} $shadow={shadow} $loading={loading} {...props}><>{children}</></StyledButton>;
+  const mergedClassName = [`button-design-${design}`, `button-size-${size}`, className].filter(Boolean).join(' ');
+  return <StyledButton type='button' $isOutline={design === 'outline'} className={mergedClassName} $design={design} $size={size} $noPadding={noPadding} $shadow={shadow} $loading={loading} {...props}><>{children}</></StyledButton>;
```
JoshLipps, line 146: perfect!

JoshLipps (approved):
Awesome thanks for that. 💯 :shipit: 🎉

atomworks (approved):

PR #638, opened 2026-04-17
SplitLayout - Drag handle and arrow alignment fix

atomworks:
On the horizontal drag handles, in both drag and closed states, the icons were out of place. Uses flex now to position these correctly.

Before: 
image

After:
image


isacoder (approved):
Thanks for fixing the alignment of the hamburger icon, Looks good :)

PR #640, opened 2026-04-21
Add HLS support to LineUIVideo (optional hls.js peer dep)

JoshLipps:
## Summary
- Adds HLS playback to `LineUIVideo` without forcing `hls.js` on every consumer.
- `hls.js` is declared as an **optional** peer dependency (`peerDependenciesMeta.hls.js.optional: true`) and loaded via dynamic `import()`, so the ~230 KB gzipped chunk is only fetched when a consumer actually passes an `.m3u8` source.
- Safari uses native HLS via `canPlayType('application/vnd.apple.mpegurl')`; other browsers fall back to `hls.js` if the consumer has installed it, with a clear console error if not.

## Why this replaces #585
[#585](https://github.com/future-standard/scorer-ui-kit/pull/585) was cut 160 commits behind `main` (pre-monorepo, pre-React 19) and made `hls.js` a hard dependency. This branch is cut fresh from `main`, targets the current `packages/ui-lib` layout, and adds HLS as an optional peer dep instead.

## Notable differences from #585
- Dynamic `import('hls.js')` with a `cancelled` flag so re-renders / StrictMode double-mounts don't leak instances.
- Effect depends only on `[src]` (was `[src, autoPlay]`, which tore down HLS on every autoplay toggle).
- Detection tightened to `/\.m3u8(?:$|[?#])/i` so query-string matches don't false-positive.
- Safari native HLS path is preferred when available; `hls.js` is only loaded when needed.
- Dropped `@types/hls.js` — `hls.js` 1.x ships its own types.

## Test plan
- [x] `npm run test:types` (ui-lib)
- [x] `npm run test:lint` (ui-lib)
- [x] `npm run test:unit` (ui-lib)
- [x] `npm run build` (ui-lib) — confirms `hls.js` is code-split into its own chunk
- [x] Manual: load a non-HLS video URL and verify no `hls.js` chunk is fetched
- [x] Manual: load an `.m3u8` in Chrome and verify playback via `hls.js`
- [x] Manual: load an `.m3u8` in Safari and verify native playback (no `hls.js` chunk fetched)
- [ ] Manual: remove `hls.js` from `node_modules` and verify the console error surfaces on an `.m3u8` src


isacoder (approved):
All is working well, thanks for adding this fix. 
Just a quick question about the versions, is there a reason why the peer version and the development version are different?. They will probably be dedupe to the same but just curious. 

Image

isacoder:
Check in slack with Josh, updated the same versions.

PR #642, opened 2026-04-23
Speed up ui-lib watch builds and add root dev scripts

JoshLipps:
## Summary
- Skip `vite-plugin-dts` in watch mode — ui-lib rebuilds drop from ~10s to ~50ms. Full `npm run build` still emits declarations.
- Externalize `hls.js` (optional peer dep) so Rollup no longer bundles a ~1MB `hls-*.mjs` chunk; dynamic import now resolves from the consumer's `node_modules`.
- Add root-level scripts (`start`, `start:all`, `start:ui-lib`, `start:types`, `start:example`, `start:storybook`) backed by `concurrently` for a cleaner dev flow.
- New `start:types` script in ui-lib runs `tsc --watch` against `tsconfig.build.json` for live `.d.ts` regeneration when needed.

## Test plan
- [x] `npm run build` in `packages/ui-lib` emits `dist/*.d.ts` and no `hls-*.mjs` chunk
- [x] `npm start` in `packages/ui-lib` rebuilds in <500ms per change
- [x] `npm start` from repo root runs ui-lib watcher + example dev server in parallel with prefixed output
- [x] `npm run start:all` adds types watcher and storybook
- [x] `npm run start:types` produces updated `.d.ts` on source edits
- [x] Verify `LineUIVideo` HLS playback still works in an app that has `hls.js` installed, and gracefully degrades when it doesn't


isacoder (changes requested):
Thanks for updating the project run options.

I was able to run on local and I think the usage feels smoth but I’ll wait for Leonard’s input before merging, since he originally reported the build time issue.

I also noticed the HLS dependency was marked as external in this PR. AI flagged that this might break consumers who don’t have HLS installed because LineUIVideo is still part of the Library.


Please take a look -  I’ve attached the AI suggestions for reference.
--

  I reproduced it locally with a minimal Vite 8.0.10 / Rolldown setup that mirrors what a downstream consumer of scorer-ui-kit@3.0.0-beta.7 would build:

  Error: [vite]: Rolldown failed to resolve import "hls.js" from "/private/tmp/vite-hls-test/lib.js".
  This is most likely unintended because it can break your application at runtime.

  Two important findings

  1. Builds fail even when LineUIVideo is never imported. I tested two entries — one that calls loadHls(), one that imports only an unrelated function from the same module. Both
  fail with the same Rolldown error. Tree-shaking happens after resolution, so the unused dynamic import still trips the bundler. Reproduced files at /tmp/vite-hls-test/{lib.js,
  entry-uses-hls.js, entry-no-hls.js}.
  2. The runtime try/catch at LineUIVideo.tsx:246 is unreachable in this scenario. The defensive console.error('[LineUIVideo] HLS playback is not available...') only matters at
  runtime — but Vite 8 / Rolldown blocks the consumer's build before any runtime exists. So the "graceful fallback" is theoretical for the v8 audience, which is exactly the
  audience this repo just migrated to (per the project memory on the Vite 8 dedupe fix).



Image

JoshLipps:
@isacoder thanks for the careful review — the AI repro was right. Pushed bd58903b which restructures the fix.

## What was wrong with my first pass

Marking `hls.js` external in `vite.config.ts` left a literal `import("hls.js")` in our published `dist/index.modern.js`. Rolldown does eager module-graph resolution before tree-shaking, so any downstream Vite 8 / Rolldown consumer would hit `Failed to resolve import "hls.js"` even when `LineUIVideo` was never imported. The runtime try/catch never had a chance to fire.

## The new shape — `scorer-ui-kit/hls` sub-entry

- Main entry (`scorer-ui-kit`) has **zero** hls.js references in its bundle. Verified with `grep`.
- HLS-aware component lives at `scorer-ui-kit/hls` as `LineUIVideoHLS`. Static `import Hls from 'hls.js'` — consumers who use `/hls` install hls.js as a peer dep.
- Shared overlay logic extracted into a private `LineUIVideoBase` so we don't duplicate ~250 lines. Rolldown auto-emits it as a shared chunk, so a consumer using both entries doesn't double-ship the overlay code.
- `peerDependenciesMeta.hls.js.optional: true` is preserved → non-HLS consumers see no install warning. Consumers using `/hls` without hls.js installed get a clear runtime error: `Could not resolve "hls.js" imported by "scorer-ui-kit". Is it installed?` (Rolldown stubs it at consumer-build time, so the build still succeeds — error fires only when navigating to the HLS page.)

## Verification matrix

| Scenario | Build | Runtime |
|---|---|---|
| Consumer never imports `/hls`, no hls.js | ✅ 792 kB consumer bundle, **zero hls.js code** | ✅ all main-entry features work |
| Consumer imports `/hls`, hls.js installed | ✅ 1.3 MB (consumer bundles hls.js) | ✅ video plays via blob MediaSource (1280×720 @ 19 s into BipBop manifest) |
| Consumer imports `/hls`, hls.js missing | ✅ Rolldown emits runtime stub | ❌ on HLS page only — clear "module not found" error |
| Existing `LineUIVideo` with `.mp4` | ✅ unchanged | ✅ unchanged |

## Behavior change to call out

The original `LineUIVideo` used to dynamically `import('hls.js')` if you fed it an `.m3u8` URL on Chrome/Firefox. The refactored `LineUIVideo` no longer does that — it just sets `video.src = src`. Safari's native HLS still works; non-Safari consumers who relied on the auto-fallback need to switch to `LineUIVideoHLS` from `scorer-ui-kit/hls`.

Repro for the "no hls.js, no `/hls` import" case (matches the AI's earlier scenario):

```bash
# In packages/example, comment out LineHLSPage import + route in App.tsx
mv node_modules/hls.js node_modules/hls.js.disabled
npm run build --workspace=packages/example
# ✓ built in 680ms, bundle has zero hls.js code
```

Ready for another look when you have time.

isacoder (approved):
Thank you for the update on the LineUIVideo all looks correct. I will wait for @atomworks  feedback before merging.

PR #636, opened 2026-04-16
Replace ESLint with Biome

isacoder:
## Description

- Replaces ESLint + `globals` with [Biome](https://biomejs.dev/) 2.4.12 as the single tool for linting and formatting across the monorepo
- Deletes `packages/ui-lib/eslint.config.mjs` and removes ESLint/globals devDependencies from `packages/ui-lib/package.json`
- Adds `biome.json` at repo root with lint + format rules equivalent to the previous ESLint setup
- Updates CI (`.github/workflows/lint.yml`) to run `npm run check` instead of the ESLint step; simplifies Node matrix to 22.x only
- Adds `check`, `check:fix`, `format`, `lint`, `lint:fix` scripts to root `package.json`
- Applies Biome formatting and lint fixes across all 311 source files in the monorepo

---

### Rule mapping from ESLint

| ESLint rule | Biome equivalent | Severity |
|---|---|---|
| `no-unused-vars` / `no-unused-imports` | `noUnusedVariables` / `noUnusedImports` | error |
| `react-hooks/exhaustive-deps` | `useExhaustiveDependencies` | warn |
| `react-hooks/rules-of-hooks` | `useHookAtTopLevel` | error |
| `@typescript-eslint/no-explicit-any` | `noExplicitAny` | warn |
| `no-console` | `noConsole` | error (allow `warn`, `error`, `info`, `debug`) |
| `import/consistent-type-specifier-style` | `useImportType` | error |
| jsx-a11y rules | `a11y.recommended` | error |

`noArrayIndexKey` is set to `warn` (not error) — 42 pre-existing usages in the codebase not worth fixing now.

---

### Differences from [future-standard/ui-kit](https://github.com/future-standard/ui-kit) biome.json

The ui-kit repo uses Biome 1.9.4. This config targets Biome 2.4.12 and reflects the breaking API changes between versions, plus two project-specific additions.

**Biome 2.x API changes (not decisions — required for 2.x)**

| Area | ui-kit (1.9.4) | This project (2.4.12) |
|---|---|---|
| Schema | `1.9.4/schema.json` | `2.4.12/schema.json` |
| File exclusions | `files.ignore: [...]` | `files.includes: ["**", "!..."]` |
| Organize imports | `organizeImports.enabled: true` | `assist.actions.source.organizeImports: "on"` |
| Console rule name | `noConsoleLog` | `noConsole` (renamed; now covers all `console.*`) |
| Overrides key | `"include"` | `"includes"` (pluralised) |

[...]

PR #648, opened 2026-04-28
Add story sweep script + PR Action; fix runtime warnings it surfaced

JoshLipps:
Adds `npm run sweep` — a single-command headless playwright pass over every Storybook story + example route — and a GH Action that runs it against the PR's deployed gh-pages preview.

Surfaces and fixes 5 real runtime warnings the sweep caught.

## Final state
- `npm run sweep` against the production build: **0 errors, 0 warnings, 0 timeouts, 0 loops** across all 86 pages (69 stories + 17 example routes).
- `npm run check` clean, `tsc --noEmit` clean, `npm run build -w packages/ui-lib` clean.

## What sweep checks (per page)
- `console.error` events
- `pageerror` (uncaught exceptions)
- `goto` timeouts (loop / hang signal — locked render prevents networkidle)
- Excessive console message count (loop signal — runaway effect re-logging)

Filters out favicon 404 noise. Exits non-zero if any page is flagged.

## Bug fixes surfaced by the sweep

| Component | Issue | Fix |
|---|---|---|
| `UserMenu` (this branch) | submenu items frequently share `href='#'` placeholder, key collapse | `key=\`${text}-${href}\`` composite |
| `PageHeader` (#636) | `key=\`tag-\`` literal — interpolation lost during Biome auto-fix; also labels in real data repeat | `key=\`tag-${index}-${label}\`` + biome-ignore |
| `MediaBox` (pre-existing) | `src=''` on load failure triggered React's empty-string warning | `src={undefined}` (element is hidden anyway, NoImage placeholder shows) |
| `SelectField` (pre-existing) | always set `defaultValue` (with `''` fallback) on the inner `<select>`, conflicting with consumer-passed `value` | only emit `defaultValue` if `value` not in props |
| `FilterLayout` (pre-existing) | passed both `value` and `defaultValue` to `<SelectField>` | drop `defaultValue` (component is controlled) |
| `CropTool` story (pre-existing) | `<CropResult src={cropImg} />` initial empty string | `src={cropImg \|\| undefined}` |

`PageHeader`/`UserMenu` regressions came from key changes in #647. The rest are pre-existing.

## CI workflow

`.github/workflows/pr-sweep.yml` triggers via `workflow_run` after the existing **PR Deploy Preview** job finishes, waits for the gh-pages URL to come up, then sweeps the deployed PR preview. Posts a summary comment ("Clean ✓" or a per-page flag list), uploads the full JSON result as an artifact, and fails the job if any page is flagged.

**Why post-deploy, not dev-server:** the local dev server's HMR + React StrictMode combination triggered transient `createRoot` / `removeChild` warnings on `/` and `/line` that did NOT appear in the production build. Sweeping the deployed gh-pages URL gets the production answer with zero false positives.

## Local usage

```sh
npm run start:all  # runs ui-lib watch, types, example, storybook concurrently
npm run sweep
```

Defaults to `/usr/bin/chromium` (no 150MB Playwright browser download). Override with `PW_CHROMIUM_PATH=/path/to/chrome` or set empty (`PW_CHROMIUM_PATH= npm run sweep`) to use Playwright's bundled chromium.


isacoder (commented):
Changes look good, thanks for adding this sanity check.

The npm run sweep did not work for Mac users, but I have pushed a workaround in this branch. If you have not objections is ready to merge.

Image

isacoder (approved):

PR #657, opened 2026-05-27
RHF PoC 3rd Iteration — integrated Input/SelectField

isacoder:
## Description

This PR lands the **React Hook Form (RHF) integration PoC** for `scorer-ui-kit` across three iterative research rounds, plus a small CI fix and a formatter sweep that landed during the spike. It is a **research / spike PR**: the goal is to validate how RHF could be wired into the existing form atoms (`Input`, `SelectField`, `Checkbox`, `RadioButton`) and to compare three integration styles side-by-side before committing to one as the official UI Kit direction.

### Source of documentation

Three findings reports are committed alongside the code and act as the design memo for each round:

- `rhf-spike-report.md` — Round 1 spike synthesis (FormField molecule).
- `findings-native-atoms.md` — Round 2 per-component change log, awkward parts, a11y notes, sanity-sweep results, recommendations.
- `findings-integrated-rhf.md` — Round 3 React Hook Form integrated to core components.

### TL;DR

The potential outcome is expected to be one of the following:

```
1. UI Kit v3 can work well with React Hook Form with none or minimal changes.
2. UI Kit v3 would need significant breaking changes but could be pushed to work.
3. New components are needed so this work will inform the new UI Kit version work and first targets. (this is what is expected, though we may be surprised)
```

#### Outcome is number 1 - UI Kit v3 can work well with React Hook Form with none or minimal changes.

Each iteration only represents a different place to put the React Hook Form wiring:

Iteration 1 - <Controller /> at the FormField wrapper layer
Iteration 2 - <Controller /> at the consumer level
Iteration 3 - <Controller /> inside the atoms

Areas where reviewer attention would be most valuable:

- **Round 3 dispatcher rule on `Input` / `SelectField`.** `isIntegrated = formContext !== null && props.name !== undefined && props.value === undefined`. The `props.value === undefined` clause is what protects rounds 1 / 2 (and any other consumer doing their own controlled wiring) from being silently hijacked by RHF when they happen to be rendered inside a `<FormProvider>`. Worth a careful read — this is the backward-compat contract.
- **`React.forwardRef` removal.** Both atoms drop `forwardRef` and use `ref` as a regular prop (React 19). A repo-wide sanity sweep is documented in `findings-native-atoms.md`; reviewer may want to confirm no external consumer of the published package relies on the `displayName` that was dropped.
- **`Checkbox` controlled mode.** New `controlled` prop bypasses the internal `useState` and treats `checked` as authoritative — needed for RHF integration to avoid stale-state fights, but it changes the contract for any caller that opts in. Default remains uncontrolled, so existing usage is unaffected.
- **`SelectField` `label` prop is now `ILabel | string`.** Standalone callers still pass the object shape; integrated mode accepts a flat string and synthesises the `ILabel` internally. Worth a quick check that the union doesn't trip downstream TS consumers.
- **`SelectField` ArrowDown handler** (in the accessibility-fix commit) advances past the disabled placeholder option and dispatches a synthetic `change` event. Bears watching if any consumer attaches their own `onKeyDown`; the new handler defers to a caller-provided `onKeyDown` first and only acts if `e.defaultPrevented` is false.
- **`react-hook-form` added to ui-lib `peerDependencies`** and externalised in `vite.config.ts`. Consumers that don't use RHF won't pay the bundle cost, but they will see a new peer-dep warning at install — confirm this is acceptable as a v2.x addition.
- **PoC scope, not production scope.** The three demo pages and reports are intended to ship in the repo as the artifact of the spike; they are not consumer-facing routes in any real deployment. If the team prefers them moved out of `example/`, that should be a follow-up.

### Pre-merge verification

| Check                                                       |  Result   |
|-------------------------------------------------------------|:---------:|
| Lint (`biome check`)                                         | ✅ PASS  |
| Type-check / build (`tsc`, `vite build`)                     | ✅ PASS  |
| Unit / integration tests                                     | ⏭️ N/A   |
| Playwright e2e — round 1 (`rhf-spike.spec.ts`, 14 tests)      | ✅ PASS  |
| Playwright e2e — round 1 keyboard (`rhf-spike-keyboard.spec.ts`) | ✅ PASS  |
| Playwright e2e — round 2 native atoms (`rhf-spike-native.spec.ts`, 8 tests) | ✅ PASS  |
| Playwright e2e — round 3 integrated (`rhf-spike-integrated.spec.ts`, 8 tests) | ✅ PASS  |
| Manual smoke test of `/rhf-spike`, `/rhf-spike-native`, `/rhf-spike-integrated` | ✅ PASS  |
| Console / log clean of new warnings                          | ✅ PASS  |
| Sweep script against live PR preview (86 pages, 0 errors)    | ✅ PASS  |
| `forwardRef` / `defaultProps` / `propTypes` / `UNSAFE_` sanity sweep | ✅ PASS  |
| Docker / deploy artifact builds                              | ⏭️ N/A   |

### Detailed steps to reproduce

**Setup**

```bash
nvm use            # repo's .nvmrc
npm ci             # install at repo root (workspaces)
```

No env vars or external services required.

**Reproduction**

Run each check the table claims:

```bash
# Lint + format + organize-imports (matches CI)
npm run check

# Type-check + library build
npm run build

# Playwright e2e — runs all four suites against the example app on :3101
cd packages/example
npm run test:e2e
```

To exercise the three integration styles by hand:

```bash
# From repo root
npm start
# then visit:
#   http://localhost:3000/rhf-spike              (round 1 — FormField molecule)
#   http://localhost:3000/rhf-spike-native       (round 2 — native label/error on atoms)
#   http://localhost:3000/rhf-spike-integrated   (round 3 — RHF baked into atoms)
```

For each demo page:

1. Click **Submit** with empty fields → required-validation errors render under each field, `aria-invalid="true"` is set, and the focused field receives `aria-describedby` pointing at the error id.
2. Type an invalid email → on blur (`mode: 'onTouched'`), the email-pattern message shows; correcting the value clears it.
3. (Round 1 only) Tab through the form keyboard-only, including the radio group and checkbox group, and submit — focus should be visible at each step and `field.onBlur` should fire (validated by the keyboard suite).
4. With all valid input, **Submit** renders a `<pre data-testid="submitted">` block with the JSON payload.

Screenshot

Screenshot 2026-05-28 at 10 35 01

### DIFF
--- a/findings-integrated-rhf.md
+++ b/findings-integrated-rhf.md
@@ -1,0 +1,305 @@
+# Integrated RHF Atoms — Findings (PoC Round 3)
+
+## TL;DR — verdict on the framing question
+
+The PoC was set up to land on one of three outcomes:
+
+1. **UI Kit v3 can work well with RHF with none or minimal changes.**
+2. **UI Kit v3 would need significant breaking changes but could be pushed to work.**
+3. **New components are needed — this work informs the new UI Kit version.**
+
+**The answer is a split: (1) for v3, (3) for the AI-Native version.**
+
+- **For v3: outcome (1) holds.** All three rounds were achieved on top of v3 atoms without breaking changes — the diff vs `main` is almost entirely additive (new optional props, widened union types, opt-in rendering shells). Round 1's `FormField` molecule is a zero-touch addition. Round 2 added native `label`/`error`/`hint` props as optionals. Round 3 added internal `useController` behind a runtime dispatcher; the only footgun is `<Input name='x' />` placed inside a `<FormProvider>` without `value` silently subscribing to RHF — and there are no such call sites in this repo. **If we wanted to ship RHF support on v3 today, we could.**
+- **The cost of (1) is design debt, which points to (3).** Round 3 works only because of a dispatcher that picks between integrated and standalone modes at runtime, based on an implicit `value === undefined` signal. Two of every field component live in one file (StandaloneInput + IntegratedInput, etc.). `SelectField` has to support both an `ILabel` object label *and* a flat string. Two error-rendering paths (`FormField` + the atom shell) coexist. These compromises exist purely because v3's API is being preserved.
+- **The AI-Native UI Kit should take outcome (3).** Build integrated as the *only* mode, drop the dispatcher, drop `FormField`, drop `ILabel`, drop `children`-based options, use `useId()` everywhere, ship a single `<FieldShell>` primitive. The recommendations section below details this.
+
+**Outcome (2) — "significant breaking changes pushed onto v3" — is not the right framing.** The work proved the breaking changes aren't necessary on v3; the cleaner API is necessary for v4.
+
+See the [Breaking changes vs `main`](#breaking-changes-vs-main) and [Recommendations for the AI-Native UI Kit](#recommendations-for-the-ai-native-ui-kit) sections for the full reasoning.
+
+## Goal
+
+Round 1 (`feature/react-form-hooks-poc`) wired `react-hook-form` (RHF) to the UI Kit via a `FormField` molecule + render-prop. Round 2 (`feature/react-form-hooks-poc-2`) refactored `Input` and `SelectField` to render their own label/error/hint, but the consumer still called `useController` for each field.
+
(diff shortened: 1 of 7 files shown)


isacoder:
This spike research has end. Closing for now we won't make any update in codebase related to React Form Hooks

PR #668, opened 2026-06-23
fix(DatePicker): exhaustive-deps + duplicate weekday keys (#644, #659)

### LINKED ISSUE #644: useExhaustiveDependencies warnings: DateTimeBlock validHourMin and SelectField props rest spread
Three `useExhaustiveDependencies` warnings deferred from the Biome migration (PR #636) — currently suppressed.

## DateTimeBlock.tsx:205, :228

```ts
const setDateHours = useCallback(
  ({ target: { value } }: React.ChangeEvent<HTMLInputElement>) => {
    // ...
    const { newHour, newMin } = validHourMin(value, displayMinutes, hasDate, allowAfterMidnight);
    // ...
  },
  [allowAfterMidnight, date, displayMinutes, hasDate, setDateCallback, validHourMin]
);
```

`validHourMin` is declared inline inside the component body, so its reference changes every render. Including it in the deps array causes `setDateHours` and `setDateMinutes` to be recreated on every render anyway — defeating `useCallback`. Same pattern in `setDateMinutes` at line 208/228.

**Fix:** wrap `validHourMin` itself in `useCallback` (with its own correct deps) so its identity is stable, then the two outer callbacks become genuinely memoised.

## SelectField.tsx:203

```ts
const renderSelect = useCallback(
  (htmlFor?: string) => (
    <StyledSelect ... {...props} ...>...</StyledSelect>
  ),
  [children, defaultValue, handleOnChange, placeholder, props, fieldState, icon, iconColor, isCompact]
);
```

`props` here is the rest object collected from the destructuring of `OwnProps & SelectHTMLAttributes<HTMLSelectElement>`. The rest object is rebuilt by JS on every render, so its identity changes every time and `useCallback` never actually caches.

**Fix options:**
1. Replace `{...props}` with explicit prop forwarding (e.g. list out the HTML attributes the consumer actually relies on) — requires API discipline but is the cleanest.
2. Memoise `props` with `useMemo` keyed on a stable serialisation — messy.
3. Drop the `useCallback` entirely if `renderSelect` is only called in one place per render (it likely is) — least intrusive.

Suppressions live in:
- `packages/ui-lib/src/Filters/atoms/DateTimeBlock.tsx`
- `packages/ui-lib/src/Form/atoms/SelectField.tsx`

Once one of the fixes lands, remove the corresponding `biome-ignore` comments.

### LINKED ISSUE #659: DatePicker: duplicate React keys (`S`/`T`) in weekday header row (en locale)
## Summary

`DatePicker` renders its weekday header row with **non-unique React `key`s**, producing the dev warning *"Encountered two children with the same key"* in English locale (`lang="en"`, the default).

## Version

- `scorer-ui-kit` **3.0.0**
- React 19, styled-components v6

## Location

`src/Filters/molecules/DatePicker.tsx` — the weekday header (`CalHRow` / `CalHCell`). In the published bundle this is the `dayGuide.map(...)` call in `dist/index.modern.js`.

## Root cause

The weekday header cells are keyed by the label string itself:

```jsx
<CalHRow>
  {dayGuide.map((day) => (
    <CalHCell key={day}>{day}</CalHCell>
  ))}
</CalHRow>
```

with

```js
const enDayGuide = ["S", "M", "T", "W", "T", "F", "S"];
```

`"S"` (Sun/Sat) and `"T"` (Tue/Thu) each appear twice, so two pairs of sibling cells share a `key`. React logs, in dev:

```
Encountered two children with the same key, `S`. Keys should be unique ...
Encountered two children with the same key, `T`. Keys should be unique ...
```

`jpDayGuide = ["日","月","火","水","木","金","土"]` is unique, so `lang="ja"` is unaffected — **the warning is English-locale only**.

## Steps to reproduce

1. Render any `DatePicker` with the default `lang="en"` (e.g. indirectly via `FilterBar` `datePickersConfig`).
2. Mount/open the calendar.
3. Observe the two duplicate-key warnings (`S` and `T`) in the console.

## Expected

Weekday header cells should have stable, unique keys; no duplicate-key warning.

## Impact

- Console noise in development.
- Breaks console-clean assertions in consumer E2E suites. We gate our Playwright runs on a zero-console-error / zero-key-warning check, and this warning trips it on every page that mounts a `DatePicker` through `FilterBar`.
- Functionally benign today (the header is static), but a latent correctness risk if the row ever becomes dynamic.

## Suggested fix

Key by the map index rather than the label:

```jsx
{dayGuide.map((day, i) => (
  <CalHCell key={i}>{day}</CalHCell>
))}
```

## How the owner was confirmed

Walking the live React fiber tree, the duplicate-keyed (`S`/`T`) `styled.button` cells have `_debugOwner === DatePicker`, and the keys map exactly to `enDayGuide`'s repeated `S`/`T` entries.

isacoder:
## Description

This PR bundles two small, self-contained **bug fixes** in the Filters/Form atoms that each clear React/Biome warnings surfaced during recent migration and sweep work, plus a related Storybook cleanup that drops a deprecated prop usage. There is no change to public component APIs and no change to rendered behavior — all three are correctness/cleanliness changes.

**What this PR does**

1. **Removes the three deferred `useExhaustiveDependencies` suppressions** from the Biome migration (PR #636) — `fixes #644`.
   - `DateTimeBlock.tsx`: `validHourMin` was declared inline in the component body, so its identity changed every render. Listing it in the `setDateHours` / `setDateMinutes` dependency arrays defeated `useCallback`. Because the function captures **no** closure variables, it is hoisted to module scope, giving it a stable identity, and removed from both deps arrays — the two callbacks now genuinely memoize.
   - `SelectField.tsx`: `renderSelect` was wrapped in `useCallback` whose `props` dependency is the rest object from destructuring, rebuilt on every render — so the cache never held. `renderSelect` is called once per render, so it is converted to a plain inline function (equivalent, no memo needed).
   - All three `biome-ignore` comments are removed.

2. **Fixes duplicate React keys in the DatePicker weekday header row** — `fixes #659`.
   - The weekday guide rendered `<CalHCell key={day}>` using the label as the key, but labels repeat within a week (EN: `S`/`S` for Sun+Sat, `T`/`T` for Tue+Thu), producing React duplicate-key warnings. Each guide entry is now modeled as `IDayGuide { key, label }` with stable unique day keys (`sun…sat`); cells are keyed on `key` and render `label`.

3. **Migrates the SelectField story off the deprecated `label.isSameRow`** to `label.direction: 'row'`. The story set `isSameRow: true`, which tripped `SelectField`'s own deprecation `console.warn` on every render; the replacement API gives the same layout with no component change. This is a story-only cleanup (no public API change) that clears the warning we observed during verification.

**Type:** Bug fix (warning cleanup). **Not** a feature, no breaking changes.

**Documentation / background:** Issues #644 and #659 in the repo. The `useExhaustiveDependencies` deferrals originate from the Biome migration (#636); the duplicate-key warning was surfaced by the Storybook story sweep (#648).

- `validHourMin` was verified to reference only its own parameters and JS built-ins (no component state/props), which is what makes the hoist safe and behavior-preserving. The date-validation rule logic itself is unchanged — only its scope moved.
- For `SelectField`, dropping `useCallback` is preferred over memoizing the `props` rest object (messy/fragile) or rewriting to explicit prop forwarding (an API-discipline change out of scope here). Since `renderSelect` is invoked exactly once per render, the memo provided no benefit.
- The SelectField story previously set the deprecated `label.isSameRow: true`, which tripped `SelectField`'s own `label.isSameRow` deprecation `console.warn` on every render. The story is migrated to the replacement API (`label.direction: 'row'`) — same layout, no component API change — which clears that warning from the story and the sweep. The deprecated `isSameRow` prop and its warning are left intact in the component for real consumers; removing them is a breaking change out of scope here.

### Pre-merge verification

| Check                                                   |  Result   |
|---------------------------------------------------------|:---------:|
| Lint (`biome lint` on the two changed atoms)            | ✅ PASS   |
| Type-check (`tsc --noEmit` in `packages/ui-lib`)        | ✅ PASS   |
| Unit tests (`vitest run` in `packages/ui-lib`, 2/2)     | ✅ PASS   |
| Story sweep — all 69 stories clean (DatePicker incl.)   | ✅ PASS   |
| Story sweep — example routes `/`, `/line`               | ⚠️ PRE-EXISTING |
| Manual smoke test of the affected flows (Storybook)     | ✅ PASS   |
| Console / log clean of new warnings (Chrome DevTools)   | ✅ PASS   |
| All 3 `biome-ignore` suppressions removed               | ✅ PASS   |
| Docker / deploy artifact builds                         | ⏭️ N/A    |

### Testing strategy — why no bespoke component tests

This change deliberately adds **no new test files**; each issue is already guarded by existing infrastructure:

- **#659 (duplicate React keys)** is covered by `scripts/sweep-stories.mjs`, the headless-Chromium (Playwright) sweep that loads every story and fails on any `console.error` / `pageerror`. React emits duplicate keys as a `console.error`, so the DatePicker story (`filters-molecules--date-picker`) is flagged pre-fix and clean post-fix — this is the same harness that surfaced the warning. No per-component test adds signal it doesn't already provide.
  - **Sweep result (this branch):** all **69/69 stories clean**, including the DatePicker story. The run exits non-zero only on two **example-app routes unrelated to this PR** — `/` (`createRoot()` called twice on the same container, in the example's `src/index.tsx` under React 19 StrictMode/HMR) and `/line` (`removeChild` `NotFoundError` in `LinksPage`/LineUI). Neither touches the three files changed here; both are pre-existing example-app issues tracked separately.
- **#644 (exhaustive-deps)** is a Biome **lint** rule with no runtime signal, so it is guarded by `biome lint ./src` — the three `biome-ignore` suppressions are removed and lint is green.

> Note: `packages/ui-lib`'s `vitest` suite currently contains only an export smoke test (`Button`, `defaultTheme`); it does **not** cover these components. Behavioral confidence for the refactor comes from the story sweep plus the manual Storybook + Chrome DevTools pass below. The one piece not asserted by either is `validHourMin`'s clamping rules — its logic is unchanged by this PR (only its scope moved), so it is out of scope here and noted as a future unit-test candidate.

### Detailed steps to reproduce

**Setup**
- `npm install` (repo root)
- `npm start -w packages/ui-lib` (watch/rebuild the library)
- `npm start -w packages/storybook` (Storybook on `:9009`)

**Reproduction & expected outcome**

1. **DateTimeBlock (#644)** — open *Filters / molecules → Date Picker*.
   - Type into the hour/minute inputs of either time block.
   - Expected: each keystroke fires `setDateCallback` with correctly computed (clamped/valid) times per the existing rules, the display updates, and the console shows **no** `useExhaustiveDependencies` or React warnings.
2. **SelectField (#644)** — open *Form / atoms → Select Field*.
   - All three variants (free width, fixed width, with icon) render; choose an option in the first select.
   - Expected: value updates, placeholder clears (`handleOnChange` fires), no console warnings. (The `label.isSameRow` deprecation warning this story used to emit is gone now that it is migrated to `label.direction: 'row'`.)
3. **DatePicker weekday keys (#659)** — same Date Picker story, EN locale.
   - Expected: the `S M T W T F S` header renders once with **no** React "two children with the same key" warning in the console.

**Static checks**
```bash
# from repo root
npx biome lint packages/ui-lib/src/Filters/atoms/DateTimeBlock.tsx \
                packages/ui-lib/src/Form/atoms/SelectField.tsx   # clean
( cd packages/ui-lib && npx tsc --noEmit )                        # clean
grep -rc "biome-ignore" packages/ui-lib/src/Filters/atoms/DateTimeBlock.tsx \
                        packages/ui-lib/src/Form/atoms/SelectField.tsx  # 0, 0

# Regression guard for #659 (needs Storybook running on :9009)
node scripts/sweep-stories.mjs   # DatePicker story reports no console.error
```

**Visual evidence:** Storybook Date Picker and Select Field stories render and operate unchanged; Chrome DevTools console is free of React duplicate-key and exhaustive-deps warnings after the change.

---

**Closes #644, Closes #659.**

### DIFF
--- a/packages/storybook/src/stories/Form/SelectField.stories.tsx
+++ b/packages/storybook/src/stories/Form/SelectField.stories.tsx
@@ -66,7 +66,7 @@ export const _SelectField = () => {
   const fixLabel = object('Fix Select Label', {
     htmlFor: 'fix_select',
     text: 'Page',
-    isSameRow: true,
+    direction: 'row' as const,
   });
 
   const freeOnChange = (value: string) => {
--- a/packages/ui-lib/src/Filters/atoms/DateTimeBlock.tsx
+++ b/packages/ui-lib/src/Filters/atoms/DateTimeBlock.tsx
@@ -111,6 +111,65 @@ interface IProps {
   isTimeRangeValid?: boolean;
 }
 
+/**
+ *
+ * Description of the rules can be found int https://docs.google.com/spreadsheets/d/1POe9uZxKXtLhQFF6DIV-RclUp7T7oXgSeSLlbYmE38g/edit?usp=sharing
+ */
+const validHourMin = (
+  textHour: string,
+  textMin: string,
+  hasDate: boolean,
+  allowAfterMidnight?: boolean
+): { newHour: number; newMin: number } => {
+  const intHour = Number(textHour.slice(-2));
+  const intMin = Number(textMin.slice(-2));
+
+  const newHour = intHour > 24 ? Number(textHour.slice(-1)) : intHour;
+  const newMin = intMin > 60 ? Number(textMin.slice(-1)) : intMin;
+
+  //Rule 8
+  if (newHour >= 24 && newMin !== -1 && allowAfterMidnight) {
+    return { newHour: 24, newMin: 0 };
+  }
+
+  // Rule 7
+  if (newHour === 0 && newMin === 0 && allowAfterMidnight && !hasDate) {
+    return { newHour: 0, newMin: 1 };
+  }
+
+  // Rule 6
+  if (newHour === 23 && newMin === 60 && !allowAfterMidnight) {
+    return { newHour: 23, newMin: 59 };
+  }
+
+  // Rule 5
+  if (newHour >= 24 && !allowAfterMidnight) {
+    return { newHour: 23, newMin };
+  }
+
+  // Rule 4
+  if (newMin === 60) {
+    return { newHour: newHour + 1, newMin: 0 };
+  }
+
+  // Rule 3
+  if (newHour > 0 && newMin === -1) {
+    return { newHour: newHour - 1, newMin: 59 };
+  }
+
+  // Rule 2
+  if (newHour === 0 && newMin === -1) {
+    return { newHour, newMin: 0 };
+  }
+
+  // Rule 1
+  if (newHour === -1) {
+    return { newHour: 0, newMin };
+  }
+
+  return { newHour, newMin };
+};
+
 const DateTimeBlock: React.FC<IProps> = ({
   allowAfterMidnight = false,
   title,
@@ -120,65 +179,6 @@ const DateTimeBlock: React.FC<IProps> = ({
   date = new Date(),
   setDateCallback = () => {},
 }) => {
-  /**
-   *
-   * Description of the rules can be found int https://docs.google.com/spreadsheets/d/1POe9uZxKXtLhQFF6DIV-RclUp7T7oXgSeSLlbYmE38g/edit?usp=sharing
-   */
-  const validHourMin = (
-    textHour: string,
-    textMin: string,
-    hasDate: boolean,
-    allowAfterMidnight?: boolean
-  ): { newHour: number; newMin: number } => {
-    const intHour = Number(textHour.slice(-2));
-    const intMin = Number(textMin.slice(-2));
-
-    const newHour = intHour > 24 ? Number(textHour.slice(-1)) : intHour;
-    const newMin = intMin > 60 ? Number(textMin.slice(-1)) : intMin;
-
-    //Rule 8
-    if (newHour >= 24 && newMin !== -1 && allowAfterMidnight) {
-      return { newHour: 24, newMin: 0 };
-    }
-
-    // Rule 7
-    if (newHour === 0 && newMin === 0 && allowAfterMidnight && !hasDate) {
-      return { newHour: 0, newMin: 1 };
-    }
-
-    // Rule 6
-    if (newHour === 23 && newMin === 60 && !allowAfterMidnight) {
-      return { newHour: 23, newMin: 59 };
-    }
-
-    // Rule 5
-    if (newHour >= 24 && !allowAfterMidnight) {
-      return { newHour: 23, newMin };
-    }
-
-    // Rule 4
-    if (newMin === 60) {
-      return { newHour: newHour + 1, newMin: 0 };
-    }
-
-    // Rule 3
-    if (newHour > 0 && newMin === -1) {
-      return { newHour: newHour - 1, newMin: 59 };
-    }
-
-    // Rule 2
-    if (newHour === 0 && newMin === -1) {
-      return { newHour, newMin: 0 };
-    }
-
-    // Rule 1
-    if (newHour === -1) {
-      return { newHour: 0, newMin };
-    }
-
-    return { newHour, newMin };
-  };
-
   const [displayHours, setDisplayHours] = useState<string>(format(date, 'mm'));
   const [displayMinutes, setDisplayMinutes] = useState<string>(format(date, 'HH'));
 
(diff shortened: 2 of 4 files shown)


JoshLipps (approved):
LGTM thanks

